Issue #107, opened 2026-04-21
differing filament settings by printer not just nozzle


tonysurma:
**still working on documenting this, so below may be incomplete**

As I get deeper into using this, I am welcoming what it enables but also found another issue of sorts.  Apologies as I try to map this out as I think the bones are there and might be a to add this.

Summary: if I have one printer type (say Prusa Core One) with several different nozzles, then this works really well.  I can open up the modified version of prusaslicer you have, pick my printer and nozzle config like normal.  The pick the synched up Filament you add to the drop down.

From there, I do a transfer of settings from the "base filament" profile to this custom one.  For example, do the compare and copy left to right in prusaslicer so the entry in this system has all the settings from "Prusament PLA".  Then I do the calibrations and when I update the settings it updates them in the database - even for the different nozzles!

The challenge is since it keys off of nozzle only, there are settings in those profiles that are different (like cooling, start gcode, etc.) and those differ by printer (for example, many different settings for XL vs Core One).  Prusaslicer handles this with a base config for the filament that inherits and then changes settings for the two different printers (thus you get "Prusament PLA @XL" and "Prusament PLA @CoreOne".)

Therefore if I take the same filament profile built by this system and use it for the two different printers one or the other settings I transferred are used, there is no way to differentiate by printer.

This happens if the two printers have the same nozzle attributes.

It also happens if the two printers have different nozzle attributes even.  It will update based upon the nozzle differences but still doesn't know about the two printers and their unique sets of settings for the filament.

In the filament editor, there is support for setting Calibrations different by nozzle and then by printer, but it doesn't appear the prusaslicer filament edition can discern which printer and apply the Calibrations.  If this worked it would cover the settings covered by Calibrations but not those outside in the other filament settings.

So possibly two paths:
First, apply calibrations from the printer + nozzle data in prusaslicer as that would solve anything in calibrations which is many settings.  Would need to think more on what the code could key off from the printer settings to do this discernment like is done with the nozzle size and hardened settings.

If one could specify a filament profile to inherit from both at the filament level (as is current - although note it doesn't seem like that is used for the settings generated in prusaslicer, even though the inherits is in the ini export...) and then at the nozzle  + printer combo level (like in the calibration boxes) all those settings (like Cooling (those not already in calibrations), Advanced, Filament Overrides, and custom G-Code) could stay out of this database and you would set the filament profile created for the filament in prusaslicer to inherit that appropriate base.

hyiger:
[...]

PR #111, opened 2026-04-23
Fix #106: variant edit silently severs parent inheritance

### LINKED ISSUE #106: Cloning from a parent overrides settings on creation and is then one time only
Here is the scenario:

1. Filament A has settings set.
2. Clone A to B.
3. Update Name, Color, Color Name.  Click Update Filament on B to save it.
4. B inherits settings from A.
5. Make changes to A (without touching B.)
6. B does not reflect those changes.
7. Clone A to C.
8. C has all the current settings from A.

The expected behavior is:

1. Filament A has settings set.
2. Clone A to B.
3. Update Name, Color, Color Name.  Click Update Filament on B to save it.
4. B inherits settings from A.
5. Make changes to A (without touching B.)
6. **B reflects those changes.** (and continues to reflect any changes unless manually entered/overwritten on the clone)
7. Clone A to C.
8. C has all the current settings from A.
9. **Going forward any changes made to A go to B & C and so on** (unless manually entered/overwritten on the clone themselves)
10. 

Therefore it appears the cloning and setting of parents is a one time only copy of settings instead of an ongoing relationship.

The reason for this appears to be that in step 3 of the current scenario, the edit screen is populated with A's settings.  Then when you click Update to save the name and color you entered, all those settings are stored and therefore immediately 'overwritten' so they don't inherit.  If you go back in, edit the Clone B and remove settings (leave temperature blank for instance) then it falls back to inherited.  But anytime later you open it to edit any value, it populates the edit screen with parent values again and you lose the inheritance.  Further, while the nozzle calibrations are inherited on create, there appears to be no way to 'blank' them out so they stay inherited.  If you blank out the values they just disappear from the clone.

Key Scenario Need:
The reason this is important is say I have 6 (or many more) colors of the same PLA filament.  My plan was to create the first filament (let's say the Black color) and then make the other 5 colors clones/inherited.  Then when I do calibrations etc. and set settings on the Parent I could enter them once and not 6 times.

Also, even if I wait until I do the calibrations to enter the 5 clones.  If I ever change a setting later I need to go back to the other 5.

Thank you, the integration with your prusaslicer extension has really upgraded our workflow.

hyiger:
Closes #106.

## The bug

Reported by @tonysurma with a clear repro:
1. Filament A has settings set.
2. Clone A to B.
3. Update B's name/color/colorName and Save.
4. B inherits from A.
5. Edit A.
6. **B doesn't reflect the edit** ← broken

## Root cause

The edit page (`/filaments/{id}/edit`) fetches via `GET /api/filaments/{id}`, which always calls `resolveFilament()` and merges parent values into the response. The `FilamentForm` prefills with those **resolved** values, so clicking Save persists copies of every parent field onto the variant — silently severing the live link.

On top of that, the array-field resolver used `variant.calibrations !== undefined ? variant : parent`. Because Mongoose materialises array schema fields as `[]` at load time, `undefined` was unreachable in practice: a fresh variant's `calibrations: []` was treated as "explicit override to empty" and never inherited.

## The fix

Three coordinated changes.

**1. New `?raw=true` on `GET /api/filaments/{id}`**
Skips inheritance resolution. Returns the variant's own overrides (mostly `null` for inheritable fields). For variants, the response also includes the populated parent doc as `_parent` so UI can render helpful placeholders without a second round-trip.

**2. Edit page uses `?raw=true`**
Form now sees blank inputs for inherited fields. A Save with no changes leaves those blank — inheritance preserved.

**3. Array-field inheritance: empty === inherit**
`compatibleNozzles`, `optTags`, `bedTypeTemps`, `calibrations`, `presets` now all inherit from the parent when the variant's array is empty. This matches what users expect and closes the silent-break. Trade-off: no way to explicitly say "this variant has zero calibrations but parent has some" — rare; add a sentinel if/when it comes up.

## Supporting changes

- `PUT` strips server-only response fields (`_parent`, `_variants`, `_inherited`) in case a client echoes the raw GET response back.
- `FilamentForm` no longer hardcodes `diameter ?? 1.75` for variants — they inherit now too.
- New `form.variantHint` i18n banner at the top of the edit form for variants explaining the blank=inherit behaviour.

## Tests

- `tests/variant-edit-preserves-inheritance.test.ts` (new, 6 tests): full round-trip guarding the regression
- `tests/resolveFilament.test.ts` (5 tests inverted): now asserts empty arrays inherit
- 647 tests pass, typecheck + lint clean

## Not addressed (issue #107)

This PR only fixes #106. The separate #107 (per-printer filament settings) is a follow-up.

### DIFF
--- a/src/app/api/filaments/[id]/route.ts
+++ b/src/app/api/filaments/[id]/route.ts
@@ -7,13 +7,31 @@ import "@/models/BedType";
 import { resolveFilament, hasVariants } from "@/lib/resolveFilament";
 import { getErrorMessage, errorResponse } from "@/lib/apiErrorHandler";
 
+/**
+ * GET /api/filaments/{id}
+ *
+ * Returns a single filament with populated references. By default, if the
+ * filament is a variant (has parentId) its inheritable fields are resolved
+ * from its parent so the response is a complete view suitable for display.
+ *
+ * Pass `?raw=true` to skip inheritance resolution and receive the variant's
+ * own values. Fields the variant does not override come back as `null`
+ * (or empty). This is what the edit page needs — prefilling the form with
+ * resolved values and then saving would copy the parent's fields onto the
+ * variant and silently sever the inheritance link (GH #106).
+ *
+ * When `?raw=true` is passed on a parent, the response shape is unchanged
+ * (parents don't inherit from anything).
+ */
 export async function GET(
-  _request: NextRequest,
+  request: NextRequest,
   { params }: { params: Promise<{ id: string }> }
 ) {
   try {
     await dbConnect();
     const { id } = await params;
+    const raw = request.nextUrl.searchParams.get("raw") === "true";
+
     const filament = await Filament.findOne({ _id: id, _deletedAt: null })
       .populate("compatibleNozzles")
       .populate("calibrations.nozzle")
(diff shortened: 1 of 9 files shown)


hyiger:
Addressed in 1e29ada — good catch, Codex was right about the regression.

The submit-side fix (\`parseNum(form.diameter) ?? (parentId ? null : 1.75)\`) wasn't enough on its own because:

1. **Form initial state still defaulted to "1.75"** — `initialData?.diameter?.toString() || "1.75"` filled the input visibly, so even a no-op save submitted 1.75. Fixed: variants now start with an empty input.
2. **Mongoose schema default** — `POST /api/filaments` called `Filament.create(body)` and the schema's `default: 1.75` materialised onto variants that didn't pass an explicit diameter. So the bug was present at creation too, not just edit. Fixed: the POST route explicitly sets `body.diameter = null` when creating a variant without one.

Added two regression tests:
- Variant whose parent has `diameter: 2.85` survives a raw GET → PUT round-trip without collapsing to 1.75.
- `POST /api/filaments` with `parentId` and no diameter produces a variant with `diameter: null`; standalone creates still get the 1.75 default.

649 tests pass, typecheck + lint clean.

PR #137, opened 2026-04-30
Fix #136: OpenPrintTag fetch fails in Docker (curl not in image)

### LINKED ISSUE #136: Docker version not possible to browse OpenPrintagDB
Environment

* Platform: Raspberry Pi 5 (ARM64)
* Docker: latest
* image: ghcr.io/hyiger/filament-db:v1.13.0

Failed to load database
Error: Failed to fetch OpenPrintTag database

Container logs show:
/bin/sh: curl: not found
tar: invalid magic
tar: short read
OpenPrintTag fetch error

Fixed by manually installed curl:
docker exec -it --user root filament-db sh
apk add --no-cache curl tar gzip

(PS Im not really good in this, took help of AI to sort it out)

hyiger:
Closes #136. Thanks @gertlind for the diagnostic container logs.

## The bug

The community-database fetch shelled out to:

```sh
curl -sL "$tarballUrl" | tar xz -C "$tmpDir"
```

which works in dev (curl is everywhere) but fails in our production Docker image because `node:22-alpine` doesn't include curl. Users on the Pi 5 ARM64 build (and any other Docker deployment) saw:

```
/bin/sh: curl: not found
tar: invalid magic
tar: short read
OpenPrintTag fetch error
```

…and the only workaround was `apk add --no-cache curl tar gzip` inside the running container.

## The fix

Replace the shell pipeline with a pure-Node fetch + tar extraction stream:

```ts
const response = await fetch(tarballUrl, {
  headers: { Accept: "application/vnd.github+json", "User-Agent": "filament-db" },
  signal: AbortSignal.timeout(60_000),
});
if (!response.ok) throw new Error(\`GitHub tarball request failed: \${response.status}\`);
await pipeline(
  Readable.fromWeb(response.body),
  tar.x({ cwd: tmpDir }),
);
```

Same on-disk effect, no host-tool dependency. Works identically in dev, Electron, and the Docker image. As a bonus, a 4xx/5xx response is now a thrown error with the status code instead of a downstream tar parse failure.

## Why we don't need to patch the Dockerfile

By eliminating the shell call, no host-side curl/tar is ever needed. Adding curl to the image would only paper over the underlying portability problem.

## Dependency

`tar@7` was already in `node_modules` as a transitive of `swagger-ui-react` and `mongodb-memory-server-core`. Promoted it to a direct dep in `package.json` so future transitive churn doesn't break OpenPrintTag fetch.

## Tests

Rewrote the two integration tests that mocked `execSync`. They now build a real gzipped tarball on disk, stream it back through a mocked global `fetch`, and assert the same end-to-end behaviour. Added a third test covering the new explicit 404 error path.

- 691 tests pass
- Typecheck + lint clean

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "react": "19.2.4",
         "react-dom": "19.2.4",
         "swagger-ui-react": "^5.32.1",
+        "tar": "^7.5.0",
         "yaml": "^2.8.3"
       },
       "devDependencies": {
@@ -798,6 +799,7 @@
       "version": "1.10.0",
       "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
       "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
+      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
@@ -1641,6 +1643,7 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1663,6 +1666,7 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1685,6 +1689,7 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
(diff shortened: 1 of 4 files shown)


hyiger:
Addressed in 4abb31d — good catch.

The `curl -L` pipeline was implicitly proxy-aware via `HTTP_PROXY` / `HTTPS_PROXY`. Node's built-in fetch isn't, by default. I wired in undici's `EnvHttpProxyAgent` and pass it as a `dispatcher` only when one of `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` (lower- or upper-case) is set in the environment, so:

- Proxy-restricted deployments behave like the old curl path again.
- The no-proxy fast path is unchanged (no dispatcher passed, no overhead).

Promoted `undici` to a direct dep (was already a transitive at 6.25.0).

Added tests:
- 6 cases for the new exported `getProxyDispatcher` helper covering the env-var matrix + empty-string degenerate case.
- An integration test asserting the dispatcher is attached to the fetch call when `HTTPS_PROXY` is set, and absent when no proxy env var is present.

701 tests pass, typecheck + lint clean.

PR #152, opened 2026-05-02
Mobile: restore Atlas credentials on app start

hyiger:
**Stacks on top of [#151](https://github.com/hyiger/filament-db/pull/151)** (mobile typecheck + CI gate). Final target is `feature/mobile-app`. Once #151 lands, this PR's base will be retargeted automatically by GitHub.

## Summary
`AtlasProvider` wrote App ID + API key to SecureStore on connect but never read them back on app start. Every cold launch left `config: null` and `isConnected: false`, so users had to re-enter the same credentials each time and re-tap Connect to use any data screen.

- On-mount `useEffect` pulls both keys from SecureStore, calls `atlasService.connect(...)`, and flips connected state when both succeed.
- Auto-reconnect failures (network down, key revoked) log a warning but leave `isConnected` false — the existing connect screen is the documented recovery path.
- New `isRestoring` flag on the context lets consumers tell the cold-start reconnect window apart from a steady-state "not connected" so they don't flash a setup prompt during boot.
- `clearConfig` now calls `atlasService.disconnect()` too — without it a signed-out user kept their cached Realm session live.

## Test plan
- [x] `cd packages/mobile && npm run typecheck` — clean.
- Manual verify on a build (cold start with stored creds → connects without re-entry; without creds → restore window completes, app sits at not-connected; clearConfig → service.disconnect actually fires).

### DIFF
--- a/packages/mobile/providers/AtlasProvider.tsx
+++ b/packages/mobile/providers/AtlasProvider.tsx
@@ -1,5 +1,6 @@
-import { createContext, useContext, useState, useCallback, type ReactNode } from "react";
+import { createContext, useContext, useState, useCallback, useEffect, type ReactNode } from "react";
 import * as SecureStore from "expo-secure-store";
+import { atlasService } from "../services/atlas";
 
 interface AtlasConfig {
   appId: string;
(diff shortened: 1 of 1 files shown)


hyiger:
@codex review

Latest commit addresses the prior review. Please re-check whether the fix is sufficient or surfaces a new concern.

hyiger:
Folded into #94 — the Atlas-restore commits (fb3b610, 3c65bd0) are now part of `feature/mobile-app`. Combining the three Mobile PRs per request to land them as one.

PR #149, opened 2026-05-02
Project filament list to summary shape; wire missing-calibration filter

hyiger:
## Summary
The list endpoint was returning the full Mongoose document for every filament — including each spool's `photoDataUrl` (a base64 image up to a couple hundred KB), `usageHistory`, and `dryCycles` — even though the list page only consumes ~10 summary fields (`FilamentSummary` in `src/types/filament.ts`).

This PR also fixes the "Missing calibration" quick filter on the list page, which was effectively a no-op: the chip was reachable, but the `noCalibration` count was hardcoded to zero and the filter case fell through to `return true` because `FilamentSummary` had no signal to test against.

- `GET /api/filaments` now uses a Mongoose aggregation pipeline that projects exactly the FilamentSummary fields, strips heavy spool subfields, and computes a single boolean `hasCalibrations` (`$gt: [{ $size: { $ifNull: ["$calibrations", []] } }, 0]`).
- `FilamentSummary` gains `hasCalibrations?: boolean`.
- `src/app/page.tsx` counts filaments with `!hasCalibrations` and filters by the same predicate.

The detail endpoint (`GET /api/filaments/{id}`) still returns the full document and remains the source of truth for fields not in the summary.

## Test plan
- [x] New `tests/filaments-list-route.test.ts` (4 tests):
  - spool.photoDataUrl / usageHistory / dryCycles stripped from list payload
  - hasCalibrations is `true` for a filament with calibrations, `false` for a bare one
  - full `calibrations` array not included in the list payload
  - existing vendor/type filters still work
- [x] Full `npm test` — 821/821 pass on this branch (4 new tests; 0 regressions).
- [x] `npm run lint` — clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #153, opened 2026-05-02
Sync expansion: bedtypes, printhistories, sharedcatalogs + bedType + amsSlots remap

hyiger:
P1 audit follow-up. Closes the v1.11 sync gap where `bedtypes`, `printhistories`, and `sharedcatalogs` were never synced, plus two newer references the existing transforms didn't cover (`calibrations[].bedType`, `printer.amsSlots[].filamentId`).

### Sync order
Now: nozzles → printers → locations → bedtypes → filaments → printhistories → sharedcatalogs.

- **bedtypes** sync runs **before** filaments so the filament transform can remap `calibrations[].bedType`. Reuses the same `reconcileByName` pre-pass used for locations (BedType has the same partial-unique-name index that would E11000 the very first sync if both desktops created the same name independently). I factored the reconciliation out of `reconcileLocationsByName` into a generic `reconcileByName(collectionName)` and now both call into it.
- **printers** still sync **before** filaments to break the `calibrations.printer` ↔ `amsSlots.filamentId` reference cycle. After filament sync settles, a new **`repairPrinterAmsSlots`** pass projects the correct filament id via the freshly-built syncId maps. Same pattern as `repairFilamentParentIds` / `repairDanglingSpoolLocations`. Doesn't bump `updatedAt`.
- **printhistories** sync runs after filaments. New `buildPrintHistoryTransform` remaps `printerId` + `usage[].filamentId` via syncId; `usage[].spoolId` is cleared on insert/update because spool subdocuments don't have stable cross-side ids. Usage entries whose filament can't be resolved on the target side are dropped (better than persisting a dangling pointer; the job ledger entry survives).
- **sharedcatalogs** sync runs last. Straight last-write-wins on the syncId/slug — no refs to remap because the payload is denormalised at publish time.

### Filament transform
- Added `bedType` remap in `buildFilamentRefsTransform` alongside the existing nozzle/printer remap. Unknown bedType on the target clears to null (same model as printer/location), keeping the calibration entry rather than dropping it.

### Spool-id limitation (documented in class header + inline)
Spool subdocuments inside Filament don't have stable cross-side identifiers. Anything that references a spool by id — `printer.amsSlots[].spoolId`, `printhistory.usage[].spoolId` — clears that id during cross-side remap. Per-filament gram totals still reconcile; per-spool attribution is dropped pending a spool-syncId migration (a separate, larger change).

### Subtle MongoDB query gotcha
The naive `{ "amsSlots.filamentId": { $ne: null } }` excludes any printer with *any* slot whose `filamentId === null`, even if a sibling slot is set — Mongo's array-positional negated-equality matches at the whole-array level. Used `{ amsSlots: { $elemMatch: { filamentId: { $ne: null } } } }` instead. Discovered via the test feedback loop on the amsSlots repair test, which is the only reason the bug isn't shipping.

## Test plan
- [x] 8 new tests in `tests/sync-service-expansion.test.ts` — bedtype push, bedtype name reconciliation, calibrations[].bedType remap across sides, amsSlots[].filamentId rewrite (with spoolId clear), amsSlots cleared for orphan filament, printhistory printerId + filamentId remap (with spoolId clear), printhistory usage entries dropped when filament unresolvable, sharedcatalog push.
- [x] Existing `tests/sync-service.test.ts` (12) + `tests/sync-service-locations.test.ts` (9) still pass — the location reconciliation now goes through `reconcileByName` but the behavior is identical.
- [x] Full `npm test` — 825/825 pass.
- [x] `npm run lint` — clean.

### DIFF
--- a/electron/sync-service.ts
+++ b/electron/sync-service.ts
@@ -43,12 +43,15 @@ interface SyncResult {
 /**
  * Bidirectional sync engine between local MongoDB and Atlas.
  * Uses last-write-wins conflict resolution based on updatedAt timestamps.
- * Nozzles, printers, and locations are synced first so filaments (and their
- * embedded spools) can have their references remapped onto the target DB's IDs.
+ * Nozzles, printers, locations, and bedtypes are synced first so filaments
+ * (and their embedded spools) can have their references remapped onto the
+ * target DB's IDs. Printhistories and sharedcatalogs sync after filaments.
  *
- * NOTE: bedtypes, printhistory, and sharedcatalogs are NOT synced yet — they
- * were added in v1.11 alongside locations and have the same gap. They'll need
- * the same treatment when their data starts diverging across desktops.
+ * Known limitation: spool subdocuments inside Filament don't have stable
+ * cross-side identifiers. Anything that references a spool by id —
+ * printer.amsSlots[].spoolId, printhistory.usage[].spoolId — clears that
+ * id during cross-side remap. Per-filament gram totals still reconcile;
+ * per-spool attribution is dropped pending a spool-syncId migration.
  */
 export class SyncService extends EventEmitter {
   private localUri: string;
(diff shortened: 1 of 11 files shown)


hyiger:
@codex review

Latest commit addresses the prior review. Please re-check whether the fix is sufficient or surfaces a new concern.

hyiger:
@codex review

Round-2 fix pushed. Please re-check whether the addressed feedback is satisfied or surfaces new concerns.

hyiger:
@codex review

Round-3 fix pushed.

hyiger:
@codex review

Round-4 fix pushed (PR #153) / round-3 fix pushed (PR #149) — please re-check.

PR #155, opened 2026-05-02
Fix #154: idempotent auto-updater + skip broken mac auto-download

### LINKED ISSUE #154: Update check failed version.
Get below error trying 

Update check failed: ZIP file not provided: [ { "url": "https://github.com/hyiger/filament-db/releases/download/v1.13.0/FilamentDB-1.13.0-mac-x64.dmg", "info": { "url": "FilamentDB-1.13.0-mac-x64.dmg", "sha512": "Gg0knIUv59fRZu02OUi+z1Hfrg+6FazHtJ7x6RNPJooXu95LxpBCtKIGV8zM1VmmTHLJjs9bhTrfah9+cG9MOw==", "size": 243143711 } } ]

After restarting following error:
Uncaught Exception:
Error: Attempted to register a second handler for 'update-get-status'
at IpcMainImpl.handle (node:electron/js2c/browser_init:2:118334)
at initAutoUpdater (/Applications/Filament DB.app/Contents/Resources/app/electron-dist/main.js:33938:28)
at createWindow (/Applications/Filament DB.app/Contents/Resources/app/electron-dist/main.js:34079:5)
at EventEmitter.<anonymous> (/Applications/Filament DB.app/Contents/Resources/app/electron-dist/main.js:34590:9)
at EventEmitter.emit (node:events:509:28)

Runnin Macbook M1, the error message indicates that  it tries to get Mac-x64.dmg version.

hyiger:
Closes #154.

Two bugs surfaced by the v1.13.0 upgrade prompt for an Apple Silicon user.

## 1. Crash on app reopen (the hard bug)

`initAutoUpdater()` was invoked from `createWindow()` in `electron/main.ts`. On macOS, closing the window nulls \`mainWindow\` but the app keeps running; clicking the dock icon fires \`app.on(\"activate\")\` → \`createWindow()\` → \`initAutoUpdater()\` AGAIN. The second call hits:

\`\`\`
Error: Attempted to register a second handler for 'update-get-status'
\`\`\`

…and crashes the app on every reopen.

**Fix**: track an \`initialized\` flag at module scope. Subsequent calls just refresh the \`mainWindow\` reference (so future status emits target the newly-opened window) and re-emit the current state immediately so the renderer doesn't need to wait for the next tick.

## 2. Cryptic "ZIP file not provided" on macOS download (the soft bug)

The release workflow runs separate matrix jobs for arm64 and x64 mac builds; each produces a \`latest-mac.yml\` and they clobber each other in the GitHub Release upload step. Combined with the fact that we don't publish mac \`.zip\` artifacts (only dmg, but electron-updater needs zip for auto-install), Apple Silicon users got pointed at the wrong installer URL and saw an opaque ZIP error.

Properly fixing macOS auto-update needs zip publishing + per-arch yml + signing — out of scope for this hotfix. **Workaround**: hide the **Download** button on macOS in \`UpdateBanner\` and surface only **View release** (promoted from a small underline link to a full button so it visually replaces Download). The original intent in \`auto-updater.ts\` has always been "macOS users open the release page and grab the dmg manually"; this just makes the UI match the intent.

The proper mac auto-update work is noted in an inline comment on \`UpdateBanner\` for a future change.

## Test plan

- [x] New \`tests/auto-updater.test.ts\` mocks electron + electron-updater; asserts each \`ipcMain.handle\` channel is registered exactly once across three \`initAutoUpdater(win)\` calls, and that the new window receives the re-emitted status on reinit.
- [x] Full \`npm test\` — 870/870 pass.
- [x] \`npm run lint\` — clean.

### DIFF
--- a/electron/auto-updater.ts
+++ b/electron/auto-updater.ts
@@ -30,6 +30,14 @@ interface UpdateInfo {
 
 let mainWindow: BrowserWindow | null = null;
 let currentState: UpdateInfo = { state: "idle" };
+/** Tracks whether initAutoUpdater has done its one-time setup (IPC handlers,
+ * autoUpdater listeners, periodic-check timers) for this process. The
+ * function is called from `createWindow()` in electron/main.ts, which on
+ * macOS runs every time the user clicks the dock icon after closing the
+ * window. Without this guard the second call hits
+ * `Error: Attempted to register a second handler for 'update-get-status'`
+ * and crashes the app on reopen (GH #154). */
+let initialized = false;
 
 function emit(update: Partial<UpdateInfo>) {
   currentState = { ...currentState, ...update };
@@ -37,7 +45,19 @@ function emit(update: Partial<UpdateInfo>) {
 }
 
 export function initAutoUpdater(win: BrowserWindow) {
+  // Always refresh the window reference — the previous window may have
+  // been closed and the renderer for the new window needs to receive
+  // future status events.
   mainWindow = win;
+
+  if (initialized) {
+    // Re-emit the current state into the new window so the renderer's
+    // initial getSyncStatus() / status listener attaches to a fresh value
+    // instead of waiting for the next tick.
+    emit({});
+    return;
+  }
+  initialized = true;
 
   // IPC surface is registered unconditionally so the dev-mode renderer can
   // still call update-get-status (and friends) without crashing with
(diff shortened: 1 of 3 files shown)


hyiger:
Codex P2 (`makeWin()` shared sendSpy) addressed in #157 as a follow-up since this PR is already merged. Each fake window now gets its own `vi.fn()`, and the redirect test additionally asserts the old window's send count is unchanged after re-init — which is the real GH #154 invariant and was unreachable with the shared spy.

PR #179, opened 2026-05-03
Fix #176: ASAR-bundle standalone server + window-show safety net

### LINKED ISSUE #176: App does not launch on Windows Desktop v1.14.0 - arm64
Somewhere in the 1.13+ and/or v1.14.0 versions, after a restart & upgrade it prompted about a service control thing (don't have screen shot) and even with an uninstall and reinstall it no longer seems to run as a desktop foreground app but only as a background app and never comes on screen.

If it helps here is what task manager shows after launch:

Image

If I get time later will try an earlier version install and/or another PC

hyiger:
## Summary

Two-pronged fix for [#176](https://github.com/hyiger/filament-db/issues/176) — slow Windows install AND "process runs but no window appears" — based on the user reports from @tonysurma and @GLAFEBE6.

### 1. `asar: true` + `asarUnpack: ["**/*.node"]` (`electron-builder.yml`)

Tony correctly diagnosed the slow-install root cause: "many, many little js files." The standalone Next.js bundle has thousands of tiny JS files; with `asar: false`, NSIS extracts each one individually and Windows Defender real-time-scans every file as it lands. ASAR collapses them into one archive that copies in a single I/O burst and is opaque to per-file AV scanning. `.node` binaries are unpacked because `dlopen()`/`LoadLibrary()` can't mmap from inside an ASAR archive — electron's asar shim transparently rewrites paths to the unpacked copy at runtime.

### 2. Window-show safety net + diagnostic log (`electron/main.ts`)

Today the window is created with default `show: true` and `loadURL` fires immediately. If the renderer hangs (KB5083631 GPU driver regression, Defender holding files, server slow to respond), there's no recovery — the user sees a phantom background process. New behavior:

- `show: false` + `ready-to-show` — paints when the renderer says it's ready (also kills FOUC).
- 20s safety-net timeout that forces the window visible even if `ready-to-show` never fires. A blank window the user can interact with beats nothing.
- Diagnostic log at `userData/logs/main.log` capturing: startup info, lock acquire/deny, `createWindow` entry, `did-fail-load`, `render-process-gone`, `unresponsive`, `child-process-gone` (covers GPU + utility process crashes). Best-effort writer that never throws — gives us real data when the next user reports this instead of "what does Task Manager show".

## Out of scope

Code signing the Windows binary was the medium-term recommendation in the issue analysis but is **not** part of this PR (per maintainer decision — paid certs aren't on the table). The diagnostic log will tell us whether SAC / Defender suspension is actually what's hitting users post-KB5083631; if so, free options like SignPath OSS sponsorship or self-signed-with-SmartScreen-rep-building can be revisited later.

## Test plan

- [x] `npm run lint` — passes (5 pre-existing warnings, none from changed files)
- [x] `npm test` — 994/994 passing (51 files)
- [x] `npm run electron:compile` — clean
- [x] `npm run build` — Next.js production build clean
- [ ] **Smoke test on Windows after CI builds the installer:**
  - [ ] Verify install time vs. v1.14.1 on a clean Windows 11 VM with Defender enabled
  - [ ] Verify NFC reader still works (ensures `asarUnpack` correctly exposed `pcsclite.node`)
  - [ ] Verify `userData\logs\main.log` is written and contains lifecycle entries
  - [ ] Verify window appears within ~20s on cold first run

If `asarUnpack: ["**/*.node"]` turns out to need a more specific glob for `mongodb-memory-server-core`'s downloaded mongod binary, that's a fast follow-up.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -9,7 +9,20 @@ directories:
 
 icon: assets/icon.png
 
-asar: false
+# Bundle the standalone Next.js server + node_modules into an ASAR archive.
+# The standalone bundle is thousands of small JS files; without ASAR the
+# NSIS installer extracts each one individually and Windows Defender real-
+# time-scans every file, producing the multi-minute install stall users
+# reported in GH #176. ASAR collapses them into one opaque archive that
+# copies in a single I/O burst and is opaque to per-file AV scanning.
+#
+# .node native binaries cannot be loaded from inside an ASAR archive — they
+# must live on disk so dlopen()/LoadLibrary() can mmap them. Unpack them
+# explicitly so Electron's asar shim transparently rewrites paths to the
+# unpacked copy at runtime.
+asar: true
+asarUnpack:
+  - "**/*.node"
 npmRebuild: false
 
 files:
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1,11 +1,34 @@
 import { app, BrowserWindow, Menu, ipcMain, dialog, utilityProcess, UtilityProcess, shell, session } from "electron";
 import path from "path";
+import fs from "fs";
 import Store from "electron-store";
 import http from "http";
 import { NfcService } from "./nfc-service";
 import { startLocalMongo, stopLocalMongo } from "./local-mongo";
 import { SyncService, SyncStatus, getDbNameFromUri } from "./sync-service";
 import { initAutoUpdater } from "./auto-updater";
+
+// ── Diagnostic log ──
+// Writes lifecycle and crash events to a file in userData so users on
+// machines where the window never appears (GH #176) can attach a log
+// instead of guessing at console output they can't see. Best-effort only —
+// never throws, never blocks startup. Entries are also mirrored to
+// console.log for the dev workflow.
+const LOG_PATH = path.join(app.getPath("userData"), "logs", "main.log");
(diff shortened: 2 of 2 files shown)


GLAFEBE6:
Thanks for all the work on this project i am excited to get it working!

i uninstalled the previously installed 1.14.1 did a fresh install of version 1.14.3
here is the main.log at first launch:

[2026-05-03T23:31:57.516Z] startup: pid=12104 platform=win32 version=1.14.3 packaged=true
[2026-05-03T23:31:57.553Z] single-instance lock acquired
[2026-05-03T23:31:57.760Z] app ready
[2026-05-03T23:31:57.873Z] server spawned
[2026-05-03T23:32:09.042Z] server.stdout: ▲ Next.js 16.2.4
[2026-05-03T23:32:09.044Z] server.stdout: - Local:         http://localhost:3456
- Network:       http://localhost:3456
✓ Ready in 0ms

No window popped up but it is showing it is running in the task manager
i go to  http://localhost:3456 and it is running
but i think i am missing an initialization of the database that would be completed during the first launch
the status says "connected"
But when I click Add filiament/Nozzel/Printer or anything else i get some errors after populating the fields like these screenshots:
addNozzleError_v1_14_3
AddPrinterError_v1_14_3
addfilamentError_v1_14_3

PR #213, opened 2026-05-10
Filament trash workflow + Import/Export page

hyiger:
## Summary

Three commits adding a soft-delete trash workflow for filaments, a discoverable `/import-export` page, and the test + documentation coverage to back both.

| | |
|---|---|
| Tests | **+66 cases** across 6 new files + 1 extended → **1099 total**, lint clean, build clean |
| Docs | `docs/api.md`, `README.md`, `docs/usage.md` updated |
| Branch | 3 commits, no merge conflicts with main |

### 1. Filament trash workflow (`8b07e3e1`)

The Filament model already had `_deletedAt` for soft-delete and the GET routes already filtered it out — this commit wires up the user-facing flow on top.

**API:**
- `GET /api/filaments/trash` — list soft-deleted filaments, lightweight projection sorted newest first
- `POST /api/filaments/{id}/restore` — un-soft-delete. Returns 409 if a new active filament has reused the trashed one's name (the partial unique index on `name` only covers non-deleted docs)
- `DELETE /api/filaments/{id}?permanent=true` — hard delete. Only allowed once a filament is already in the trash; refuses if trashed variants still reference it

**UI:**
- New `/trash` page with Restore + "Delete forever" buttons per row, plus an "Empty trash" bulk action
- Settings page gets a Trash tile
- The bulk-delete bar on the filament list now shows a small "open trash" link so the destination is discoverable right when you delete

### 2. `/import-export` page (`74ac7f2d`)

Mirrors the Import/Export dropdown on the filament list as a standalone page. The dropdown stays — most users will use it because it's right where they manage filaments — but this gives Settings users a discoverable entry point and a stable URL for documentation/links. Reuses the same dialog components (`PrusamentImportDialog`, `ImportAtlasDialog`, `SpoolCsvImportDialog`) so behaviour is identical.

### 3. Test + docs coverage (`ec7d3af3`)

Tests landed in this commit (route-level only — model-level tests already existed):

- `tests/nozzles-route.test.ts` (14) — covers the **printer-enrichment reverse lookup** that powers the v1.11 differentiate-nozzles-by-printer feature
- `tests/printers-route.test.ts` (14) — CRUD + `Filament.calibrations.printer` cascade guard
- `tests/bed-types-route.test.ts` (13) — CRUD + calibration ref guard
- `tests/print-history-get.test.ts` (10) — GET filtering + `?limit` clamping (existing print-history test only covered POST/DELETE)
- `tests/filaments-import-export-route.test.ts` (11) — HTTP-level paths (missing file → 400, soft-delete exclusion in exports)
- `tests/snapshot.test.ts` (+4) — v3 collection round-trip for Location, PrintHistory, and `spool.locationId` ObjectId rehydration

Documentation updates:
- `docs/api.md` — trash workflow + `GET /api/spools/export-csv` (both were entirely missing)
- `README.md` — `/trash` and `/import-export` added to project structure
- `docs/usage.md` — "Restoring or permanently deleting from the trash" + new "Bulk Import / Export" section

## Test plan

- [ ] `git fetch && git checkout feature/filament-trash && npm install && npm run dev`
- [ ] Delete a filament from the bulk-action bar on the filament list — verify it disappears and the "open trash" link appears
- [ ] Visit `/trash` — verify the deleted filament appears with timestamp, color swatch, vendor/type
- [ ] Click **Restore** — verify it comes back to the main list
- [ ] Delete it again, click **Delete forever**, confirm — verify it's gone from `/trash`
- [ ] Try restoring after creating an active filament with the same name — verify the 409 error message
- [ ] Settings → **Import / Export** tile → verify all dialog actions work (Prusament, Atlas, Spool CSV, file upload, exports as direct download)
- [ ] `npm test` — verify 1099 passing across 61 files
- [ ] `npm run build` — verify build clean


hyiger:
@codex addressed in 041c4096.

You were right — a `deleteOne` on one peer would get resurrected by the syncCollection pairing logic on the next cycle. Fix:

- Added `_purged: boolean` to the Filament model as a stronger tombstone than `_deletedAt`.
- The DELETE `?permanent=true` path now sets `{ _purged: true, _deletedAt: now }` instead of `deleteOne`.
- `electron/sync-service.ts` got a new conflict-resolution branch that propagates `_purged` across peers before the existing soft-delete and last-write-wins logic. Purge wins over a remote update that happened after the local purge — one-way signal.
- `GET /api/filaments/trash` and `POST /api/filaments/{id}/restore` both filter purged rows so they never reappear in any user-visible place.

Tests:
- `tests/trash-route.test.ts` — assertions updated to expect a tombstone (`findById` returns the row with `_purged: true`) instead of null. +3 new cases for tombstone hiding, restore refusal, idempotent re-purge.
- `tests/sync-service-expansion.test.ts` — +4 cases covering purge propagation in both directions, both-purged no-op, and the purge-wins-over-newer-remote-update edge case.

Docs: `docs/api.md` updated to document the tombstone semantics on the permanent-delete + trash + restore endpoints.

Total: 1106 tests, lint clean, build clean.

PR #246, opened 2026-05-22
Clear the NFC status pill after erasing a tag

hyiger:
## Summary

After programming a tag, the status pill shows `Loaded: <filament>`. Going to **Settings → Erase Tag** and erasing the tag left the pill **still showing `Loaded: <filament>`** — it only cleared once the user physically lifted the now-blank tag off the reader.

**Root cause:** the pill's `loadedTagName` is only cleared by the tag-removed effect (`if (!status.tagPresent) setLoadedTagName(null)`). The erase flow in `settings/page.tsx` calls `nfcFormatTag()` but never touches the `NfcProvider` state — so a tag that's erased while still on the reader keeps reporting its old (now-wiped) filament.

## Fix

Add `notifyTagErased()` to the NFC context, mirroring the existing `notifyTagWritten()`. It clears `loadedTagName`, the lingering `tagReadResult`, and closes any open read dialog — all now stale once the tag is blank. The Settings erase handler calls it immediately after a successful `nfcFormatTag()`.

The pill then drops to its normal "tag present, not decoded" state (a blank tag is still physically on the reader), instead of falsely reporting the erased filament.

## Test plan
- [x] `eslint` + `tsc` clean
- [ ] **Needs hardware verification** — write a tag, confirm `Loaded: <name>`, erase via Settings, confirm the pill clears *without* lifting the tag. The change is renderer-only state wiring; it can't be exercised in the web preview (no NFC reader / Electron).

### DIFF
--- a/src/app/settings/page.tsx
+++ b/src/app/settings/page.tsx
@@ -8,9 +8,11 @@ import { useTranslation } from "@/i18n/TranslationProvider";
 import { LOCALES } from "@/i18n";
 import { useTheme, type ThemePreference } from "@/components/ThemeProvider";
 import { useIsElectron } from "@/hooks/useIsElectron";
+import { useNfcContext } from "@/components/NfcProvider";
 
 export default function SettingsPage() {
   const { t, locale, setLocale } = useTranslation();
+  const { notifyTagErased } = useNfcContext();
   const [restoring, setRestoring] = useState(false);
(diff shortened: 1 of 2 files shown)


hyiger:
Folded into #245 (cherry-picked as 6b48bac) so both NFC fixes ship together. Closing in favour of that PR.

PR #324, opened 2026-05-22
fix: restore focus on dialog close and move focus on wizard step change

xxiaoxiong:
Fixes #320

## Problem

Three dialog components (`NfcReadDialog`, `ImportAtlasDialog`, `PrusamentImportDialog`) never capture/restore `document.activeElement`, so focus drops to the document body on close. The wizard dialogs also never move focus into the newly-rendered step on a connect-select-confirm transition.

## Impact

Keyboard and screen-reader users lose their place when the NFC dialog (which auto-opens on every scan) closes, and after each wizard step. This is a **P2 accessibility** issue.

## Solution

Applied the same focus management pattern already used in `UnsavedChangesDialog`:

### 1. Capture previous focus on mount

```ts
const previouslyFocused = document.activeElement as HTMLElement | null;
```

### 2. Restore focus on cleanup

```ts
return () => {
  document.removeEventListener('keydown', handleTab);
  previouslyFocused?.focus?.();
};
```

### 3. Move focus to first interactive element on wizard step change

```ts
useEffect(() => {
  const firstFocusable = dialog.querySelector<HTMLElement>(
    'button:not([disabled]), [href], input:not([disabled]), ...'
  );
  firstFocusable?.focus();
}, [step]);
```

## Changes

- **NfcReadDialog.tsx**: Added `previouslyFocused` capture/restore
- **ImportAtlasDialog.tsx**: Added `previouslyFocused` capture/restore + step focus management
- **PrusamentImportDialog.tsx**: Added `previouslyFocused` capture/restore + step focus management

## Testing

**Manual keyboard navigation test:**
1. Focus an input field
2. Trigger NFC scan (dialog auto-opens)
3. Press Escape to close
4. ✅ Verify focus returns to the input field

**Wizard step test:**
1. Open `ImportAtlasDialog` or `PrusamentImportDialog`
2. Complete first step (connect/input)
3. ✅ Verify focus moves to first interactive element in next step
4. Complete wizard or press Escape
5. ✅ Verify focus returns to trigger element

## Size

**S** — 3 files, ~15 lines added per file (focus capture/restore + step focus management)

### DIFF
--- a/src/components/ImportAtlasDialog.tsx
+++ b/src/components/ImportAtlasDialog.tsx
@@ -31,9 +31,11 @@ export default function ImportAtlasDialog({ onClose, onImported }: Props) {
   const [importProgress, setImportProgress] = useState("");
   const dialogRef = useRef<HTMLDivElement>(null);
 
-  // Focus trap
+  // Focus trap: capture previous focus, trap tab, restore on cleanup
   useEffect(() => {
     if (!dialogRef.current) return;
+    
+    const previouslyFocused = document.activeElement as HTMLElement | null;
     dialogRef.current.focus();
 
     const dialog = dialogRef.current;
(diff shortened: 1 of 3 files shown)


hyiger:
Thanks for this, @xxiaoxiong — and apologies for the collision. #320 was fixed in #329, which has since merged to `main`, so this PR now conflicts with it (both touch the same three dialog components).

The merged fix landed the same focus capture/restore for `NfcReadDialog`, `ImportAtlasDialog` and `PrusamentImportDialog`, plus the wizard step-focus — and during its Codex review it also addressed two refinements that this PR's approach still has:

- **Focus restore must not be keyed to `[onClose]`.** `onClose` is created inline by the parent pages, so a parent re-render fires the effect cleanup and bounces focus out of the *still-open* modal. #329 puts the capture/restore in an empty-deps effect (mount/unmount only).
- **Step-focus must prefer the step's field over the header close button.** That's exactly the Codex comment on this PR (`PrusamentImportDialog.tsx:114`) — focus landing on the `×` instead of the spool-ID input. #329 queries `input, select, textarea` first and only falls back to `button, [href]`.

So the issue is already resolved on `main`. Closing this as superseded — but the diagnosis here was spot on, and genuinely appreciated. 🙏

PR #326, opened 2026-05-22
Sync engine: conflict-resolution + clone-name regex fixes (#298, #317)

hyiger:
## Summary

PR 4 of 8 in the code-review issue sweep.

### #317 — sync conflict-resolution (P2)

Three defects were flagged in `syncCollection`. **Two are real and fixed; the third doesn't hold up.**

1. **Resurrection on a timestamp tie** — a row soft-deleted locally at time `T` while the remote `updatedAt` was also `T` (a delete right after an edit — equal-ms is common) fell through to the resurrect branch. Both delete-vs-update comparisons now use `>=` so the delete wins the tie.
2. **NaN timestamp stall** — comparisons used `new Date(doc.updatedAt).getTime()` directly; a doc missing `updatedAt` → `NaN` → every comparison false → the row never synced. New NaN-safe `SyncService.readTimestamp()` helper; conflict comparisons use `readTimestamp(...) ?? 0`.
3. **"Fresh `_id` breaks SharedCatalog embedded refs" — verified NOT a bug.** The `payload` is a self-contained snapshot copied verbatim; its embedded ids reference each other *within the blob*, not live rows. A fresh document `_id` never touches the payload. No code change — documented in the commit for issue close-out.

### #298 — nozzle clone-name regex over-match (P1)

The clone-name helpers matched peers with a prefix-only anchor (`^base`): cloning "E3D" pulled in "E3D V6" / "E3D Revo", and a peer "E3D Revo #5" pushed the "E3D" counter to "#6". Added `clonePeerNamePattern` (anchored at **both** ends) used by `nextCloneName`, the clone route, and the dbConnect nozzle-split migration. Removed two now-dead `escapeRegExp` copies.

## Test plan

- [x] `tests/nozzle-clone-naming.test.ts` — 9 new cases (exact-match naming, regex-metachar base names, prefix-sibling exclusion)
- [x] `tests/sync-service-expansion.test.ts` — 2 new #317 cases (tie → delete wins; missing `updatedAt` → still merges)
- [x] Existing `nozzle-printer-uniqueness` suite (15) still green
- [x] `npm run lint`, `npx tsc --noEmit`, `npm run electron:compile` — clean

### DIFF
--- a/electron/sync-service.ts
+++ b/electron/sync-service.ts
@@ -464,14 +464,18 @@ export class SyncService extends EventEmitter {
         }
 
         if (localDeleted && !remoteDeleted) {
-          // Deleted locally — propagate if deletion is newer
-          const localDeletedAt = new Date(localDoc._deletedAt).getTime();
-          const remoteUpdatedAt = new Date(remoteDoc.updatedAt).getTime();
-          if (localDeletedAt > remoteUpdatedAt) {
+          // Deleted locally — propagate if the deletion is at least as
+          // recent as the remote update. GH #317: `>=` (not `>`) so the
+          // delete wins on a timestamp tie — an equal-millisecond
+          // delete-right-after-edit must not resurrect the row. NaN-safe
+          // via readTimestamp ?? 0.
+          const localDeletedAt = SyncService.readTimestamp(localDoc._deletedAt) ?? 0;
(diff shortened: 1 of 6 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #327, opened 2026-05-22
Harden import paths: validation, prototype-pollution, BOM, trash safety

hyiger:
## Summary

Audit sweep #3 — import-path correctness and security. Closes #276, #296, #297, #307, #308, #309.

- **#296 — parseCsv prototype pollution**: row objects are built with `Object.create(null)`, so a header literally named `__proto__` / `constructor` / `prototype` becomes an ordinary own key instead of triggering the prototype setter (which silently dropped the column *and* mutated `Object.prototype`).
- **#297 — INI / PrusaSlicer import strands a trashed filament**: both import loops now resurrect-and-update a trashed (non-purged) filament that owns the imported name rather than creating a duplicate active row. A duplicate would strand the trashed one — its restore would 409 forever on the name conflict. Both loops also gained a 10k-section import cap mirroring `parseCsv`'s `maxRows`.
- **#307 — prusament import validates the spool shape**: the renderer-supplied spool payload is fully validated (numeric weights/diameter, `#RRGGBB` colour, required strings) before any DB write — the `$push` bypasses subdocument validators the dedicated spool routes rely on.
- **#308 / #276 — `runValidators` on import updates**: import update calls pass `{ runValidators: true, context: "query" }` so schema validators (including the `tdsUrl` scheme guard) run on the upsert path.
- **#309 — import-csv BOM**: import-csv strips a leading UTF-8 BOM so Excel-saved CSVs no longer fail the required-column check on a `<BOM>Name` first header cell.

## Test plan

- [x] New `tests/import-validation.test.ts` — 7 tests covering each issue
- [x] Existing INI-upsert test updated to assert the new resurrect behavior
- [x] `npm test` — 1232 passed
- [x] `npm run lint` clean

### DIFF
--- a/src/app/api/filaments/import-csv/route.ts
+++ b/src/app/api/filaments/import-csv/route.ts
@@ -48,7 +48,11 @@ export async function POST(request: NextRequest) {
     const sizeError = checkFileSize(file);
     if (sizeError) return sizeError;
 
-    const content = await file.text();
+    // GH #309: strip a leading UTF-8 BOM (U+FEFF). Excel-saved CSVs —
+    // and the app's own xlsx exports — begin with a BOM; left in place
+    // it becomes part of the first header cell, fails HEADER_MAP, and
+    // the required-column check rejects an otherwise-valid file.
+    const content = (await file.text()).replace(/^\uFEFF/, "");
     const lines = content.split(/\r?\n/).filter((l) => l.trim() !== "");
 
     if (lines.length < 2) {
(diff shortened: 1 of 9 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

Issue #293, opened 2026-05-22
Test: Filament.test.ts 'remaining weight' block re-implements the logic instead of testing it


hyiger:
**Severity:** P2 — Test quality

**Problem:** The whole `describe` block at tests/Filament.test.ts:466-599 computes results inline in the test body and asserts on its own arithmetic (e.g. line 506 `expect(spool.totalWeight - filament.spoolWeight).toBe(-90)` is a tautology on test-authored literals). The real `getRemainingGrams` / length helpers in `src/lib/inventoryStats.ts` are never exercised here.

**Impact:** The block looks like it covers remaining-weight logic but covers none of it — a broken `getRemainingGrams` stays green here.

**Suggested fix:** Call the actual helpers from `src/lib/inventoryStats.ts`, or delete the block (it is already covered by `inventoryStats.test.ts`).

---
_Filed from an automated full-codebase review. Verify the cited lines before fixing._

hyiger:
Fixed in #330, merged to `main`.

PR #330, opened 2026-05-22
Fix test quality: honest scope, real helpers, pinned boundaries

hyiger:
## Summary

Audit sweep #3 — test-quality fixes. Closes #264, #293, #294, #295.

- **#264** — the transaction-abort test was titled `"...with rollback semantics"` but its own comments admitted it cannot verify rollback (single-node memory server, not a replica set) and it discarded the captured pre-state with `void before`. Retitled to honest scope — `"...surfaces as a 500 and creates no PrintHistory row"` — and the dead pre-state capture removed. It still asserts the real failure shape (500, no PrintHistory row).
- **#293** — the `Filament.test.ts` "remaining weight" block re-implemented the maths inline and asserted on its own literals (e.g. `expect(spool.totalWeight - filament.spoolWeight).toBe(-90)` — a tautology); the real `getRemainingGrams` helper was never exercised. Replaced with tests that round-trip the `spools` subdocument array through Mongoose and call the actual `src/lib/inventoryStats.ts` helper against the persisted document.
- **#294** — the `parseCsv` row-limit test only checked that *something* threw. Added a describe block pinning the exact boundary — accepts `maxRows` / rejects `maxRows + 1`, header counted toward the cap, and the default 10,000-row limit — so an off-by-one or 2x-too-large cap would now fail a test.
- **#295** — `snapshot.test.ts` re-registered models via plain `import` instead of `mongoose.model(name, schema)`. Switched to the registry-based pattern (used by `locations-route.test.ts` / `print-history.test.ts`) so a future `.populate()` on the snapshot route won't throw "Schema hasn't been registered".

## Test plan

- [x] `npm test` — 1224 passed

### DIFF
--- a/src/lib/parseCsv.ts
+++ b/src/lib/parseCsv.ts
@@ -41,6 +41,12 @@ export function parseCsv(
   opts: CsvParseOptions = { header: true },
 ): Array<Record<string, string>> | string[][] {
   const maxRows = opts.maxRows ?? 10_000;
+  // `maxRows` caps emitted DATA rows (per CsvParseOptions). The guard
+  // below runs mid-parse against the raw row count, which in header mode
+  // includes the not-yet-stripped header row — so allow one extra raw
+  // row there. Without this the header silently counted toward the cap,
+  // turning `maxRows: N` into an `N-1` data-row ceiling in header mode.
+  const rawRowCap = opts.header ? maxRows + 1 : maxRows;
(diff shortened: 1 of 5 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #336, opened 2026-05-22
Resolve moderate npm audit findings + restore audit CI gate (#253)

hyiger:
## Summary

CI ran `npm audit --audit-level=moderate || true` — the `|| true` swallowed the exit code, so moderate+ advisories never failed the workflow. `npm audit` reported 3 moderate findings; all are now resolved and the gate is real again.

### Findings resolved
- **`brace-expansion`** ([GHSA-jxxr-4gwj-5jf2](https://github.com/advisories/GHSA-jxxr-4gwj-5jf2)) — reachable through `electron-builder` (dev dependency). Fixed by `npm audit fix`: `5.0.5 → 5.0.6`.
- **`uuid <11.1.1`** ([GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq)) — reachable only through `exceljs@4.4.0`, which declares `uuid ^8.3.0`. The advisory is a missing buffer bounds check in `v3`/`v5`/`v6` **when a `buf` arg is passed**; exceljs calls `uuid.v4()` with no args (`lib/xlsx/xform/sheet/cf-ext/cf-rule-ext-xform.js`), so it isn't actually reachable. Pinned anyway via an `exceljs → uuid ^11.1.1` override so `npm audit` goes fully clean — avoids the breaking `exceljs@3.4.0` downgrade that `npm audit fix --force` would apply.

### CI gate restored
With `npm audit` now reporting **0 vulnerabilities**, the `|| true` is dropped — a new moderate+ advisory will fail the workflow instead of silently passing.

## Test plan
- [x] `npm audit --audit-level=moderate` → `found 0 vulnerabilities` (exit 0)
- [x] exceljs write→read round-trip verified against `uuid@11.1.1` (`uuid.v4()` resolves via CJS `require`)
- [x] `npm run build` succeeds
- [x] `package.json` / `package-lock.json` in sync (`npm ci`-safe)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,8 +39,13 @@ jobs:
       - name: Run tests with coverage
         run: npm run test:coverage
 
+      # GH #253: this step previously ended in `|| true`, which swallowed
+      # the audit exit code so moderate+ advisories never failed CI. The
+      # known findings are now resolved (brace-expansion bumped; uuid
+      # pinned >=11.1.1 via an exceljs override), so the audit is a real
+      # gate again — a new moderate+ advisory will fail the workflow.
       - name: Security audit
-        run: npm audit --audit-level=moderate || true
+        run: npm audit --audit-level=moderate
 
       - name: Build
         run: npm run build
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

PR #356, opened 2026-05-24
fix: exclude parent filaments from inventory aggregates

hyiger:
## Summary

User-reported: the inventory total at the top of the filaments page counts parent filaments alongside variants and standalones, so a parent of color variants double-counts what the user actually owns. The desired model: creating a filament adds to the count; the moment it becomes a parent (≥1 non-deleted variant pointing at it), it drops out, because the parent is a template for its variants — the variants own the spools, calibrations, and remaining weight.

## Change

One new derived value at the top of `src/app/page.tsx`:

```tsx
const inventoryFilaments = useMemo(
  () => filaments.filter((f) => !f.hasVariants),
  [filaments],
);
```

Wired into every aggregate on the inventory page so the headline total never disagrees with the breakdowns:

- Header total `{N} filament(s)`
- `<FilamentStats>` props (By Type / By Vendor / Colors breakdowns)
- Quick-filter chip counts (All / Low stock / Has spools / No calibration)

Source of truth is the `hasVariants` boolean that's already shipped by the list aggregation since v1.26. The list rendering itself doesn't change — parents still appear as grouping headers with their `{N} colors` badge. This is purely a count-semantics fix, not a layout change.

## What did NOT change

- `types.length` / `vendors.length` in the header strip — those are "distinct categories I track," not inventory totals; parents share types/vendors with their variants so the distinct counts are unaffected in practice.

## Verified in dev preview

Against a DB with 52 rows and 7 parents:

- Header: `45 filament(s) · 18 type(s) · 16 vendor(s)`
- All chip: `45`
- By Type breakdown sums to 45 internally (PLA: 8, PETG: 6, PC: 5, ASA: 4, TPU: 3, ...)
- Has spools: 9 (unchanged — parents never had spools to inflate it)
- No calibration: 34 (parents-with-no-cal no longer counted)

## Tests + lint

- `npm test` — **1363 passed (1363)**.
- `npm run lint` — clean.

## Test plan

- [ ] Inventory page total drops by exactly the number of parent filaments after this lands.
- [ ] Stats panel (By Type / By Vendor / Colors) sums match the header total.
- [ ] List visual layout unchanged — parents still grouped with their color variants.
- [ ] Creating a brand-new standalone filament increases the count by 1.
- [ ] Adding the first variant to a standalone (turning it into a parent): variant counts as +1, parent flips to not-counted, net change = 0 (per user's mental model).

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -298,24 +298,37 @@ export default function Home() {
     });
   }, [fetchFilaments]);
 
+  // Inventory aggregates exclude parent filaments. Parents don't
+  // represent a physical roll on the shelf — they're a template for
+  // their color variants (the variants own the spools, calibrations,
+  // and remaining weight). Counting them in totals double-counted what
+  // the user actually has and made "By Type" / "By Vendor" disagree
+  // with the headline number. Auto-detected via the `hasVariants` flag
+  // shipped by `/api/filaments`; parents collapse out of every count
+  // here but still render in the list as grouping headers.
+  const inventoryFilaments = useMemo(
+    () => filaments.filter((f) => !f.hasVariants),
+    [filaments],
+  );
+
   // Group filaments: parents with their variants, standalone filaments as-is
   // Client-side quick filter (low stock / has spools / missing calibrations).
   // Applied before grouping so a parent whose variants are filtered out is
   // still shown standalone if it matches itself.
   const quickFilterCounts = useMemo(() => {
     const counts: Record<QuickFilter, number> = {
-      all: filaments.length,
+      all: inventoryFilaments.length,
       lowStock: 0,
       hasSpools: 0,
       noCalibration: 0,
     };
-    for (const f of filaments) {
+    for (const f of inventoryFilaments) {
       if (isLowStock(f)) counts.lowStock++;
       if ((f.spools?.length ?? 0) > 0) counts.hasSpools++;
       if (!f.hasCalibrations) counts.noCalibration++;
     }
     return counts;
-  }, [filaments]);
+  }, [inventoryFilaments]);
 
   const visibleFilaments = useMemo(() => {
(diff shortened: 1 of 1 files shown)


hyiger:
Codex P2 addressed in [`0e554da`](https://github.com/hyiger/filament-db/commit/0e554da). Non-`all` filters now resolve against `inventoryFilaments` instead of `filaments`, so the chip badge and the rendered row count agree on every filter (`noCalibration` was the case Codex flagged — a parent without calibrations would otherwise show up as a row while being excluded from the badge). `all` keeps using `filaments` because parents intentionally render as grouping headers in that view.

Verified in dev preview against a 63-row / 8-parent DB: `No calibration` chip 44, rendered rows 44 — match.

@codex review

hyiger:
Codex round-2 P2 (line 344, variant enrichment) addressed in [`c1f2bfa`](https://github.com/hyiger/filament-db/commit/c1f2bfa). New `parentLookup` built from the full `filaments` array (stable across filter state), and the variant-enrichment merge factored into `enrichVariant(v, parent)` so both the grouped-under-parent and orphaned-variant code paths apply the same fallbacks. Filtered views (`noCalibration` etc) now show variants with inherited nozzle / bed / cost / density / spool / net values even when the parent row is excluded from the filter.

The line-320 thread (chip-vs-filter mismatch) was already addressed by the previous commit `0e554da` being reviewed — the `visibleFilaments` source was switched to `inventoryFilaments` for non-`all` filters there.

Verified in dev preview against a real case: variant 'Overture PEG Black' (own nozzle = null, no calibrations) under parent 'Overture PETG' (nozzle = 240). In the `noCalibration` filter view before this commit: nozzle / bed / cost cells all `—`. After: 240°C, 80°C, $14.39 — all inherited from the parent via `parentLookup`.

@codex review

PR #358, opened 2026-05-24
fix: filter calibration printer tabs to printers that own a compatible nozzle

hyiger:
## Summary

User-reported: when editing a filament with `Bambu 0.4mm` as the only compatible nozzle, the **Calibrations** section's printer-tab row still offered **Prusa Core One** as a calibration target — even though Core One doesn't physically have that nozzle installed, so any calibration captured against (Core One, Bambu 0.4mm) is meaningless.

A nozzle lives on exactly one printer at a time (GH #232, enforced server-side by `findNozzleConflicts`), so the set of "relevant" printers for this filament can be derived directly: union of `nozzle.printers` across `form.compatibleNozzles`. The `NozzleOption.printers` reverse-mapping is already populated by `/api/nozzles`, so no extra fetch.

## Three changes in `FilamentForm.tsx`

- **New `relevantPrinters` memo** computes the intersection. The printer-tab row renders from this list instead of the full `printers` array. "Default (any printer)" always shows.
- **`effectiveSelectedPrinter` derived value** falls back to `"default"` when `selectedPrinter` lingers as a stale id outside the relevant set (e.g. user removes the compatible nozzle that was the only reason its printer tab existed). Derived rather than synced-via-effect to stay within the project's `react-hooks/set-state-in-effect` rule. Calibration-key lookup, override detection, and scope-name display all read from `effectiveSelectedPrinter`.
- **Calibration row filter** for symmetry: when a specific printer tab is selected, calibration rows for nozzles NOT installed on that printer are also hidden. The Default / Any-printer tab keeps all compatible nozzles so the user can capture baseline calibration regardless of which printer currently owns each one.

## Verified in dev preview

Test filament with `compatibleNozzles = [Bambu 0.4mm]`:
- Printer tab row: **Default (any printer)** + **Bambu Labs H2D** only. Prusa Core One and Codex QA Printer no longer offered.
- H2D tab: 1 calibration card for Bambu 0.4mm.
- Default tab: 1 calibration card for Bambu 0.4mm (baseline).

## Tests + lint

- `npm test` — **1363 passed (1363)**.
- `npm run lint` — clean.

## Test plan

- [ ] Filament with one compatible nozzle on one printer: only that printer tab + Default offered.
- [ ] Filament with compatible nozzles spanning two printers: both printer tabs + Default offered; clicking each printer's tab shows only that printer's nozzle row.
- [ ] Default / Any-printer tab: every compatible nozzle row appears (baseline calibration).
- [ ] Remove a compatible nozzle that was the only reason a printer tab existed → tab disappears and any selection of that printer falls back to Default automatically.
- [ ] Filament with no installed-nozzle matches across any printer (e.g., compatible nozzle that's been retired): no printer tabs row appears, Default is implicit; calibration rows still render under the implicit Default scope.

### DIFF
--- a/src/app/filaments/FilamentForm.tsx
+++ b/src/app/filaments/FilamentForm.tsx
@@ -763,9 +763,22 @@ export default function FilamentForm({ initialData, onSubmit, onDirtyChange }: P
             // (printerId === "default") is always kept — it's the
             // baseline that applies regardless of which printer has
             // the nozzle at the moment.
+            //
+            // FAIL-OPEN when ownership data isn't available (codex
+            // round-2 P1 on PR #358). The catalog is fetched async via
+            // `/api/nozzles`; if the user saves while it's still
+            // loading — or if the fetch failed — `nozzles` is empty,
+            // the lookup returns `undefined`, and a strict predicate
+            // would silently delete every valid per-printer
+            // calibration. Treat absence of ownership data as
+            // "uncertain" and keep the entry; only drop when we have
+            // positive evidence (catalog loaded, nozzle found,
+            // printers populated, and the printer is not in the
+            // installed list).
             if (printerId === "default") return true;
             const nozzle = nozzles.find((n) => n._id === nozzleId);
-            return nozzle?.printers?.some((p) => p._id === printerId) ?? false;
+            if (!nozzle || !nozzle.printers) return true;
+            return nozzle.printers.some((p) => p._id === printerId);
           })
           .map(([key, cal]) => {
             const [printerId, nozzleId, bedTypeId] = key.split(":");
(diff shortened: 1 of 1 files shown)


hyiger:
Codex round-1 P2 addressed in [`c8b327b`](https://github.com/hyiger/filament-db/commit/c8b327b). The save-time `calibrations` filter in `handleSubmit` already dropped entries whose nozzle was no longer compatible; it now also drops entries where the printer is non-`default` AND no longer owns the nozzle. The `default` baseline scope is preserved.

Verified end-to-end against a seeded filament with `compatibleNozzles=[Bambu 0.4mm]` and two calibration entries — `(H2D, Bambu 0.4mm)` valid + `(Core One, Bambu 0.4mm)` stale. Edit form correctly hides the Core One tab; clicked Update Filament; `GET /api/filaments/{id}?raw=true` afterward returned just the H2D calibration. The stale Core One entry was pruned.

@codex review

PR #359, opened 2026-05-24
fix: clear phantom tagPresent when auto-read connect fails

hyiger:
## Summary

**User-reported**: status pill stuck at "Tag detected" with no tag on the reader. Unplug + replug doesn't help — the same stuck state recurs on every replug.

## Root cause

`SCARD_STATE_PRESENT` can fire spuriously on some PC/SC drivers (documented for the ACS driver on the ACR1552U on macOS, GH #230). The reader-status handler in `electron/nfc-service.ts` already skips the FIRST event per reader to dodge the plug-in phantom — but a later event with `SCARD_STATE_CHANGED` toggled can still set `tagPresent: true` without a real tag in the field. After that flip, the service has no recovery path:

- No subsequent status event clears it — the driver thinks `present=true` is current.
- Unplug + replug resets state momentarily, but the same phantom runs on every replug and re-sticks `tagPresent: true`.

## Fix

When the auto-read in `electron/main.ts` fails with the `"Cannot connect to tag"` error (the connect retries having exhausted ~6s), treat the present bit as phantom and call a new **`NfcService.clearPhantomPresence()`** method:

- Resets `readerPresent` to an empty map.
- Emits a corrective `tagPresent: false` status if the service was carrying `tagPresent: true`.

The renderer pill recovers to "Ready, place a tag" once the corrective status update propagates.

## Behavior

- **Real tag, normal read** — unchanged. `connect()` succeeds, `tagPresent: true` stays, decode runs.
- **Phantom present, no tag** — old: pill stuck at "Tag detected" forever. New: pill briefly shows "Tag detected" (during the ~6-second connect retry window), then recovers to "Ready, place a tag" once the connect failure propagates.
- **Real tag with transient connect issue** — same path as phantom: clear-and-recover. If the tag is genuinely there, the next physical perturbation (lift + replace) generates a fresh status event and the auto-read tries again.

## Verification

Cannot reproduce end-to-end in the browser dev preview — this codepath requires the Electron renderer + a real PC/SC reader to exercise the phantom-present transition. The fix is small, self-contained, and the error string match is exact (`"Cannot connect to tag"`). It will be exercised on a packaged build with the reader hardware as part of the test plan.

Full test suite **1363/1363** green, lint clean.

## Docs

`CLAUDE.md` gains a note alongside the existing SLIX2 caveat documenting the phantom-present recovery path.

## Test plan

- [ ] In a packaged Electron build with the ACR1552U (or any ACS reader) plugged in, no tag on it: pill shows "Ready, place a tag". If it momentarily flashes "Tag detected" then recovers — that's the corrective clear working.
- [ ] Place a real tag → pill goes "Tag detected" → decoded → "Loaded: <name>".
- [ ] Lift the tag → pill goes "Ready, place a tag".
- [ ] Unplug + replug reader with no tag: pill should NOT get stuck at "Tag detected".
- [ ] Real tag with deliberate read-failure (e.g., remove tag mid-read): pill recovers to "Ready" rather than sticking on the failure.


hyiger:
Codex P2 addressed in [`d2f3a74`](https://github.com/hyiger/filament-db/commit/d2f3a74). `clearPhantomPresence()` no longer calls `readerPresent.clear()` — it only force-clears the service-level `tagPresent` flag. The per-reader map is left alone so subsequent status events keep maintaining it organically; another reader that legitimately had a tag won't have its presence wiped by the corrective clear of a different reader's phantom. CLAUDE.md note updated to reflect the narrower behavior.

@codex review

PR #386, opened 2026-05-25
feat: round-trip Parent column on filament CSV/XLSX import (closes #379)

hyiger:
## Summary

- Follow-up to [#378](https://github.com/hyiger/filament-db/pull/378), which added `Parent` and `Variant Count` columns to the filament-level export. The re-import path didn't read them, so a round-trip flattened every variant into a standalone filament — [#379](https://github.com/hyiger/filament-db/issues/379).
- Shared filament importer (`src/lib/importFilaments.ts`, used by both `/api/filaments/import-csv` and `/import-xlsx`) now recognises the `Parent` column. Honoured **only on create/resurrect**; silently ignored on update because re-parenting via re-import is a surprising UX. `Variant Count` is read-only and ignored.
- Two-pass driver: rows without `Parent` go first, so an in-batch parent is present in the lookup map by the time the variant rows resolve. Works regardless of CSV row order (real exports sort by name, so variants often precede their parents alphabetically).
- Parent must be active and must not itself be a variant. Self-references are rejected. All failure modes land in `skippedRows` with a named reason.

## Scope note

Issue title says "Spool CSV importer" but the substantive round-trip described is filament-level (export filaments → reimport → cluster preserved). The spool importer requires filaments to exist already and doesn't create them, so the fix belongs in the filament importer. Spool importer is unchanged — extra `parentName` columns in a spool CSV are silently ignored as they were before.

## Test plan

- [x] `npm test` — 1410 tests pass (8 new in `tests/importFilaments.test.ts` covering header mapping, in-batch round-trip in either row order, missing/variant/self-referential parent, and the update-ignores-Parent rule)
- [x] `npm run lint`
- [x] `npx tsc --noEmit`
- [ ] Manual: export filaments with variants → delete all → re-import → confirm cluster preserved

### DIFF
--- a/public/openapi.json
+++ b/public/openapi.json
@@ -1397,7 +1397,7 @@
       "post": {
         "tags": ["Filaments"],
         "summary": "Import filaments from CSV",
-        "description": "Upload a CSV file with header row. Requires Name, Vendor, Type columns. Only fields present are updated.",
+        "description": "Upload a CSV file with header row. Requires Name, Vendor, Type columns. Only fields present are updated. The optional `Parent` column (matching the filament CSV export) attaches a newly-created filament as a variant of the named parent — honoured on create/resurrect only; ignored on update. The parent must exist among active filaments and must not itself be a variant. `Variant Count` is read-only and ignored on import.",
         "requestBody": {
           "required": true,
           "content": {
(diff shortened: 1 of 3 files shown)


hyiger:
Addressed Codex's P2 in e0391b4. Extracted a `trimmedParentName` helper that both the two-pass router and `processRow`'s parent resolution now share — a whitespace-only `Parent` cell now routes to pass 1 (same as an empty cell) instead of being delayed into pass 2. Added a regression test exercising the exact ordering described (variant row first, whitespace-parent row second).

hyiger:
@codex review

PR #390, opened 2026-05-29
Clarify Clone vs Create variant (#388)

hyiger:
## Summary

User reported that the two buttons on a filament's detail page both appeared to produce the same outcome and the tooltips didn't explain what was different.

The **structural outcome is identical** — both create a variant under the same parent. The actual difference is which fields get pre-filled on the new-filament form:

| Button | Pre-fills | When to use |
|--------|-----------|-------------|
| Clone (`?cloneId=`) | name (with "(Copy)" suffix), color, colorName, vendor, type | "I want the **same color** again, slightly different" |
| Create variant (`?parentId=`) | vendor + type only | "I want a **different color** of the same material" |

## Fix

Rather than rework behavior (PR #377 deliberately allowed Clone on variants to make sibling-creation discoverable), tighten the label + tooltip on each button so the user can tell them apart at a glance:

- **Clone → Duplicate** (clearer English; "clone" implied a wholesale copy that confused the parent/variant relationship)
- **Duplicate tooltip**: *"Duplicate this filament — the new variant starts with the **SAME** color and a copy of the name."*
- **Create variant tooltip**: *"Add a new color variant — inherits this filament's settings, but you pick a **DIFFERENT** color."*

The capitalised **SAME** / **DIFFERENT** contrast is what makes the choice obvious.

For consistency, also renamed the page title on `/filaments/new?cloneId=` ("Clone Filament" → "Duplicate Filament") and the new-page picker section ("Clone Existing" → "Duplicate Existing"). No code changes — pure i18n.

## What's NOT changed

- Underlying routing: still `?cloneId=` and `?parentId=` (renaming would break bookmarks / NFC links / docs)
- Behavior: PR #377's intentional choice to show Duplicate on variants stays
- Key counts: en.json and de.json both at 1211

## Test plan
- [x] No tests reference the literal "Clone" string
- [x] Both locale files parse and have matching key counts


hyiger:
@codex review

PR #391, opened 2026-05-29
Add Spool Inventory page (#389)

hyiger:
Resolves #389 (pulled forward from #127 task 5).

Companion to the filament list at `/`: same data, opposite lens. Where the filament list groups spools UNDER their filament, `/inventory` groups filaments under their LOCATION — so users can audit a shelf, see at a glance where everything lives, and update the common per-spool things without bouncing through each filament's detail page.

## What you can do on the new page

- **See it all at once** — stats header (total spools, location count, total active grams) above collapsible groups per location, each with its own summary chip
- **Filter** — search box (filament/label/lot), location kind dropdown, type/vendor text inputs, "include retired" toggle
- **Edit inline** — click the weight cell to update from a scale; pick a new location from the move-to dropdown; toggle retire with a one-click button (confirms on retire)
- **Click through** — every filament name links to its detail page

## API

`GET /api/spools/by-location` — single-shot aggregation over `Filament.spools[]`. Groups by `locationId`, lookups the Location doc, surfaces variant inheritance hints (parent `spoolWeight` / `netFilamentWeight`) so the client can compute `% remaining` without a second fetch.

Query params: `kind`, `type`, `vendor`, `includeRetired=1`. Spools whose `locationId` is null land in a synthetic "no location" group sorted to the END of the response — surfaces stragglers without burying real shelves.

## Inline edits use existing endpoints

Weight update, location move, retire/unretire all go through `PUT /api/filaments/{id}/spools/{spoolId}` — the same endpoint the SpoolCard on the filament detail page uses. No new mutation route needed and the semantics stay identical across both surfaces.

## What's NOT in this PR

Out of scope for v1, opened space for follow-ups if you want them:

- Bulk operations (select N spools → move them all to drybox X)
- Multi-sort within a group (current sort: filament name, then label, ascending)
- Photo gallery view (we have `photoDataUrl` on each spool but this page is table-only)
- Print-history snippet per spool (which jobs used this spool last?)

## Tests

`tests/spools-by-location-route.test.ts` — 9 route-level cases pinning the aggregation behaviour: groups by locationId, "no location" group at the end, kind/type/vendor filters, retired toggle, dryCycle count + lastDryAt, parent inheritance hints, soft-delete exclusion.

## i18n

37 new keys (en.json + de.json, parity preserved at 1248).

## Test plan
- [x] `npx vitest run tests/spools-by-location-route.test.ts` → 9/9
- [x] `npm run lint` + `npx tsc --noEmit` clean
[...]

PR #400, opened 2026-05-30
Fix inventory totals + honor INHERITABLE_FIELDS on filters (follow-up to #391)

hyiger:
Follow-up to PR #391 — addresses three P2 findings from Codex's post-merge review that didn't make it into the original PR.

## What's wrong on main today

1. **Inventory totals over-report by `N × empty-spool-mass`** — the aggregation sums gross `spools.totalWeight` (filament + tare) instead of remaining filament. The header and per-location summaries don't agree with the row-level `%`-remaining bars or with the existing `getRemainingGrams` helper used elsewhere.
2. **Filters miss variants that inherit `type`/`vendor` from a parent** — `?type=PLA` and `?vendor=Polymaker` match the variant's own raw value before the parent `$lookup` runs, so a variant that inherits either field is dropped from filtered results even though the rest of the app treats it as that type/vendor.
3. **Search filter on the page leaves group totals stale** — when the text search narrows a 20-spool shelf to 2 matches, the cloned group keeps `count` and `totalGrams` from the unfiltered server response. Header renders "20 spools · 18000g" above two rows.

## Fixes

- **`totalGrams` now sums `max(0, totalWeight − effectiveTare)`** per spool, where `effectiveTare = spoolWeight ?? parent.spoolWeight ?? null`. When either input is null the row contributes 0 — matches `getRemainingGrams`'s null posture (a filament without `spoolWeight` is excluded because we can't separate filament from packaging).
- **Parent `$lookup` moved BEFORE the type/vendor `$match` stages**, and its projection now includes `type` and `vendor`. Both filters now match on `$ifNull("$type", "$_parent.type")` (same for vendor). The spool row projection also returns the EFFECTIVE values so the page's chips don't render blank for an inheriting variant.
- **Client-side search filter recomputes** `count` and `totalGrams` from the filtered `spools[]` using the same `remainingGrams` helper the rows use.

## Tests

11/11 pass. Updated the existing "groups by location" test to assert post-tare values (drybox 700g not 1900g, shelf 900g not 1100g). Added two new tests:

- **Tare regression**: a variant filament with no `spoolWeight` of its own + a parent that has it — aggregation must reach through the self-`$lookup` and subtract the parent's tare. Asserts 1300g vs the pre-fix 1800g.
- **Inheritance regression**: `?type=PLA` matches a variant created via `Filament.collection.insertOne` (bypasses Mongoose required-field validation) with no own `type`/`vendor` but a parent that has both. Confirms the inheriting variant lands in the result AND that the response carries the effective values.

## Test plan

- [x] `npx vitest run tests/spools-by-location-route.test.ts` → 11/11

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -802,7 +802,6 @@
       "version": "1.10.0",
       "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
       "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
@@ -1646,7 +1645,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1669,7 +1667,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1692,7 +1689,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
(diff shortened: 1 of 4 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #464, opened 2026-05-30
Sync: tolerate concurrent E11000 races + document subdoc-id caveat (#438, #439)

### LINKED ISSUE #439: Sync: local-only push can E11000 on syncId when two processes sync the same Atlas
## Summary
`stripForTransfer` drops `_id` but preserves `syncId` on the local-only push path. If two simultaneous sync cycles run on the same Atlas — e.g. desktop client and a Docker instance both syncing to the same cluster — both can pass the "local-only → push" branch concurrently and try to `insertOne` the same `syncId`. The `syncId` index is `unique, sparse` on every synced model, so the second insert hits E11000 and `trySync` reports the whole collection as failed.

The in-process `this.syncing` guard prevents this on a single SyncService instance, but there's no shared lock across processes.

## Evidence
- `electron/sync-service.ts:594, 600` — local-only insert path.
- e.g. `src/models/Printer.ts` — `syncId: { type: String, unique: true, sparse: true }`.

## Suggested fix
Change the `insertOne` to an upsert keyed on `syncId` (idempotent), or catch `E11000` in the insert path and treat as a no-op since the doc already exists on the target.

hyiger:
## Summary
- **#439** — Local-only push and remote-only pull paths in `syncCollection` now treat a `code: 11000` duplicate-key error as a no-op (concurrent peer won the `syncId` race). New `isDuplicateKeyError` helper, exported for testing.
- **#438** — Extended the SyncService header comment to cover the `calibrations[]._id` / `amsSlots[]._id` cross-side regeneration caveat (today nothing keys off these subdoc ids cross-sync; adding a future feature that does will need a subdoc `syncId` migration).

## Tests
- New unit tests for `isDuplicateKeyError` covering Error-with-code, plain-object-with-code, wrong code, no code, and non-error inputs.
- All 17 sync-service tests pass.

Closes #438
Closes #439

### DIFF
--- a/electron/sync-service.ts
+++ b/electron/sync-service.ts
@@ -3,6 +3,25 @@ import { randomUUID } from "crypto";
 import { MongoClient, ObjectId, Document } from "mongodb";
 
 /**
+ * Recognise the MongoDB driver's duplicate-key error so the local-only
+ * push / pull paths can treat a concurrent peer winning the
+ * `syncId`-unique race as a no-op rather than a sync failure
+ * (GH #439). The driver surfaces this as `MongoServerError` with
+ * `code === 11000`; `MongoBulkWriteError` from `insertMany` carries
+ * the same code on its `writeErrors[]` array, but the local-only
+ * paths here use `insertOne`, so just checking the top-level code
+ * is sufficient.
+ */
+export function isDuplicateKeyError(err: unknown): boolean {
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

hyiger:
@codex review

PR #465, opened 2026-05-30
Cap scan-stream SSE concurrent subscribers (#428)

hyiger:
## Summary
Adds a 100-subscriber concurrent cap on the `/api/scan/stream` SSE endpoint, returning 429 + `Retry-After: 60` when exceeded. The endpoint is intentionally non-origin-guarded so PrusaSlicer / Bambu Studio can subscribe over the LAN, but the underlying `scanBus` sets `maxListeners(0)` — so a misbehaving / hostile client could open thousands of long-lived connections.

## Tests
- New regression test: open 100 connections, verify 101st returns 429 with `Retry-After: 60`, verify slot frees on cancel.
- All 7 scan-stream tests pass.

Closes #428

### DIFF
--- a/src/app/api/scan/stream/route.ts
+++ b/src/app/api/scan/stream/route.ts
@@ -29,6 +29,23 @@ export const dynamic = "force-dynamic";
 const HEARTBEAT_MS = 25_000;
 const encoder = new TextEncoder();
 
+/**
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #469, opened 2026-05-30
Electron: updater gates + Mongo rollback + NFC handle/CC safety (#433, #435, #436, #437)

### LINKED ISSUE #435: Electron: startLocalMongo leaves dead state if URL parsing throws after create()
## Summary
`MongoMemoryServer.create()` sets `mongod`, then `mongod.getUri()` and `new URL(uri)` run. If `new URL(uri)` throws (malformed URI — unlikely but possible), `mongod` is left set with `uri` still null. The next `startLocalMongo()` call hits the `if (mongod && uri) return uri` guard which is false (uri is null), so it re-enters `MongoMemoryServer.create()` while a previous instance is still running — port conflict, second mongod orphaned.

## Evidence
- `electron/local-mongo.ts:31-49` — the create + URL-parse block.
- `electron/local-mongo.ts:14` — the early-return guard.

## Suggested fix
```ts
try {
  uri = mongod.getUri();
  const parsed = new URL(uri);
  ...
} catch (err) {
  await mongod.stop().catch(() => {});
  mongod = null;
  uri = null;
  throw err;
}
```

### LINKED ISSUE #436: Electron NFC: reader handles not released on retry-path connect() failure
## Summary
In `connect()`'s retry loop, when a previous attempt set `this.activeReader` but a later loop iteration overrides it, the previous reader handle isn't released. `disconnectReader` exists but is never called in this path. `withConnection`'s `finally` calls `this.disconnect()` which only releases the *current* `activeReader`. PC/SC handles are scarce on Linux `pcscd`; the leak shows up as "no readers" after a few reconnect cycles.

## Evidence
- `electron/nfc-service.ts:237-291` — `connect()` retry loop.
- Same file:226 — `disconnectReader` exists but isn't called for previous-attempt readers.

## Suggested fix
In the retry loop, when overriding `this.activeReader`, call `this.disconnectReader(prev)` first. Also call `disconnectReader` on failed-attempt readers inside `trySharedConnect`'s fallback path.

### LINKED ISSUE #437: Electron NFC: writeTag/formatTag clobber non-OpenPrintTag tags without 0xE1 CC check
## Summary
The write and format paths overwrite block 0 without verifying the tag carries an NFC-Forum Type 5 CC byte (`raw[0] === 0xE1`). The read path checks this; the write paths don't. A user with a non-blank tag of a different format (proprietary, RFID inventory tag, transit card) in the field at the moment they trigger Format/Write gets its block 0 overwritten — potentially bricking it for its original use.

## Evidence
- `electron/nfc-service.ts:489-533` — `writeTag`.
- `electron/nfc-service.ts:535-574` — `formatTag`.
- `electron/ndef.ts:252` — `0xE1` check exists on the read path.

## Suggested fix
Read block 0 in both `writeTag` and `formatTag` first and refuse with a user-facing error if `block0[0] !== 0xE1`. The renderer can prompt the user before retrying with a `forceOverwrite: true` option for the rare legitimate "bring a blank tag I formatted out-of-band" case.

hyiger:
## Summary
- **#433** `update-check` IPC + 6-hourly periodic check early-return when state is `downloading` / `ready`.
- **#435** `startLocalMongo` URI-parse block wrapped in try/catch that rolls back `mongod` + `uri` on throw.
- **#436** PC/SC handle leaks: failed `trySharedConnect` releases any partial handle; retry loop tracks every reader it claimed and releases them all on hand-off / final throw.
- **#437** `writeTag` + `formatTag` refuse the operation when block 0 isn't 0xE1 (CC) or 0x00 (blank).

Closes #433
Closes #435
Closes #436
Closes #437

### DIFF
--- a/electron/auto-updater.ts
+++ b/electron/auto-updater.ts
@@ -79,6 +79,15 @@ export function initAutoUpdater(win: BrowserWindow) {
 
   ipcMain.handle("update-check", async () => {
     if (!app.isPackaged) return { ok: false, error: "dev-mode" };
+    // GH #433: don't re-check while a download is in flight or an
+    // update is sitting ready to install. `autoUpdater.checkForUpdates()`
+    // re-emits `update-available` / `update-not-available`, which would
+    // overwrite the in-progress `downloading` / `ready` state in
+    // `currentState` and blow away the progress UI. The download itself
+    // continues internally but the renderer briefly sees a stale state.
+    if (currentState.state === "downloading" || currentState.state === "ready") {
+      return { ok: true, skipped: currentState.state };
+    }
     try {
       await autoUpdater.checkForUpdates();
       return { ok: true };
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #473, opened 2026-05-30
Bambu inheritance preservation + FilamentForm negatives (#403, #406)

### LINKED ISSUE #403: Bambu Studio per-id POST severs variant field inheritance
## Summary
The per-filament Bambu Studio sync route stamps every parsed field onto the target document even when it's a variant currently inheriting those fields from its parent. After one sync, density/cost/diameter/temperatures/etc. are pinned on the variant — silently severing inheritance for every field the Bambu profile carries.

## Evidence
- `src/app/api/filaments/[id]/bambustudio/route.ts:165-189` — `buildStructuredUpdate(parsed.filament, existing)` writes type/vendor/color/density/diameter/cost/maxVolumetricSpeed/temperatures directly onto the variant via `$set`. The route never consults `existing.parentId` or `resolveFilament`.
- Same class as #106 / #223 / #265 which the rest of the codebase guards against.
- `POST /api/filaments/[id]` (PrusaSlicer sync) at lines 322-489 handles this carefully; the Bambu route does not.

## Suggested fix
Detect `existing.parentId` and either route inheritable scalars to the parent (matching the PrusaSlicer-sync calibration-target rule) or drop fields whose existing variant value is `null` AND parent value already matches, so the variant continues to inherit.

hyiger:
## Summary
- **#403** `buildStructuredUpdate` is now variant-aware. When `existing.parentId` + `existing.parent` are provided, inheritable scalars whose parsed value matches the parent's are SKIPPED so inheritance keeps resolving dynamically via `resolveFilament`. Both the per-id and the 3-phase bulk Bambu routes load the parent and pass it through.
- **#406** `min="0"` / `max="100"` on FilamentForm fan / retraction / loading / calibration inputs. Server validators on calibrations[] / fanMin/Max/BridgeSpeed already exist; the remaining settings-bag fields stay UI-gated per the issue's documented limitation.

## Tests
- 5 new tests in `tests/bambuStudioApply.test.ts` pinning the inheritance-preserve behavior.
- All 19 existing bambu-route tests pass.

Closes #403
Closes #406

### DIFF
--- a/src/app/api/filaments/[id]/bambustudio/route.ts
+++ b/src/app/api/filaments/[id]/bambustudio/route.ts
@@ -180,6 +180,18 @@ export async function POST(
     }
 
     const existingWithParent = {
+      // Codex P1 on PR #473 round 2: inheritable scalars MUST ride
+      // along so `buildStructuredUpdate` can detect a stale variant
+      // override and emit $unset. Pre-fix these were stripped, leaving
+      // the unset branch unreachable in this route.
+      type: existing.type ?? null,
+      vendor: existing.vendor ?? null,
+      diameter: existing.diameter ?? null,
+      density: existing.density ?? null,
+      cost: existing.cost ?? null,
+      maxVolumetricSpeed: existing.maxVolumetricSpeed ?? null,
+      shrinkageXY: existing.shrinkageXY ?? null,
+      shrinkageZ: existing.shrinkageZ ?? null,
       temperatures: existing.temperatures as Record<string, unknown> | undefined,
       bedTypeTemps: existing.bedTypeTemps,
       settings: existing.settings as Record<string, unknown> | undefined,
--- a/src/app/api/filaments/[id]/bambustudio/route.ts
+++ b/src/app/api/filaments/[id]/bambustudio/route.ts
@@ -167,7 +167,28 @@ export async function POST(
       return errorResponse("Filament not found", 404);
     }
 
-    const update = buildStructuredUpdate(parsed.filament, existing);
+    // GH #403: when `existing` is a variant, load its parent so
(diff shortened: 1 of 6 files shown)


hyiger:
@codex review

hyiger:
@codex review — addressed your P1 finding on variant override clearance. `buildStructuredUpdate` now returns `{ set, unset }`; when the parsed Bambu value matches the parent AND the variant has a diverging local override, the field is added to `unset` so a Mongo `$unset` clears the stale value and inheritance resumes via `resolveFilament`. Bulk + per-id routes both compose `$set` + `$unset` into one atomic write. Six new tests pin: stale override clearance, three no-op safety cases, $set-wins-on-divergence, root-filament guard, and empty-string-as-inheriting parity.

hyiger:
@codex review — addressed round 2 P1. The route augment helpers (bulk + per-id) were stripping the inheritable scalars that `buildStructuredUpdate` needs to detect stale variant overrides — the unit test passed against a richer hand-built shape so the regression slipped through. Both helpers now thread type/vendor/diameter/density/cost/maxVolumetricSpeed/shrinkageXY/shrinkageZ through. New route-level integration test in tests/bambustudio-route.test.ts creates a real parent + variant with diverging density and asserts the variant's density is unset in MongoDB after a parent-matching sync.

hyiger:
@codex review — addressed P2 r3. `buildStructuredUpdate` now treats `type` and `vendor` as required and leaves their stale variant overrides in place rather than queueing them for $unset (which would fail schema validation under `runValidators: true`). Optional fields keep the existing unset-on-revert path. Two new tests pin the behaviour: unit-level (stale type+vendor+density → only density in unset) and route-level (stale type+vendor+density round-trip through the real route → 200, type+vendor stay pinned, density cleared).

Issue #450, opened 2026-05-30
NFC status pill: silent on permission-denied / reader-busy states (macOS)


hyiger:
## Summary
The NFC status pill shows three states: no reader / ready / tag detected. There's no "PC/SC permission denied" or "reader busy with another app" state surfaced. On macOS without granted permissions or with another app holding the reader, the pill stays at "No reader connected" — the user has no actionable hint about what to fix.

## Evidence
- `src/components/NfcStatus.tsx:22-40` — state set is limited to the three above.

## Suggested fix
Surface `status.lastError` from the NfcService (when set) as a muted second line under the pill, with a link to the Smart Filament Workflow guide. Translate the common error shapes (permission denied, busy, no PC/SC daemon) to plain language.

Filed from full-codebase review on 2026-05-29.

hyiger:
Fixed by PR #475 / #476.

Issue #420, opened 2026-05-30
Inventory page lacks batch actions for shelf-audit workflow


hyiger:
## Summary
The Inventory page's stated purpose is "audit a shelf or drybox at a glance and update common per-spool fields". But its rows have no selection checkbox and there is no header / footer batch action bar. Moving 8 spools off a shelf one-by-one defeats the use case the page was built for.

## Evidence
- `src/app/inventory/page.tsx:402-411` — `<SpoolEditRow>` rows are individual without selection state.
- `src/app/page.tsx:643-647` already has the selection-checkbox pattern that can be reused.

## Suggested fix
- Add a row checkbox + a sticky "Move N spools to…" / "Retire N" action bar.
- Hit the existing `PUT /api/filaments/{id}/spools/{spoolId}` endpoint per selected row (no new mutation route needed).
- Honour the same retired-spool exclusion in the move dropdown.

Filed from full-codebase review on 2026-05-29.

hyiger:
Fixed by PR #475 / #476.

PR #476, opened 2026-05-30
Inventory batch actions + NFC error states (#420, #450)

hyiger:
## Summary

Two related issues from the v1.32.3 audit, both about a UI surface failing to expose state the user needs.

### #420 — Inventory batch actions

The Inventory page's stated purpose is *\"audit a shelf at a glance and update common per-spool fields\"*. Pre-fix, every move and retire was a single-row action — moving 8 spools off a shelf meant 8 individual clicks/dropdowns.

- Per-row checkbox column with a group-level select-all (indeterminate / checked / unchecked).
- Sticky action bar appears when ≥1 row is selected:
    1. **Move N to…** dropdown (every Location + a \"no location\" sentinel)
    2. **Retire N** / **Unretire N** (auto-detects from selected rows' state)
- Selection set keyed by `filamentId:spoolId` so it survives filter changes; the bar count reflects visible-AND-selected so the user can't accidentally write to rows that scrolled out of view.
- Batch writes go through the existing per-spool `PUT /api/filaments/{id}/spools/{spoolId}` — no new mutation route. Sequential (avoids thundering-herd retries on slow LANs). Partial-success is surfaced explicitly via toast (\"3 of 5 updated, 2 failed\").
- Destructive retire goes through `ConfirmDialog`.

### #450 — NFC permission / busy / no-daemon hints

The NFC status pill fell back to *\"No reader connected\"* on macOS when Smart Card permission was denied or another app (CryptoTokenKit, vendor utility) held the reader. No actionable hint about what to fix.

- New `classifyNfcError` in `src/lib/nfcErrorClassify.ts` maps opaque pcsclite messages → `permission` / `busy` / `no-daemon` / `generic`. Substring-matched on the lowercased message so driver wording variants don't drop into the generic fallback.
- NfcService records the most recent classified error on `status.lastError` and clears it on the next successful reader status event — a transient hiccup doesn't stick the pill in an error state.
- NfcStatus.tsx renders the code as a translated, plain-language hint (red dot, error takes precedence over reader/tag state). Tooltip surfaces the raw upstream message for diagnostics.
- `classifyNfcError` lives in `src/lib/` (not co-located with NfcService in `electron/`) so the vitest config — which excludes `electron/` — can exercise it. nfc-service.ts re-exports it for the existing call-sites.

### Tests
- 18-case unit test pinning every classification bucket plus priority-order ties (permission > busy > no-daemon) — `tests/nfcErrorClassify.test.ts`.
- New `inventory.batch.*` and `nfc.error.*` keys in EN+DE; locales stay at parity (1255 keys each).

## Test plan
- [x] `npx vitest run tests/nfcErrorClassify.test.ts` — 18/18 pass
- [x] `npm run lint` — clean
- [x] `npm test` — 1525/1525 pass
- [ ] @codex review


hyiger:
@codex review

hyiger:
@codex review — addressed both P2 findings:

1. **service not available wording**: classifier now matches `service not available` / `service unavailable` (the canonical macOS `SCardEstablishContext: Service not available` message). Round 1 only checked `no service` as a substring so the actual message dropped through to generic. +1 test pins the new wordings.

2. **batch loop state reset on rejection**: wrapped the per-row loop in try/finally so a rejected fetch (LAN drop) can't leave `batchBusy` true with the selection stuck. Per-row updateSpool is now in its own try/catch so a single rejection counts as a failed row in the aggregate toast rather than aborting the batch. Refresh is best-effort `.catch(() => {})` after the reset.

hyiger:
@codex review — addressed P2 r2. `visibleSelectedRows` now skips groups present in the `collapsed` Set, so collapsing a group with selected rows drops them from the action-bar tally and prevents the hidden-write path. The underlying `selectedKeys` set is unchanged — re-expanding restores the original selection (lossless round-trip).

hyiger:
@codex review — addressed P2 r2: `withConnection` now calls `clearLastError()` on the success path. A completed tag round-trip proves the reader + PC/SC daemon are healthy, so a lingering error from a transient busy/permission state is cleared. All three tag-ops (readTag/writeTag/formatTag) flow through withConnection so the single fix catches them all.

PR #478, opened 2026-05-30
Add Release Bump workflow

hyiger:
## Summary

Adds `.github/workflows/release-bump.yml` — a `workflow_dispatch` workflow that automates the per-release version bump (package.json + package-lock.json + public/openapi.json) so the bump commit can land via PR instead of a direct push to main.

## Why

The new branch protection rule on `main` requires every change to land via PR with green CI. The release process per `CLAUDE.md` previously required a direct push for the version bump commit, which would now need admin bypass. This workflow removes that exception — the bump goes through the same gate as every other change.

## How to use

```bash
gh workflow run release-bump.yml -f version=1.32.5
```

Or via the Actions tab → **Release Bump** → **Run workflow**.

The workflow:
1. Validates the version input (semver, not already used, tag doesn't exist)
2. Creates a `bump/v<version>` branch
3. Updates the three version-bearing files
4. Commits + pushes
5. Opens a PR with auto-merge enabled (squashes once CI passes)
6. Writes a summary telling you to pull main + push the tag

After the PR auto-merges:
```bash
git checkout main && git pull
git tag v<version>
git push origin v<version>
```

The tag push triggers the existing `release.yml` build matrix as before.

## Test plan
- [ ] After merge, try `gh workflow run release-bump.yml -f version=1.32.5-test` on a throwaway version to confirm the flow works end-to-end before the next real release
- [ ] Verify the bump PR's CI runs and auto-merges
- [ ] Verify the `npm install --package-lock-only` step doesn't churn unrelated lockfile entries

### DIFF
--- a/.github/workflows/release-bump.yml
+++ b/.github/workflows/release-bump.yml
@@ -14,24 +14,24 @@ name: Release Bump
 #   gh workflow run release-bump.yml -f version=1.32.5
 # Or via the Actions tab → Release Bump → Run workflow.
 #
-# ── PAT REQUIREMENT ──────────────────────────────────────────────────
-# Codex P2 on PR #478: PRs created or updated by the default
-# GITHUB_TOKEN start their `pull_request` workflow runs in an
-# approval-required state. That means each bump PR would sit waiting
-# for a maintainer to click "Approve and run workflows" before CI can
-# fire, breaking the auto-merge promise.
+# ── PAT REQUIREMENT (mandatory) ──────────────────────────────────────
+# Codex P1 on PR #478 round 2: events triggered by GITHUB_TOKEN do NOT
+# create new workflow runs (except for workflow_dispatch /
+# repository_dispatch). That means if this workflow opened the bump PR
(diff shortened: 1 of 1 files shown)


hyiger:
@codex review

hyiger:
@codex review — addressed P2. Workflow now prefers `secrets.RELEASE_BUMP_TOKEN` (user-provided fine-grained PAT) and falls back to GITHUB_TOKEN with explicit notice/warning telling the user that manual approval will be needed on the resulting PR's CI. Header comment documents the one-time PAT setup. PAT path → fully unattended; fallback path → one click per bump.

hyiger:
@codex review — addressed P1. You're right that GITHUB_TOKEN-triggered events don't queue pull_request runs at all, not just in approval-required state. Workflow now hard-requires `secrets.RELEASE_BUMP_TOKEN` with a fail-fast step that prints the scopes + Settings path, and drops every `|| secrets.GITHUB_TOKEN` fallback. The fallback path I added in round 1 was silently broken — thanks for catching it.

PR #484, opened 2026-05-31
Multi-color phase 3: NFC + tag-based importers (#477)

hyiger:
## Summary

Third of 5 phases for multi-color filament support. **Multi-color data now flows through every tag-based pathway end to end** — write a tag with secondaries, read it back, the form pre-populates; import a multi-color material from the OpenPrintTag DB, all slots land in the schema; export to CSV and re-import, no drift.

### NFC binary path (the big change)

| Direction | Site | Behavior |
|---|---|---|
| **Encode** | `src/lib/openprinttag.ts` | New `OPT_KEY.SECONDARY_COLOR_0..4` constants (spec keys 20–24). `OpenPrintTagInput` gains `secondaryColors?: string[]`. Encoder emits each populated slot as its own CBOR key→byte-string pair, capped at 5. |
| **Decode** | `src/lib/openprinttag-decode.ts` | New `secondaryColors?: string[]` on `DecodedOpenPrintTag`. Reads spec keys 20–24 in slot order; dense array in the result. New `bytesToRgbHex()` helper centralises the alpha-truncation rule both primary and secondary paths use. |
| **Write button** | filament detail page | Passes `filament.secondaryColors` through `generateOpenPrintTagBinary`. |
| **Read dialog** | `NfcReadDialog` | "Create new filament" URL params now carry `secondaryColors=...` comma-encoded. |
| **New filament page** | `src/app/filaments/new/page.tsx` | Parses the URL param, validates per-entry hex, surfaces via `initialData.secondaryColors` so the form pre-populates from a tag read. |

### OpenPrintTag DB importer

- `OPTMaterial` gains `secondaryColors: string[]` parsed from the spec's `secondary_color_0..4` top-level YAML fields
- Importer's conditional-update branch only adopts the OPT db's secondary colors when the existing row has none — same "only set if currently null/sentinel" posture as the existing color/density/etc. fields. Never overwrites user-set arrays.

### CSV/XLSX import + export

- New "Secondary Colors" column between Color and Diameter, comma-separated hex string
- Header map accepts `secondary colors`, `secondarycolors`, `secondary color` (case-insensitive)
- Importer trims whitespace, drops malformed hex, caps at 5
- Export → import round-trips without drift

## Documented spec-superset gap

OpenPrintTag spec's `color_rgba` is RGB or RGBA. We always store/emit RGB and drop alpha; translucency rides finish tags 5/6 with real CSS opacity instead. Decoder now truncates a 4-byte input to its RGB triple — a tag written by a previous version with alpha still decodes successfully, just loses the alpha component (which was never round-tripped to a UI surface that rendered alpha anyway). Two existing tests updated to reflect this.

## What this PR does NOT do

- No slicer-bound export changes — that's Phase 4
- No docs — that's Phase 5

## Tests

- **+12 new** (4 NFC round-trip, 5 CSV import, 3 CSV export)
- Full suite: **1594/1594 pass** (was 1582)
- Lint + tsc clean

## Phase tracking

This PR ticks [Phase 3 in #477](https://github.com/hyiger/filament-db/issues/477).

## Test plan
- [x] `npm run lint` clean
- [x] `npm test` 1594/1594
- [ ] @codex review
- [ ] Manual: import a multi-color material via `/openprinttag` (e.g. a coextruded Bambu silk), verify the imported filament's secondary colors render correctly on the detail page


hyiger:
@codex review

hyiger:
@codex review — addressed both P2 findings: (1) OPT importer preserves null primary when secondaryColors is non-empty (no phantom gray slot for coextruded materials); (2) browser `.bin` download route now passes `secondaryColors` and `color ?? undefined` to match the Electron NFC write path.

hyiger:
@codex review — addressed P2 r2: CSV import now preserves null primary for coextruded filaments. When secondaryColors has entries AND the Color cell is empty/null, explicitly writes `doc.color = null` so the schema default `#808080` doesn't re-introduce a phantom gray primary on round-trip. Regression test pins the export → import → re-export identity end-to-end.

hyiger:
@codex review — addressed P2 r3: OPT update branch now clears the gray sentinel to null when adopting coextruded secondaries from the OPT db. The conditional-update branch was only running the `color` branch when `payload.color` was truthy; for coextruded materials it's null so the sentinel survived. Now we explicitly clear sentinel→null alongside the secondary-color adoption.

PR #485, opened 2026-05-31
Multi-color phase 4: slicer export policy + UX notice (#477)

hyiger:
## Summary
- Slicer presets are single-color. Route `filament_colour` for both PrusaSlicer and OrcaSlicer bundle generators through `displayColor()` so a coextruded filament (null primary, secondary list) falls back to the first secondary instead of emitting no color at all. Otherwise the slicer renders its bare default for coextruded materials, which loses every color the user picked.
- Amber notice on the filament detail page's "Export for slicer" disclosure menu — fires only when `secondaryColors[]` is non-empty, so single-color users see no extra chrome.
- New i18n key `detail.slicerExport.multiColorNotice` in EN + DE (1275/1275 parity).
- Pinning tests on `filamentToSlicerKeys` + `filamentToOrcaSlicerKeys`: solid multi-color → primary only, coextruded with null primary → first-secondary fallback, secondaries never appear anywhere else in the keyset.

Tracks #477 phase 4 of 5.

## Test plan
- [x] `npm run lint`
- [x] `npx tsc --noEmit`
- [x] `npm test` — 1599 passing
- [ ] Manual: filament with secondaries → detail page → "Export for slicer" dropdown → notice visible above the three export buttons
- [ ] Manual: coextruded filament → PrusaSlicer / OrcaSlicer export → `filament_colour` = first secondary

### DIFF
--- a/src/app/filaments/[id]/page.tsx
+++ b/src/app/filaments/[id]/page.tsx
@@ -846,7 +846,17 @@ function FilamentDetail() {
                 <path strokeLinecap="round" strokeLinejoin="round" strokeWidth={2} d="M19 9l-7 7-7-7" />
               </svg>
             </summary>
-            <div className="absolute right-0 z-20 mt-1 w-64 bg-white dark:bg-gray-800 border border-gray-200 dark:border-gray-700 rounded shadow-lg py-1">
+            <div className="absolute right-0 z-20 mt-1 w-72 bg-white dark:bg-gray-800 border border-gray-200 dark:border-gray-700 rounded shadow-lg py-1">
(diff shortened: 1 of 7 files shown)


hyiger:
@codex review

hyiger:
Addressed in 03b557f3 — replaced `displayColor()` at the slicer-export site with a local `slicerExportColor()` that returns `null` when neither primary nor secondary exists, so `set` is a no-op and the slicer falls back to its own default (rather than emitting a forced `#808080` the user never picked). Pinning tests added for the empty-coextruded case on both generators.

@codex review

PR #487, opened 2026-05-31
Label printing for Brother PT-P710BT (Electron + spike + simulator)

hyiger:
## Summary

End-to-end label-printing support for the Brother PT-P710BT (P-touch CUBE), plus a UI cleanup pass on the filament detail page. Five commits, scoped tight per phase. **Do not merge until I've tested with the real printer Tuesday.**

### Background

The PT-P710BT is Bluetooth Classic only (SPP/RFCOMM). Web Bluetooth can't reach it (BLE GATT only). Per-OS native BT addons are a maintenance burden. Path: let the OS handle pairing and treat the resulting `/dev/tty.PT-P710BT-XXXX-Serialport` (macOS) / `COMx` (Windows) / `/dev/rfcomm0` (Linux) as a plain serial device via the `serialport` npm.

Encoder follows Brother's [Raster Command Reference for PT-E550W/P750W/P710BT](https://download.brother.com/welcome/docp100064/cv_pte550wp750wp710bt_eng_raster_102.pdf); validated against the [robby-cornelissen Python reference impl](https://github.com/robby-cornelissen/pt-p710bt-label-maker).

### Commits

| # | SHA | What |
|---|---|---|
| 1 | `0c5e199c` | CLI spike (`scripts/print-label-spike.ts`) + simulator (`scripts/print-label-sim.ts`). Two npm aliases: `label:print`, `label:sim`. |
| 2 | `6775326a` | Filament detail action row consolidated into `Export ▾` / `Sync ▾` / `Variants ▾` dropdowns (was 7 inline buttons, would have been 8 with Print Label). |
| 3 | `c8541692` | Brother raster encoder extracted to `src/lib/labelEncoder.ts` (pure JS, browser-safe). 19 new unit tests pin the byte sequence + grayscale packing. |
| 4 | `defa067a` | `<PrintLabelDialog>` with live preview canvas + per-print QR mode toggle (instanceId vs. deep-link URL, sticky default in localStorage). `src/lib/labelBitmap.ts` is the browser-side compositor. |
| 5 | `d434e771` | Electron IPC + serialport + Settings device picker. `electron/label-printer.ts` is transport only; encoding stays in the renderer. |

### Surface area

- **New files**: `scripts/print-label-spike.ts`, `scripts/print-label-sim.ts`, `src/lib/labelEncoder.ts`, `src/lib/labelBitmap.ts`, `src/components/PrintLabelDialog.tsx`, `src/components/LabelPrinterSettings.tsx`, `electron/label-printer.ts`, `tests/labelEncoder.test.ts`.
- **Modified**: `src/app/filaments/[id]/page.tsx` (button row refactor + dialog mount), `src/app/settings/page.tsx` (settings panel mount), `electron/main.ts` (IPC handlers + timeout helper extended), `electron/preload.ts` + `src/types/electron.d.ts` (API surface), `electron-builder.yml` + `package.json` (serialport packaging).
- **Deps**: `qrcode` + `@types/qrcode` (dev), `serialport` + `@serialport/parser-byte-length` (prod). `@serialport/bindings-cpp` is the native module; rebuild via the updated `electron:rebuild` script.

### Validation

- All 1622 tests pass (`npm test`)
- Lint clean (`npm run lint`)
- tsc clean (`npx tsc --noEmit`)
- Electron main bundle compiles (1.3 MB)
- Spike round-trip verified end-to-end with three label shapes (short instanceId + long name → 95mm, URL + short name → 34mm, mid → 69mm). Decoded preview matches source bit-for-bit.

### Open / known

1. **Real printer not yet on hand** — the Electron transport path (commit 5) is untested against actual hardware. Tuesday's plan: pair, pick in Settings, test-print, run real labels.
2. **Brother's official position** is "PT-P710BT Bluetooth is iOS/Android only" — the desktop SPP path is community-validated, not vendor-blessed. Worst case if a firmware update breaks the protocol: pin a known-working firmware version in docs.
3. **macOS 15 Bluetooth TCC** may prompt the first time the serial port opens. Plan to handle a denial gracefully (currently surfaces as a print toast error).
4. **CSP**: I didn't add anything that needs a CSP exception, but verify the dialog preview's data-URL render works under the Electron CSP (it should — `img-src data:` is already allowed).

### Test plan
- [ ] Tuesday: pair real PT-P710BT, configure in Settings, test print.
- [ ] Print on a multi-color filament (instanceId QR) — verify the swatch text renders crisply.
- [ ] Print with URL QR (longer payload) — verify the scannable region is dense enough.
- [ ] Trigger the "no printer configured" path (clear selection in Settings, try to print) — verify the toast points to Settings.
- [ ] Run `npm run label:sim --in <generated.bin>` on a downloaded payload from the web dialog — verify validation passes with no warnings.


hyiger:
@codex review

hyiger:
Addressed in 50b4caf7. Two-part fix per your second suggestion ("use a configured public/base URL"):

- Added `label-printer-get-public-url` / `set-public-url` IPC handlers with main-process validation (parseable, http(s), non-loopback).
- Added a 'Public base URL for label QR codes' field in Settings → Label Printer with the desc explaining why localhost won't scan.
- Dialog now resolves the effective deep-link via: configured public URL → non-localhost `window.location.origin` → null. When none, the URL-mode radio renders disabled with a pointer to Settings; web users with a real domain Just Work.

@codex review

hyiger:
Addressed in af44c310 — extracted to `electron/loopback-host.ts` so the helper is testable, then expanded `isLoopbackHostname` to cover every URL.hostname output shape:

- bracketed `[::1]` (the original miss)
- the full 127/8 IPv4 loopback range
- uncompressed `0:0:0:0:0:0:0:1` (+ bracketed)
- IPv4-mapped IPv6 in **both** dotted form (`::ffff:127.x.x.x`) **and** Node's auto-normalised hex form (`::ffff:7fNN:YYYY`) — discovered during testing that Node's URL parser auto-converts the dotted form to hex, which would have been the next gap.

36 new unit tests in `tests/loopback-host.test.ts` lock down each case + end-to-end through `new URL(...).hostname`. Renderer regex in PrintLabelDialog updated to cover the same shapes (it's the UX gate; main-process validator is the security boundary).

@codex review

hyiger:
Addressed in 019b3f5c — normaliser now strips trailing dots (`.replace(/\.+$/, '')`) so DNS absolute-name forms like `localhost.`, `127.0.0.1.`, and even `localhost..` collapse to the canonical shape before the equality checks fire. Seven new test cases pin the variants end-to-end through `new URL(...).hostname`. Full loopback test count: 43.

@codex review

hyiger:
Both addressed in 054b5a6e:

1. **Serial port cleanup**: replaced `settle()` with `settleWithCleanup(port, err)` — closes the SerialPort (if open) before rejecting, preferring the original error over any close error. Every post-open failure path (write, drain, .on('error')) now routes through cleanup, so a Bluetooth drop or write error no longer leaves the OS handle dangling.

2. **QR fit-to-band**: new `renderQrForTape()` helper probes at scale=1 to discover the module count, then picks the largest fitting scale. Throws a clear error when even scale=1 wouldn't fit ("QR payload too long for 24mm tape — would render at N dots, max 116"). Applied to both render functions in `labelBitmap.ts` AND the CLI spike in `scripts/print-label-spike.ts` so the preview, the print, and the simulator all agree.

@codex review

hyiger:
Both addressed in d07d9533:

1. **In-transport timeout** (`electron/label-printer.ts`): added `PRINT_TIMEOUT_MS = 25_000` timer inside the Promise body that routes through `settleWithCleanup` → `port.close()` → reject. Fires before the 30s IPC timeout so the transport's own cleanup runs first. Cleared on every settle path so fast successes don't leak.

2. **QR quiet zone in fit calc** (`src/lib/labelBitmap.ts` + `scripts/print-label-spike.ts`): renderers now use `margin: QR_QUIET_ZONE_MODULES (4)` and the fit calc uses `widthWithQuietZone = modules + 8` instead of the bare module count. Largest scale we pick guarantees the spec-required quiet zone fits inside the 116-dot print band. Visually verified the rendered preview now has a clear white border around the QR data.

@codex review

hyiger:
Addressed in be2b01de — hoisted the watchdog timer to the top of the Promise body, before `new SerialPort()` and before `.open()`, so a stalled open phase fires the same close-then-reject path as any other failure. Added belt-and-braces in the open callback: if `settled` is already true when open() finally returns, best-effort close any handle that was acquired so the OS doesn't leak it.

Also caught (during verify): the project tsconfig excludes `electron/` so a `const`→`let` slip in the previous commit was invisible to `npm run lint`. Fixed in this commit and verified with `npx tsc --noEmit electron/label-printer.ts` + esbuild bundle.

@codex review

hyiger:
Addressed in 32b30d2f — instanceId radio's checked + highlight now use `fallbackQrMode` (matching the URL radio) so the visible selection always tracks what would actually print.

@codex review

hyiger:
Addressed in 77fd1bd1 — added query-string + fragment rejection in the public-URL validator. Errors name the offending character so the user knows what to strip. A pasted URL like `https://example.com/?utm=x` now fails fast instead of producing `https://example.com/?utm=x/filaments/<id>` on print.

@codex review

hyiger:
Both addressed in 0a819cda:

1. **Bare `?` / `#` in public URL**: raw input check (`url.includes('?')`) before parsing — catches the bare-delimiter case the structured check missed. Per RFC 3986 those chars are always delimiters so the raw check is strictly stronger than `parsed.search` / `parsed.hash`; dropped the redundant parsed checks.

2. **node-gyp-build missing from packaged Electron build**: added the serialport runtime closure to `electron-builder.yml` — node-gyp-build (the .node runtime loader), node-addon-api (defensive), debug, ms. Without these, the packaged app would have crashed at first label-printer interaction with 'Cannot find module node-gyp-build'.

@codex review

hyiger:
Addressed in 70604813 — added `::` (and the uncompressed `0:0:0:0:0:0:0:0`) to isLoopbackHostname's match list, placed right after the `0.0.0.0` check since they're the same bind-anywhere semantic in v6. Six new test cases pin the shapes end-to-end. Full loopback test count: 49.

@codex review

hyiger:
Addressed in d1c83c8f — extracted `isLoopbackHostname` to `src/lib/loopbackHost.ts` (pure regex/string ops, no Node OR browser deps). Both sides import the same code:

- Electron main: `import { isLoopbackHostname } from "../src/lib/loopbackHost"` (esbuild bundles inline).
- Renderer (PrintLabelDialog): replaced the local regex with `isLoopbackUrl(window.location.origin)` (convenience wrapper around the hostname check).

Now a user opening the web UI through `localhost.`, `[::]`, an IPv4-mapped IPv6 form, or any of the 11 patterns covered by the test suite gets URL mode correctly disabled — matching what the main-process validator would have rejected at Settings save time. UX gate and security boundary can't drift again.

Full suite: 1671 tests passing.

@codex review

hyiger:
Addressed in 8d85d93d — added two patterns for the IPv4-mapped IPv6 unspecified address (`::ffff:0.0.0.0` dotted, `::ffff:0:0` hex-normalised + padded variants). Same v6-mapped pattern as the 127/8 fix in round r4 but for the bind-anywhere semantic. Seven new test cases pin the variants end-to-end. Loopback test count: 56.

@codex review

hyiger:
Doc cleanup + rename pass landed in 1381ed86 + e1d3f2f7:
- Dead i18n keys (`downloadedElectronStub`, `electronStubNotice`) removed
- Stale `(commit 5)` comments updated to describe landed architecture
- README.md, docs/usage.md + DE, CLAUDE.md gained user-facing + agent-facing docs for the label printer
- scripts/print-label-spike.ts → scripts/print-label.ts (no longer a spike; production CLI sharing the same encoder as the renderer)

@codex review

hyiger:
Addressed in 7e80cc66 — added a priority-0 `instanceId` branch to `/api/filaments/match` so the bare-hex QR codes the label printer dialog encodes in instance-ID mode actually resolve to a filament. Input validator restricts to 1-32 hex chars (no regex injection), case-insensitive match, no-match falls through to name/vendor/type when supplied (graceful for trashed filaments), soft-deleted filaments excluded.

Five new regression tests + OpenAPI updated to surface the new parameter as the highest-confidence match. Full suite: 1683 tests passing.

The dialog's i18n help text was already accurate ("Compact 5-byte hex identifier — the same value the NFC tag carries. Scans into Filament DB's match endpoint.") — aspirational when first written, now actually true.

@codex review

hyiger:
Addressed in e5958a5a — switched to the same case-insensitive regex pattern the name/vendor/type branches use:

```
{ instanceId: { $regex: `^${escapeRegex(trimmed)}$`, $options: 'i' } }
```

Input validator widened to `[0-9a-fA-F]{1,32}` so uppercase / mixed-case input passes through (escapeRegex still protects against injection). New regression test pins the 'uppercase stored / lowercase queried' direction so both sides are exercised.

@codex review

hyiger:
Addressed in 3947af85 — loosened the validator from hex-only to length-only (1-128 chars; `escapeRegex` is the real injection defense). OpenAPI dropped the hex pattern. Replaced the rejection-only test with three more thorough ones: non-hex IDs match (`custom-id-123`), regex-special chars match literally, and regex-special queries don't accidentally match unrelated stored IDs.

@codex review

hyiger:
Addressed in cf4730b9e7f20c1907793f50bc7c0fef3cd2d071 — restructured into two stages: exact-case match first (deterministic when the query case matches a stored case), then case-insensitive fallback only when the CI hit is unique. Multiple CI hits surface both as candidates instead of an arbitrary pick.

Two new regression tests pin both directions of the collision case.

@codex review

hyiger:
Re-review on cf4730b9e7f20c1907793f50bc7c0fef3cd2d071.

@codex review

hyiger:
Addressed in 6cc05893cea765d12f3e759bfe0b36eadcee7eb0 — `ESC i K` byte now `0x08` (bit 3 set = no-chain mode → feed + cut after the last label). Matching test pin updated.

@codex review

hyiger:
Addressed in 08a017c7e7cd2bdd2e0d6ce8401732512c0f2552 — ESC i K bit 3 now varies with `autoCut`: 0x08 (no-chain) when autoCut=true, 0x00 (chain) when autoCut=false. Restores the autoCut option's documented contract and pins both directions in the test suite.

@codex review

PR #494, opened 2026-06-01
Fix Electron dev workflow (#488 rebuild + #489 offline-mode data safety)

hyiger:
## Summary

Two related bugs in the Electron dev experience.

### #488 — electron:rebuild only built one native module

```diff
- "electron:rebuild": "electron-rebuild -f -o @pokusew/pcsclite -o @serialport/bindings-cpp"
+ "electron:rebuild": "electron-rebuild -f -o @pokusew/pcsclite,@serialport/bindings-cpp"
```

`electron-rebuild`'s `--only` takes a comma-separated list, not repeated `-o` flags — the second `-o` was silently ignored. Fresh checkouts running the documented rebuild then launched Electron and crashed with `NODE_MODULE_VERSION` mismatch on pcsclite. Verified locally that both modules now build (`Building modules: @pokusew/pcsclite, @serialport/bindings-cpp`).

### #489 — Offline Mode in dev silently hits .env.local

In dev (`!app.isPackaged`), Electron skips its own server and the renderer is loaded from a separately-running `next dev` that reads `MONGODB_URI` from `.env.local`. The connection-mode wizard still runs and Electron honors it on the main-process side (starts/stops local mongo, persists config) — but none of that reaches the renderer. A user clicking through Offline Mode sees a "Local" status pill while every read and write goes to whatever `.env.local` points to.

The right architectural fix would re-route the dev renderer through the embedded mongo (by relaunching `next dev` with the embedded mongo URI in its env), but that's invasive enough to be its own feature. This PR adds the minimum signal we can ship today so the footgun doesn't fire silently:

- New `get-runtime-mode` IPC handler returning `{ isPackaged }`.
- New `<DevModeBanner>` (in `src/components/`) — amber banner at top of the app shell, **Electron + !isPackaged only**. Dismissable; dismissal persists in `sessionStorage` so navigation doesn't keep re-summoning it, but a full restart surfaces it again.
- Mounted in `ClientProviders` alongside the existing `UpdateBanner`. Renders nothing in the web app or packaged Electron.

## Test plan
- [x] `npm test` — 1688 passing
- [x] `npm run lint` — clean
- [x] Electron main bundle compiles
- [x] `npm run electron:rebuild` actually rebuilds both modules
- [ ] Manual: launch `npm run dev` + `npx electron .` → confirm the amber banner appears
- [ ] Manual: launch packaged build → confirm NO banner appears


hyiger:
@codex review

PR #495, opened 2026-06-01
i18n batch: missing keys + German polish (#490 + #491)

hyiger:
## Summary

Two related i18n issues batched into one PR since they touch the same locale files.

### #490 — Missing keys leak raw IDs into aria labels

12 `t("...")` keys missing from both en.json and de.json — most are aria-labels, so the fallback (raw key) is invisible visually but screen readers happily announce "filter dot aria dot quick" to the user. Added each one with sensible English + German.

Per the issue's suggestion, also added **`tests/i18n-key-coverage.test.ts`** — walks `src/` for every literal `t("key")` call and asserts each exists in both locales AND that the two locales have identical key sets. ~250ms, no React/Vite. Replaces the manual `jq 'keys|length'` parity check we've been doing by hand. 5 new tests, full suite now 1693.

### #491 — German polish

- `share.publishSection`: "Neuen Share veröffentlichen" → "Neuen geteilten Katalog veröffentlichen". Consistent with adjacent `share.existingShares`.
- Three `new Date().toLocaleDateString()` sites in `src/app/share/*` and the filament-detail SpoolCard switched to `formatDate(date, locale)`. German users now see `22.05.2026` instead of `5/22/2026`.
- Four label-printer locale strings switched from informal imperative (`Koppele`, `Lass`) to formal `Sie` form, matching the rest of Settings copy.
- Out of scope: the all-caps `KI-ANBIETER` reading is a CSS choice on the section header (`uppercase` class), not an i18n issue. The borrowed `optional` in `form.parentFilamentHint` is fine as a German loanword.

## Test plan
- [x] `npm test` — 1693 passing (5 new i18n-coverage)
- [x] `npm run lint` — clean
- [x] `jq 'keys|length'` on both locales — 1340/1340 parity
- [ ] Manual: open in DE locale → verify the four touched strings + the date format


hyiger:
@codex review

PR #496, opened 2026-06-01
Bind SkipToMain href to current pathname (#493)

hyiger:
## Summary

Audit #493 found the accessibility tree showed `localhost:3456/settings#main-content` as the skip-link's target value after navigating to `/dashboard`, `/inventory`, etc. The bare `href="#main-content"` is semantically correct (browsers resolve fragments at click time) but some AX trees cache the resolved URL on first paint.

Fix: bind the href to `usePathname()` so the rendered attribute value mutates on every route change. Forces the AX cache to refresh. Click behavior is identical — `#main-content` and `/dashboard#main-content` resolve to the same target when activated from `/dashboard`.

One-line component change. Lint + tsc clean.

## Test plan
- [x] `npm run lint` clean
- [ ] Manual in Electron: navigate Settings → Dashboard → Inventory → Compare → Filaments, inspect AX tree on each, confirm skip-link value matches the current page


hyiger:
@codex review

hyiger:
Addressed in 675af51bb0521939cd611bb7b8c85e7fa040be6e — appended `useSearchParams().toString()` to the rendered href so /compare?ids=... keeps its state when the skip link is activated. Wrapped in a local `<Suspense fallback={null}>` boundary since useSearchParams() can suspend during App Router static rendering / transitions.

@codex review

hyiger:
@codex review

PR #497, opened 2026-06-01
Add search/filter to Compare and Share filament pickers (#492)

hyiger:
## Summary

With catalogs of dozens of filaments, the /compare and /share pages required scrolling through a long checkbox list to find items, with no search or filter — the rest of the app's pages have that affordance, the pickers didn't.

New **`<FilamentPicker>`** in src/components/ wraps the checkbox-list pattern with:

- **Search input** — substring match across name, vendor, type, color
- **Type filter chips** sorted by catalog frequency (single-select; click again to clear)
- **Selected-count badge** with a "show selected only" toggle
- **Empty states** for "no matches", "no selected matches", "catalog empty"
- **Selection cap** prop — /compare uses 8 (matches its table column cap)

Controls auto-hide below 12 filaments so small catalogs don't get unnecessary chrome.

Wired into both /compare (with cap=8) and /share (no cap). The Share page's `FilamentOption` interface gained a `type` field — the API already returns it, only the type definition was missing.

i18n: 11 new keys per locale (1339 total each).

## Test plan
- [x] `npm run lint` clean
- [x] i18n parity 1339/1339
- [ ] Manual: /compare with 12+ filaments → search, type filter, selected-only toggle, max-8 cap behavior
- [ ] Manual: /share with 12+ filaments → same as above without cap


hyiger:
@codex review

hyiger:
Addressed in 58718d979132e2dca50b632285a719436eac2571 — derived an `effectiveSelectedOnly = showSelectedOnly && selectedCount > 0` at render time. The filter predicate and empty-state copy both use the derived value, so the moment the selection set goes empty the filter becomes a no-op even if the persisted toggle is still true. Same fallback pattern PR #487's PrintLabelDialog uses for its QR-mode derivation.

@codex review

hyiger:
Re-applying in 54febad86ce4236dfe04604ad5d5c9bec7dcf446 — the r1 commit landed empty due to a harness file-state race. Actual diff verified before this push.

@codex review

PR #530, opened 2026-06-02
Security hardening: IPC guard, byte validation, SSRF, dead handler (#506, #523)

hyiger:
## Summary
- **#506** `check-atlas-connectivity` now requires `assertTrustedSender` — same sub-frame attack surface that #432 closed on `trigger-sync`. Polled by `SyncStatusIndicator`, so a compromised sub-frame could exhaust Atlas connection-pool slots / billable connections.
- **#523.1** `label-printer-print` per-byte validation that bytes are integers in [0, 255]. Without it, `new Uint8Array(bytes)` silently coerces floats/NaN/Infinity/strings/null to 0 and wraps out-of-range mod 256 — Brother's positional raster protocol leaves the printer in a wrong-mode / chain-stuck state for the next print. Mirrors #278's nfc-write-tag pattern.
- **#523.2** `/api/prusament` now does manual hop-by-hop fetch with `assertExternalUrl` on every redirect target + `ssrfDispatcher` pinning the socket to the SSRF-validated IP. Same pattern `tdsExtractor` and `embed-check` already use. Risk on hardcoded prusament.com was low; this restores the "all outbound fetch goes through the dispatcher" invariant.
- **#523.3** Dead `show-message` IPC handler + preload bridge + type declaration deleted. Was unguarded AND had zero renderer callers — perfect material for UI-spoofing prompts parented to the trusted main window if any code in the renderer is ever compromised.

## Test plan
- [x] `npm run lint` clean
- [x] `npx vitest run tests/prusament.test.ts` — 8/8 pass
- [ ] CI green
- [ ] Codex review

Fixes #506, #523.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1114,8 +1114,18 @@ ipcMain.handle("label-printer-print", async (event, bytes: number[]) => {
   // `ESC i z` media width, `ESC i M`/`K` mode bits, `0x1A`/`0x0C`
   // trailer — and a stray byte in the wrong slot leaves the printer
   // in a wrong-mode / chain-stuck state for the next print.
-  if (!bytes.every((b) => Number.isInteger(b) && b >= 0 && b <= 255)) {
-    throw new Error("bytes must contain only integers in [0, 255]");
+  //
+  // Codex P2 round 1: use an index loop, not Array.prototype.every —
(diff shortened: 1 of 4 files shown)


hyiger:
@codex review

hyiger:
@codex review

Issue #511, opened 2026-06-02
Hybrid sync streams full filament documents (incl. photoDataUrl blobs) every cycle


hyiger:
`SyncService.syncCollection` (`electron/sync-service.ts:616-617`) calls `localCol.find({}).toArray()` and `remoteCol.find({}).toArray()` with no projection on every sync cycle — every 5 minutes by default (`startPeriodicSync` line 192). For the filaments collection this pulls every embedded `spools[].photoDataUrl` (base64), `usageHistory[]`, `dryCycles[]`, `calibrations[]`, etc. across the wire on both sides, even though the diff loop at lines 619-625 only needs `syncId`, `updatedAt`, `_deletedAt`, and `_purged` to decide which docs need pushing/pulling.

The buildup phase before `syncCollection` repeats the pattern six more times for the syncId→_id remap maps — lines 295/296 (nozzles), 316/317 (bedtypes), 336/337 (printers), 359/360 (locations), 411/412 (filaments), 482/483 (filaments rebuilt post-sync). Each pair feeds only `new Map(docs.filter(d => d.syncId).map(d => [d.syncId, d._id]))`, so the entire payload past `_id` and `syncId` is thrown away. Filaments are read in full THREE times per cycle (411, 482, plus the syncCollection at 616).

On a hybrid-mode install with a few hundred filaments carrying photo attachments, every 5-minute tick streams tens to hundreds of MB out of Atlas — Atlas egress is metered and cycle latency scales linearly with collection size. The public GET /api/filaments aggregation already drops these exact heavy fields (documented in CLAUDE.md) — the sync engine should do the same.

**Suggested fix:**
- In the six map-build reads (lines 295/316/336/359/411/482), add `.project({ _id: 1, syncId: 1 })` — pure perf win, zero behavior change.
- In `syncCollection` (line 616), split into a lightweight diff phase with `.project({ syncId: 1, updatedAt: 1, _deletedAt: 1, _purged: 1 })` then `findOne(_id)` to hydrate only the docs the diff identifies as needing a push/pull/update.

`electron/sync-service.ts:616` (primary) plus 295/316/336/359/411/482 (secondary).

_Filed from full-audit run (high confidence, adversarially verified). Atlas-tier cost + latency win._

hyiger:
[...]

PR #532, opened 2026-06-02
Sync correctness: respect _purged on atlas import, project sync reads (#499, partial #511)

hyiger:
## Summary

- **#499** Atlas import resurrection path filters on \`_purged: { \$ne: true }\` to match every other resurrection site. Pre-fix, an Atlas import containing a name the user permanent-deleted locally would flip \`_deletedAt\` back to null while leaving the tombstone set — the row became invisible (list + trash filter on \`_purged: { \$ne: true }\`) but occupied the partial-unique active-name index.
- **#511 (partial)** Six syncId→id map-build reads + the repair pass's two filament reads now project to \`{_id, syncId}\` (+ updatedAt / parentId where consumed). The maps are only used for id translation — streaming full filament docs (incl. spools[].photoDataUrl base64) across the wire just to throw all but two fields away was pure Atlas egress waste.

**Deferred:** the bigger syncCollection-level split (line 616 — slim diff projection + per-changed-doc hydrate) is the issue's main fix suggestion but more invasive. Carving out for a separate PR with its own review pass — this PR captures the safe, behavior-neutral half.

## Test plan
- [x] \`npm run lint\` clean
- [x] \`npx tsc --noEmit\` clean
- [x] sync-service-locations + sync-service-expansion + sync-service-error-isolation: 39/39 pass
- [ ] Full CI green
- [ ] Codex review

Fixes #499. Partially addresses #511 (syncCollection split deferred).

### DIFF
--- a/electron/sync-service.ts
+++ b/electron/sync-service.ts
@@ -291,9 +291,11 @@ export class SyncService extends EventEmitter {
         this.syncCollection(localDb, remoteDb, "nozzles"),
       ));
 
-      // Build nozzle syncId→ID maps for reference remapping
-      const localNozzles = await localDb.collection("nozzles").find({ _deletedAt: null }).toArray();
-      const remoteNozzles = await remoteDb.collection("nozzles").find({ _deletedAt: null }).toArray();
+      // Build nozzle syncId→ID maps for reference remapping.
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #535, opened 2026-06-02
TDS extractor: drop unsupported bedRangeMin/bedRangeMax (#518)

hyiger:
## Summary
- The TDS extraction prompt asked the AI for \`bedRangeMin\` / \`bedRangeMax\` and the interface declared them — but \`Filament.temperatures\` only models \`bed\` + \`bedFirstLayer\` (no bed range, unlike nozzle). Every bedRange value the AI returned was silently dropped after we paid for it.
- The one fallback that used the dropped field (\`new/page.tsx:423\`) was also semantically off: it routed the AI's "minimum operating bed temp" to the form's first-layer bed temp, which is typically the HIGHER value (better adhesion), not the lower.
- Option (a) from the issue: dropped bedRangeMin/Max from the interface + prompt, fixed the misleading fallback to use \`bed\` instead. Prompt now tells the AI to use the midpoint for the single \`bed\` field when a range is given.

## Test plan
- [x] \`npm run lint\` clean
- [x] \`npx tsc --noEmit\` clean
- [x] tdsExtractor tests: 33/33 pass
- [ ] Full CI green
- [ ] Codex review

Fixes #518.

### DIFF
--- a/src/app/filaments/new/page.tsx
+++ b/src/app/filaments/new/page.tsx
@@ -420,7 +420,14 @@ function NewFilamentContent() {
         nozzleRangeMin: d.temperatures?.nozzleRangeMin ?? null,
         nozzleRangeMax: d.temperatures?.nozzleRangeMax ?? null,
         bed: d.temperatures?.bed ?? null,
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #544, opened 2026-06-02
_purged tombstones for PrintHistory + SharedCatalog (#524.5)

hyiger:
## Summary
The last sub-item of #524 — PrintHistory and SharedCatalog had only `_deletedAt`, no `_purged` permanent-delete tombstone like Filament, and no purge path.

- Both models gain an indexed `_purged: Boolean`. The hybrid-sync engine's generic `_purged` branch already propagates it — adding the field is all that's needed.
- `DELETE …?permanent=true` on both routes sets the tombstone, gated on the row already being soft-deleted/unpublished (mirrors Filament's trash→purge gate). Idempotent.
- Read paths unchanged: a `_purged` row always has `_deletedAt` set, so existing `_deletedAt: null` filters hide it. Snapshot round-trips it automatically.

**Scope note:** this brings parity with Filament + a callable purge path. Physically pruning converged tombstones to shrink the per-cycle sync fetch (issue option b) is a larger lazy-prune design Filament itself doesn't solve — left for a follow-up.

## Test plan
- [x] New tests: permanent-delete gating + idempotence for both routes (print-history + share)
- [x] print-history + share-route + sync-service-expansion: pass
- [ ] Full CI green
- [ ] Codex review

Partially addresses #524 (sub-item 5).

### DIFF
--- a/src/app/api/print-history/[id]/route.ts
+++ b/src/app/api/print-history/[id]/route.ts
@@ -170,6 +170,31 @@ export async function DELETE(
   try {
     await dbConnect();
     const { id } = await params;
+
+    // GH #524.5: ?permanent=true sets the `_purged` tombstone, mirroring
+    // the Filament permanent-delete path. Gated on the row ALREADY being
+    // soft-deleted (`_deletedAt != null`) so an accidental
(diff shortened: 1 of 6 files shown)


hyiger:
@codex review

PR #546, opened 2026-06-02
Bulk-op perf + UX (#525.1, #525.2)

hyiger:
## Summary

Closes the two remaining items under #525 (the broader perf-projection sweep landed in v1.34.2 via PR-H).

### #525.1 — `POST /api/spools/import` N+1
A bulk paste of N spools for the **same** filament used to do N `findOne` + N `save()` round-trips (each `save()` a full hydrate+persist against possibly-remote Atlas). Now:
- filament lookups are cached by `name|vendor` (same shape as the existing `locationCache`)
- `save()` is deferred to **once per touched filament** after all its rows are applied to the in-memory doc
- per-row outcomes tracked in an index-keyed `rowResults` array so result ordering is preserved
- each filament's single save wrapped in its own try/catch — a `VersionError` on one material reports against only its rows, the rest of the batch still completes

New regression test: a 20-row same-filament import issues **exactly one** `findOne` + **one** `save()`.

### #525.2 — filament bulk-delete UX (list page)
- live **`Deleting {done}/{total}`** progress counter on the button
- a **Stop** control that aborts between rows (in-flight delete finishes; no further deletes start) → `Deleted {count} before stopping`
- an **aggregated error dialog** (first 10 failures + "…and N more") instead of an ever-growing error toast that overflowed on large batches
- selection now only drops the rows actually deleted, so aborted/failed rows stay selected for retry

`ConfirmDialog` gains a `hideCancel` option (acknowledge-only notice); the existing focus trap already tolerates the missing cancel button. New i18n keys land in **both** en + de.

## Test plan
- `tests/spools-import-route.test.ts` — 24/24 (incl. new N+1 regression test)
- `tests/i18n-key-coverage.test.ts` — green (en/de parity)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #543, opened 2026-06-02
Add label-printer transport tests (#526 remainder)

hyiger:
## Summary
The second half of #526 — `electron/label-printer.ts` had no coverage of its `settleWithCleanup` invariants (electron/ is outside the enforced coverage scope, so CI never flagged regressions in the always-close-before-reject guarantees the code's comments document as Codex rounds 4-6 on PR #487).

`tests/label-printer.test.ts` mocks `serialport` with a controllable fake and pins: happy path, every failure mode routing through cleanup, the post-open async error event, the hung-open stall watchdog (no leaked port), the late-open-after-timeout close, and idempotent settle. Pure test addition.

The labelBitmap half of #526 shipped in #541 (coverage exclusion), so this closes #526.

## Test plan
- [x] `tests/label-printer.test.ts` — 9/9 pass
- [x] `npm run lint` clean
- [ ] Full CI green
- [ ] Codex review

Fixes #526.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #547, opened 2026-06-02
Fix spool-import doc dedup + NUL cache-key delimiter (Codex P1+P2 on #546)

hyiger:
## Why

Two issues Codex flagged on **#546** that merged into **v1.34.3** before the review was seen. (Codex posted these as *review submissions* anchored to file lines, not issue comments, and the branch-update merge commits moved the anchors — so the merge-readiness poll showed no comments/threads. Lesson logged: also check `reviews[]`, not just `comments[]` + `reviewThreads`.)

## P1 — silent spool data loss

A CSV mixing omitted and matching vendor for the **same** filament:

```
filament,vendor,totalWeight
PLA,,800          # no vendor → matches by name
PLA,Vendor A,900  # matching vendor → same _id, DIFFERENT cache key
```

`resolveFilament()` cached by `(name, vendor)`, so these two rows issued two `findOne()`s and hydrated **two separate Mongoose document instances for the same `_id`**. The save loop keyed `touched` by `_id` and saved only the first instance — the second row's spool was pushed onto the second instance, never saved, **yet the row reported `ok: true`**. The spool silently vanished.

**Fix:** resolve (or create) the per-`_id` `touched` bucket *before* any mutation, and mutate only `bucket.doc`. Every row for a given filament now accumulates onto the one instance that actually gets saved. Regression test imports alternating omitted/matching-vendor rows and asserts all three spools persist (pre-fix: 2).

## P2 — route.ts became a binary file

The cache-key delimiter smuggled a literal **NUL byte (`0x00`)** into the source (`` `${name}␀${vendor}` ``), making git treat `src/app/api/spools/import/route.ts` as binary — `git diff` reported only `Binary files differ`, so every future change to this route was opaque to review/tooling.

**Fix:** `JSON.stringify([name, vendor])` — collision-free (no single-char delimiter is safe for arbitrary name/vendor strings), text-safe, no NUL.

Also drops the now-unused constructor-callback param in the label-printer test fake (the `autoOpen:false` mock never invokes it) to clear a lint warning introduced in #543.

## Test plan
- `tests/spools-import-route.test.ts` — 25/25 (incl. new #546 regression)
- `tests/label-printer.test.ts` — 9/9
- Confirmed `route.ts` contains no NUL byte (`grep -P '\x00'` empty)


hyiger:
@codex review

hyiger:
Fixed in 05a72798 — moved the per-_id bucket resolution to AFTER all per-row validation (right before the spool mutation), so a row that fails purchaseDate/openedDate validation no longer registers its filament for save(). Regression test asserts save() is never called when a filament's only row fails validation. The P1 canonical-instance fix is preserved.

@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #548, opened 2026-06-02
Fix #510: stop useUnsavedChanges leaking a history-guard entry per form visit

hyiger:
Closes #510.

## The bug

`useUnsavedChanges` pushes one synthetic **guard** history entry on mount to intercept the browser Back button. But it never consumed that entry on the programmatic exit paths — every consumer did `router.push(dest)` directly after save/discard. So the guard got buried under the destination, and a fresh guard was pushed on the next form visit: **the history stack accumulated dead entries across edits**, and Back from a destination could land on a stale form URL.

## The fix

The hook now owns programmatic navigation:

- New **`navigate(url)`** (and the link-click branch of `confirmNav`) consumes the guard via `history.back()` **before** pushing the destination — so it's never buried. A one-shot `pendingLeaveRef` + the existing `popstate` listener sequence the back→push.
- The unmount cleanup also pops the guard — but **only when it's still the live current entry** (`window.history.state?.unsavedGuard === true`). This guard-the-pop condition is the crux: the earlier reverted attempt (the one Codex flagged as incomplete) popped *unconditionally* on unmount and went back one entry too far whenever the destination was already current (the `router.push` path). Checking that our guard is genuinely current is what makes it safe.
- `guardActiveRef` tracks whether the guard is live, so React strict-mode double-invoke (dev) and unexpected unmounts are handled without leaking.

All 10 form pages (nozzles / printers / locations / bed-types / filaments × new/edit) now route save **and** discard through `navigate` / `confirmNav` instead of calling `router.push` directly. The redundant per-page `useRouter` + `handleDiscard` were removed. The popstate-discard path keeps its existing `go(-2)` behaviour (GH #285). The filaments/new "populate from" dialog is untouched (it's not a navigation).

## On testing

No automated test is added: the repo's vitest environment is `node` with **no jsdom/testing-library harness** (there are zero component/hook tests), and jsdom's history API doesn't faithfully model `pushState`/`back`/`popstate` ordering — a test there would assert the mock, not the behaviour. The logic is documented inline instead. Happy to add one if we stand up a jsdom harness.

## Test plan
- Manual: dirty a form → Back button prompts; Discard leaves cleanly; Save navigates with no extra history entries accumulating across repeated edits.

### DIFF
--- a/src/app/bed-types/[id]/edit/page.tsx
+++ b/src/app/bed-types/[id]/edit/page.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useEffect, useState } from "react";
-import { useParams, useRouter } from "next/navigation";
+import { useParams } from "next/navigation";
 import Link from "next/link";
 import BedTypeForm from "@/app/bed-types/BedTypeForm";
 import { useToast } from "@/components/Toast";
@@ -11,7 +11,6 @@ import { useTranslation } from "@/i18n/TranslationProvider";
 
 export default function EditBedType() {
   const params = useParams();
-  const router = useRouter();
   const { toast } = useToast();
   const { t } = useTranslation();
   const [bedType, setBedType] = useState(null);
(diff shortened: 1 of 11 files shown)


hyiger:
@codex review

PR #562, opened 2026-06-03
fix(settings): accept displayed DELETE token in danger-zone confirm (#559)

### LINKED ISSUE #559: Delete All Data prompt asks for DELETE but only accepts lowercase delete
## Summary

The Settings danger-zone confirmation copy tells the user to type `DELETE`, but the destructive action only enables when the user types lowercase `delete`.

## Reproduction

1. Open Settings > Danger Zone.
2. Click `Delete All Data`.
3. Type `DELETE`, matching the instruction and placeholder.
4. Observe the `Delete Everything` button.
5. Replace the value with lowercase `delete`.

## Observed

`DELETE` leaves the button disabled. Lowercase `delete` enables the button and the wipe succeeds.

## Expected

Either accept the displayed confirmation token (`DELETE`) or update the instruction and placeholder to lowercase `delete`. The current mismatch makes a dangerous action safer than intended, but it also makes the confirmation UI look broken at the exact point where users need clarity.

## Audit context

Found during a full destructive-flow pass in the Electron app against an isolated disposable database. The final wipe itself completed successfully once lowercase `delete` was entered.

hyiger:
## Summary
The **Delete All Data** confirmation tells the user to type `DELETE` (both `settings.deleteTypeConfirm` and the `settings.deleteTypePlaceholder` show `DELETE`), but the **Delete Everything** button only enabled on lowercase `delete`. Typing the displayed uppercase token left the button disabled — the confirmation UI looked broken right where users need clarity, and the dangerous action was effectively gated on guessing the wrong case.

## Fix
Compare the input against the displayed token `DELETE` (trimmed) so the instruction, placeholder, and the enable-gate all agree.

`src/app/settings/page.tsx` — one-line change to the disabled predicate.

## Testing
- ESLint clean.
- No test pinned the old lowercase behavior.

Fixes #559

### DIFF
--- a/src/app/settings/page.tsx
+++ b/src/app/settings/page.tsx
@@ -1003,7 +1003,7 @@ export default function SettingsPage() {
                     setDeleteInput("");
                   }
                 }}
-                disabled={deleteInput !== "delete" || deleting}
+                disabled={deleteInput.trim() !== "DELETE" || deleting}
                 className="px-4 py-1.5 bg-red-700 text-white rounded text-sm hover:bg-red-600 disabled:opacity-30 disabled:cursor-not-allowed transition-colors"
               >
                 {deleting ? t("settings.deleting") : t("settings.confirmDelete")}


hyiger:
@codex review

PR #563, opened 2026-06-03
fix(list): keep parents with their own spool under the Has spools filter (#552)

### LINKED ISSUE #552: Has spools quick filter returns no rows despite active spool
## Finding
The `Has spools` quick filter on the filament list returned an empty table even though an active spool was visible on the same list and on the filament detail page.

## Environment
- Filament DB v1.34.5
- Electron app, offline/local mode
- Walkthrough date: 2026-06-02 America/Los_Angeles / 2026-06-03 UTC

## Repro
1. Start with a filament that has an active spool.
2. In this walkthrough, `Electron Galaxy PLA` had one active spool with `525g` remaining after logging usage.
3. Open the filament list `/`.
4. Confirm the parent row shows a spool progress indicator (`53% remaining`).
5. Click the `Has spools` quick filter.

## Actual
The table becomes empty.

## Expected
The filament with the active spool should remain visible under `Has spools`.

## Notes
Inventory and dashboard both agreed there was one active spool, so this looks isolated to the quick-filter predicate or the list data shape used by that filter.

hyiger:
## Summary
The **Has spools** quick filter on the filament list returned an empty table even when an active spool was visible on the same list — reported in the v1.34.5 walkthrough (`Electron Galaxy PLA`, one active spool, 53% remaining).

## Root cause
The filter and its chip count both resolve against `inventoryFilaments`, which excludes every parent (`!f.hasVariants`). Parent exclusion exists so a parent's gram/spool totals don't double-count its variants' inventory. But when the parent carries **its own** spool, that roll is real — and the presence filter dropped it.

## Fix
`src/app/page.tsx`:
- "Has spools" now resolves against the full filament list and counts the same way (`(f.spools?.length ?? 0) > 0`), so a parent with its own spool stays findable and the chip badge still matches the rendered rows.
- `lowStock` / `noCalibration` keep resolving against `inventoryFilaments` (unchanged) — a parent that "matches" those by virtue of exclusion must not appear, per the prior Codex P2 on #356.

## Testing

Fixes #552

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -330,27 +330,41 @@ export default function Home() {
     };
     for (const f of inventoryFilaments) {
       if (isLowStock(f)) counts.lowStock++;
-      if ((f.spools?.length ?? 0) > 0) counts.hasSpools++;
       if (!f.hasCalibrations) counts.noCalibration++;
     }
+    // #552: "Has spools" is a presence check, not an inventory aggregate.
+    // A parent is excluded from `inventoryFilaments` because its gram/spool
+    // totals live on its variants — but a parent can still carry its OWN
+    // spool, and that roll is real. Count every filament with its own
+    // spool, parents included, so the chip badge matches the rows the
+    // filter renders (see the matching source switch in `visibleFilaments`).
+    counts.hasSpools = filaments.filter(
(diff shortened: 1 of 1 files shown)


hyiger:
@codex review

PR #564, opened 2026-06-03
fix(list): resolve inherited values on variant search-result rows (#553)

hyiger:
## Summary
When a color variant is shown as its own search-result row on the filament list, inherited values disappeared from the columns — searching `Red` showed `Electron Galaxy PLA Red` with `—` for the **Nozzle** and **Bed** columns, even though the detail page, compare page, and expanded parent/variant view all resolved them correctly (#553).

## Root cause
Search is server-side. A name search returns only the matching variant; its parent doesn't match the regex and is dropped by the aggregation's `$match`. The list page merges inherited scalars (nozzle/bed/cost/density/spool weights) **client-side** from the parent row — but in a search result the parent isn't in the fetched set, so there's nothing to inherit from.

## Fix
Resolve the inheritable scalars server-side in the list aggregation using the existing `_parent` `$lookup` (which runs against the full collection regardless of the search `$match`):
- `temperatures.nozzle` / `temperatures.bed`, `cost`, `density`, `spoolWeight`, `netFilamentWeight` now fall back to the parent via `$ifNull` (collapses null + missing — matches `resolveFilament`'s scalar-inheritance rule).
- Standalones / parents have no `_parent` doc, so `$arrayElemAt(…, 0)` is null and they're unaffected.
- The client-side merge stays as defense in depth.

## Testing
- 2 new cases in `tests/filaments-list-route.test.ts`: a searched variant inherits the parent's scalars when the parent is filtered out; a variant's own values still win.
- Full `filaments-list-route`, `variant-edit-preserves-inheritance`, `filaments-parents-route` suites pass; ESLint + `tsc --noEmit` clean.

Fixes #553

### DIFF
--- a/src/app/api/filaments/route.ts
+++ b/src/app/api/filaments/route.ts
@@ -94,7 +94,29 @@ export async function GET(request: NextRequest) {
           // uses at read time. Without this the list drops inherited
           // multi-color data and the list-row swatch can't render it.
           // (Codex P2 on PR #482.)
-          pipeline: [{ $project: { calibrations: 1, optTags: 1, secondaryColors: 1 } }],
+          //
+          // GH #553: also carry the inheritable scalar fields (temperatures,
+          // cost, density, spoolWeight, netFilamentWeight) so the projection
+          // can resolve them server-side. The list page used to merge these
+          // client-side from the parent row, but a name search returns only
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #565, opened 2026-06-03
fix(bed-types): block delete when a filament names the surface in bedTypeTemps (#557)

### LINKED ISSUE #557: Bed type deletion ignores filament bed-temperature references
## Summary

Deleting a bed type is blocked when it is installed on printers or used by calibrations, but it currently succeeds when active filaments reference that bed type from `bedTypeTemps`.

## Reproduction

1. Create a bed type.
2. Create one or more active filaments that include that bed type in their bed-temperature settings.
3. Send `DELETE /api/bed-types/:id` for the referenced bed type.
4. Fetch the active filaments again.

## Observed

During the destructive-flow audit, referenced nozzles and locations returned clear `400` guard errors, but the referenced bed type returned `200` / `Deleted`. The active filaments remained and still had bed-temperature data pointing at the deleted bed type.

## Expected

Bed type deletion should either block while any active filament references it in bed-temperature settings, or explicitly clear/migrate those references. The current behavior can leave dangling references and silently removes a selectable calibration surface that existing filament data still depends on.

## Audit context

Found while running the Electron app against an isolated disposable MongoMemoryServer database and exercising destructive flows end to end.

hyiger:
## Summary
Deleting a bed type was blocked when it was installed on printers or used by calibrations, but it succeeded while active filaments referenced it in their per-bed-type temperature table — leaving data that names a now-deleted surface (#557).

## Context
`bedTypeTemps[].bedType` is a **free-text slicer surface key**, deliberately *not* a `BedType` ObjectId ref (see the docblock in `src/models/Filament.ts`). So there's no FK to follow — but a user who created a bed type "Textured PEI", named it in a filament's per-surface temps, then deleted the catalog entry, silently loses a selectable surface their data still depends on, with none of the guard errors the parallel nozzle/printer/location deletes give.

## Fix
`src/app/api/bed-types/[id]/route.ts`:
- Load the target bed type up front (needed for the name match + a clean 404).
- Add a third guard: block the delete with a 400 when any **active** filament names the bed type (by name) in `bedTypeTemps[].bedType`, mirroring the existing calibration/printer guards.

`public/openapi.json` — 400 description broadened to name all three reference paths.

## Testing
- New tests: refuses when an active filament names it (and isn't soft-deleted); ignores soft-deleted filaments + unrelated surface names. Full `bed-types-route` suite passes.

Fixes #557

### DIFF
--- a/public/openapi.json
+++ b/public/openapi.json
@@ -2719,7 +2719,7 @@
             }
           },
           "400": {
-            "description": "Bed type is still referenced",
+            "description": "Bed type is still referenced by a filament calibration, a printer's installed bed types, or a filament's per-bed-type temperatures",
             "content": {
               "application/json": {
                 "schema": {
--- a/src/app/api/bed-types/[id]/route.ts
+++ b/src/app/api/bed-types/[id]/route.ts
@@ -74,6 +74,14 @@ export async function DELETE(
     await dbConnect();
     const { id } = await params;
 
+    // Load the target up front: we need its name to match against the
+    // free-text `bedTypeTemps[].bedType` keys below, and a clean 404 when
+    // it's already gone.
+    const bedType = await BedType.findOne({ _id: id, _deletedAt: null }).lean();
+    if (!bedType) {
+      return errorResponse("Not found", 404);
+    }
+
     // Prevent deleting a bed type that is referenced by any filament calibration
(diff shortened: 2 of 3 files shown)


hyiger:
@codex review

PR #566, opened 2026-06-03
fix(detail): refresh printer slots after retiring a loaded spool (#558)

### LINKED ISSUE #558: Retiring a loaded spool leaves stale printer-slot assignment visible
## Summary

Retiring a spool that is loaded in a printer slot clears the assignment in the backend, but the filament detail page continues to show the stale printer-slot assignment until state is refreshed.

## Reproduction

1. Create a printer with a slot.
2. Create a filament with an active spool.
3. Assign that spool to the printer slot.
4. Open the filament detail page and click `Retire` on the loaded spool.

## Observed

The backend behaved correctly:

- `GET /api/spools/:spoolId/assignment` returned `null` after retire.
- Default inventory excluded the retired spool.
- `includeRetired=1` included it.

However, the detail card still displayed the old `Printer slot : <printer> · <slot>` assignment and its clear action after the retire completed. Unretiring the spool left the stale slot text visible in the same local state, even though the backend assignment remained cleared.

## Expected

Once retire succeeds, the detail UI should immediately clear or refetch the spool's printer-slot assignment so the user does not see a loaded state that no longer exists.

## Notes

This is narrower than #268. That issue's backend integrity problem appears fixed; this audit only reproduced a stale detail-page UI state.

hyiger:
## Summary
Retiring a spool that's loaded in a printer AMS slot clears the assignment server-side, but the filament detail card kept showing the stale `Printer slot: <printer> · <slot>` text (and its clear action) until a manual refresh. Unretiring then carried the stale text forward even though the backend assignment stayed cleared (#558).

## Root cause
The spool PUT handler already clears the slot on retire (`assignSpoolToSlot(Printer, spoolId, null)`), and `GET /api/spools/:spoolId/assignment` returns `null` afterward. But the card's slot text is **derived client-side** from the `printers` state, and `handleUpdateSpool` only refreshed `filament.spools` — never `printers`.

## Fix
`src/app/filaments/[id]/page.tsx`:
- Re-pull printers after a successful retire (`data.retired === true`) so the derived slot text matches the server.
- Extracted the existing `fetch("/api/printers")` + `setPrinters` into a `refreshPrinters` helper, now shared with `handleAssignSlot`.

## Testing
- The backend slot-clear-on-retire is already covered by `tests/spool-slots.test.ts`; this PR is the client-side refresh that surfaces that cleared state. The new path mirrors the existing `handleAssignSlot` refresh.

Fixes #558

### DIFF
--- a/src/app/filaments/[id]/page.tsx
+++ b/src/app/filaments/[id]/page.tsx
@@ -515,6 +515,15 @@ function FilamentDetail() {
     }
   };
 
+  // Re-pull printers so every spool card's *derived* AMS-slot assignment
+  // (read off amsSlots[].spoolId, not stored on the spool) reflects the
+  // latest server state. Used after any write that the server may have
+  // reconciled slots for — slot assign/clear and spool retire (#558).
+  const refreshPrinters = async () => {
+    const pr = await fetch("/api/printers");
+    if (pr.ok) setPrinters(await pr.json());
+  };
+
   const handleUpdateSpool = async (
     spoolId: string,
     data: {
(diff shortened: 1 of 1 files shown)


hyiger:
@codex review

PR #567, opened 2026-06-03
fix(nfc): friendly empty-tag prompt for blank/unformatted tags (#556)

hyiger:
## Summary
Placing a blank/unformatted NFC tag on the reader surfaced the raw low-level error **`Invalid CC magic byte: 0x0`** in the NFC Read Error dialog (#556).

## Root cause
A blank tag reads back all-zero memory, so its NFC-Forum CC byte is `0x00`. `parseNdefFromTag` threw the generic `Invalid CC magic byte` error for any non-`0xE1` first byte — *before* the no-NDEF check that the main-process auto-read routes to the friendly empty-tag UI.

## Fix
- `electron/ndef.ts` — distinguish the blank case (`raw[0] === 0x00`) with its own message; a genuinely wrong-format tag (non-zero, non-`0xE1` CC byte) still throws `Invalid CC magic byte`.
- `electron/main.ts` — route the blank-tag message, alongside the existing No-NDEF paths, to the renderer's `empty: true` state, which already renders the friendly **"Empty Tag — … write data to it"** dialog (no new copy needed). The `0x00 == blank` convention already matches the write path's guard in `nfc-service.ts`.

## Testing
- `tests/ndef.test.ts`: new case asserts the blank-tag message for all-zero memory; the wrong-format assertion now uses a non-zero CC byte. Full `ndef` / `nfc-roundtrip` / `nfc-binary-parsing` suites pass.

## Note
This is the auto-read path (placing a tag) that produced the reported dialog. The write succeeds afterward as before — this is first-use UX only.

Fixes #556

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1281,7 +1281,15 @@ async function initNfc(): Promise<void> {
             }
             // Blank/erased tags have no NDEF data — tell the renderer so it
             // can show an "empty tag" indication instead of silently ignoring.
-            if (err.message?.includes("No NDEF TLV") || err.message?.includes("No NDEF record")) {
+            // Covers an erased NDEF-formatted tag (No NDEF TLV/record) and a
+            // never-formatted blank tag whose all-zero memory has no CC byte
+            // (#556) — both are the friendly "write me to initialize" case,
+            // not a raw error worth surfacing.
+            if (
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

PR #568, opened 2026-06-03
fix(api-docs): mount Swagger in its own React root to silence StrictMode dev overlay (#554)

hyiger:
## Summary
Opening `/api-docs` in Electron dev mode tripped the Next.js dev "Issues" overlay with a legacy React lifecycle warning (`Using UNSAFE_componentWillReceiveProps in strict mode is not recommended … ModelCollapse`), making the page feel broken during local QA (#554).

## Root cause
`swagger-ui-react`'s internal components still use the legacy `UNSAFE_componentWillReceiveProps` lifecycle. Under the app's React StrictMode (Next's default), React logs the warning and the dev overlay surfaces it as a blocking issue.

## Why not the obvious fixes
- **#321 already removed** a global `console.error` monkey-patch (it intercepted *all* logging app-wide) and deliberately left the warnings — so re-patching console is out.
- Disabling `reactStrictMode` app-wide would drop StrictMode's dev checks everywhere — a real regression for a P3 third-party warning.

## Fix
Mount `SwaggerUI` in its **own React root** (`createRoot` in an effect). A root created here is not a descendant of the StrictMode that Next wraps the app root in, so swagger's legacy lifecycles run warning-free while the rest of the app keeps StrictMode. Details:
- A fresh child node per effect run avoids the "container already passed to createRoot" warning under StrictMode's dev double-invoke.
- The unmount is deferred to a microtask to dodge the "synchronous unmount during render" warning on teardown.
- No new dependency, no CSP change, no console patching. CSS/UI unchanged (`<style jsx global>` rules are global and still match).

## Verification (manual, `next dev`)
- `/api-docs` renders all **95** operations, title `Filament DB API 1.34.5 OAS 3.0`.
- Console: **no warnings/errors** (previously the UNSAFE lifecycle warning fired and tripped the overlay).
- Exactly one `.swagger-ui` node after StrictMode's double-invoke — no orphaned root.

Fixes #554

### DIFF
--- a/src/app/api-docs/page.tsx
+++ b/src/app/api-docs/page.tsx
@@ -1,21 +1,60 @@
 "use client";
 
-import dynamic from "next/dynamic";
+import { useEffect, useRef } from "react";
+import type { Root } from "react-dom/client";
 import Link from "next/link";
 import "swagger-ui-react/swagger-ui.css";
 import { useTranslation } from "@/i18n/TranslationProvider";
 
-const SwaggerUI = dynamic(() => import("swagger-ui-react"), { ssr: false });
-
 export default function ApiDocsPage() {
   const { t } = useTranslation();
+  const hostRef = useRef<HTMLDivElement>(null);
 
-  // GH #321: this page used to monkey-patch the global `console.error`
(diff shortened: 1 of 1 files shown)


hyiger:
@codex review

PR #576, opened 2026-06-03
fix(form): snap NFC-decoded density/diameter to the input step (#570, P1)

hyiger:
## Summary
Reading a programmed tag via **Add Filament → New Filament from NFC Tag** filled **Density = `1.2392578125`**, and the density input's `step="0.01"` then rejected it ("the two nearest valid values are 1.23 and 1.24") — so an NFC-read filament **could not be saved** without hand-editing the field (#570, P1).

## Root cause
OpenPrintTag stores `density` and `filament_diameter` as **CBOR half-floats** (10-bit mantissa → 1/1024 resolution). A tag programmed with density `1.24` decodes to `1.2392578125`; a `2.85` mm diameter decodes to `2.849609375`. The form threaded those raw values straight into its `step="0.01"` density/diameter inputs. (The scan modal only *looked* correct because it displays `toFixed(2)`.)

## Fix
Decode must stay lossless — the precision/round-trip decode tests assert the exact float (e.g. `toBeCloseTo(1.2345678, 4)`), so rounding at decode is out. The fix lives at the **form's inbound conversion**: snap `density` and `diameter` to the field's `0.01` step via a new pure helper `snapToStep(value, step)`. Snapping an already-clean stored value (`1.24`, `2.85`, `1.75`) is a no-op, so edit / clone / TDS paths are unaffected.

## Verification (`next dev`)
`/filaments/new?from_nfc=1&name=Test&vendor=Acme&density=1.2392578125&diameter=2.849609375`:
- density input → `1.24`, diameter input → `2.85`, both `checkValidity() === true`
- whole form `checkValidity() === true` → submits

## Tests
`tests/snapToStep.test.ts` (9 cases): half-float density/diameter artifacts, floating-point dust trimming, integer + finer steps, and the non-finite/non-positive-step guards. ESLint + `tsc` clean.

Fixes #570

### DIFF
--- a/src/app/filaments/FilamentForm.tsx
+++ b/src/app/filaments/FilamentForm.tsx
@@ -6,6 +6,7 @@ import { useTranslation } from "@/i18n/TranslationProvider";
 import CollapsibleSection, { expandAndScrollToSection } from "@/components/CollapsibleSection";
 import FormToc, { FormTocMobileButton, type TocEntry } from "@/components/FormToc";
 import FilamentSwatch from "@/components/FilamentSwatch";
+import { snapToStep } from "@/lib/snapToStep";
 import {
   filterColorSuggestions,
   lookupCssNamedColor,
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

PR #578, opened 2026-06-03
fix(list): clarify filament counts + filter/empty/sort QoL (#573, #575)

hyiger:
## Summary
Filament-list polish from the v1.34.6 QA pass. Fully resolves the count-confusion bug (#573) and addresses three of the seven QoL items in #575 (the rest land in follow-up PRs).

### #573 — counts don't agree across the app
Dashboard showed 74, the list header 63, CSV export 75. The list collapses each parent's variants into a group, so the headline (parents excluded) legitimately differs from per-record totals. The header now reads **"63 filament(s) · 74 incl. variants"** when variants exist, so the gap is self-explanatory instead of looking broken.

### #575.1 — stale type/vendor counts on filter
"N type(s) · N vendor(s)" were taken from the global filter-dropdown options, so they stayed frozen while searching/filtering. They now derive from the currently-fetched set (which already reflects the search/type/vendor filters).

### #575.2 — no empty state
A client-side quick filter (e.g. **Low stock**) matching nothing rendered a bare header-only table. Now shows **"No filaments match the current filters."**

### #575.6 — sort puts blanks first
Ascending sort listed null/blank values (e.g. un-priced filaments) above the lowest real value. The comparator now sinks nulls/blanks to the bottom **regardless of direction**.

## Changes
- `src/app/page.tsx` — dual header count, memoized filtered type/vendor counts, empty-state branch.
- `src/lib/sortFilamentList.ts` — `getSortValue` returns `null` for missing numerics; `compareFilaments` sorts nulls/blanks last in both directions.
- `src/i18n/locales/{en,de}.json` — `filaments.stats.totalWithVariants`, `filaments.noMatch`.

## Verification
- `next dev`: header reads "63 filament(s) · 74 incl. variants · 21 type(s) · 17 vendor(s)"; the Low-stock (0-match) filter shows the empty message.
- `tests/sortFilamentList.test.ts` updated for nulls-last (94 cases pass); ESLint + `tsc` + i18n key-coverage clean.

Fixes #573
Addresses #575 (items 1, 2, 6 — items 3/4/5/7 in follow-up PRs)


hyiger:
@codex review

Issue #572, opened 2026-06-03
NFC reader doesn't detect a tag already resting on it (only on fresh tap)


hyiger:
**Version:** v1.34.6 (macOS desktop app)

### Summary
A tag placed on the reader *before* entering a read context is never detected. On the Filaments list nothing happens, and on the "New Filament from NFC Tag" form no fields populate. Detection only fired after the tag was physically lifted and re-tapped — and in testing it only began working after the reader was unplugged and re-plugged. Suggests reads trigger solely on the tag-arrival edge and/or the reader session can get into a stuck state.

### Steps to reproduce
1. Rest a programmed tag on the reader.
2. Navigate to the Filaments list (status badge reads "Ready — place tag") — no reaction.
3. Open **+ Add Filament** — form does not populate.
4. Lift the tag and tap it again — only now does it read (after a reader re-plug in our case).

### Expected
On entering a read context, the app should scan for / detect a tag that is already present, not only react to the arrival event. The reader should also auto-recover from a stuck session.

PR #581, opened 2026-06-03
fix(nfc): clarify the ready-state badge's behavior in its tooltip (#575 item 7)

hyiger:
## Summary
Completes **#575**. Items 1–6 shipped in v1.34.6 (#578/#579); this addresses the last item.

### #575.7 — "Ready — place tag" badge over-promises
The pill reads as a promise that any tag near the reader gets picked up, but a read only fires on a **fresh placement** (the arrival edge), and the scan result opens **app-wide**, not only inside the Add form. Added a tooltip hint to the ready state:

> *"Place a tag on the reader to read it — the result opens anywhere in the app. A tag already resting on the reader may need to be lifted and placed again."*

"place" is kept (not "tap") because this is a hold-on-reader desktop device (ACR1552U), where you rest the tag on the reader.

This is the **wording** half of item 7. Actually *detecting* an already-resting tag (and recovering a stuck reader session) is the deeper reader-session behavior tracked in **#572** (deferred to a hardware-in-the-loop session).

## Changes
- `src/components/NfcStatus.tsx` — per-state `hint` appended to the pill's tooltip for the ready state.
- `src/i18n/locales/{en,de}.json` — new `nfc.status.readyPlaceTagHint`.

## Notes
- Renderer-only; no change to NFC read/write mechanics. `NfcStatus` renders only in Electron, so this can't be exercised in the web preview — verified by inspection + the i18n key-coverage parity test. ESLint + `tsc` clean.

Closes #575

### DIFF
--- a/src/components/NfcStatus.tsx
+++ b/src/components/NfcStatus.tsx
@@ -41,6 +41,13 @@ export default function NfcStatus() {
 
   let dotColor: string;
   let label: string;
+  // Optional plain-language hint appended to the tooltip for states whose
+  // short pill label can read as over-promising. GH #575 item 7: "Ready —
+  // place tag" implies a tag resting on the reader will be picked up, but a
+  // read only fires on a fresh placement (the arrival edge) — detection of an
+  // already-resting tag is tracked separately in #572. The hint sets accurate
+  // expectations without bloating the space-constrained pill.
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

PR #590, opened 2026-06-04
fix(label): scope lpinfo to the usb backend to stop the list-devices IPC timeout

hyiger:
**Regression in v1.34.9** (the #588 label-USB transport). Hit live: Settings → Label Printer errored with `Couldn't list printers: IPC timeout: label-printer-list-devices took longer than 15000ms`.

## Cause
`listLabelPrinters()` ran a bare `lpinfo -v`, which enumerates **all** CUPS backends — including the network ones (`snmp`/`dnssd`) that probe the LAN with their own timeouts. Measured on the affected machine:
- `lpinfo -v` (all backends): **13.59s** — right at the 15s IPC ceiling, tips over with any overhead
- `lpinfo --include-schemes usb -v`: **0.12s**, still returns `usb://Brother/PT-P710BT?serial=…`

(My CLI smoke test happened to hit a fast path, so this slipped through.)

## Fix
Restrict `lpinfo` to the USB backend with `--include-schemes usb`. We only parse `usb://` lines anyway, so it's both ~100× faster and strictly correct. Verified end-to-end: `listLabelPrinters()` now returns in **118ms** and still surfaces the Brother (badged).

Regression test pins that `lpinfo` is called with `--include-schemes usb`.

## Verification
- ✅ Listing 118ms (was 13.6s+); 12 label tests pass; eslint + electron:compile clean.

Will ship as **v1.34.10**.


hyiger:
@codex review

hyiger:
@codex review

PR #612, opened 2026-06-09
feat(openprinttag): check filaments for community updates (#607, Phase 1)

hyiger:
Phase 1 of [#607](https://github.com/hyiger/filament-db/issues/607) — re-sync settings from OpenPrintTag. Implements the manual "check this filament for updates" loop from the [plan posted on the issue](https://github.com/hyiger/filament-db/issues/607); the weekly job, bulk surface, per-filament ignore, and global toggle are deferred to Phase 2/3.

## What it does

A filament imported from the OpenPrintTag community DB already carries its source linkage (`settings.openprinttag_slug` + `_uuid`), but nothing read it back. Now a teal **"Check for updates"** button on the detail page (shown only when the filament is OPT-linked) compares it against the *current* upstream material and lets you adopt the changes you want.

## The crux: not clobbering your edits

Import and each sync stamp `settings.openprinttag_snapshot` — what OPT last offered, per field. On re-check each differing field is classified:

- **adopt** — the field was unset, still holds the gray `#808080` sentinel, or still equals what OPT last wrote → pre-checked, safe to take.
- **conflict** — the local value diverged from the snapshot, i.e. *you* edited it → badged "edited", left unchecked, never auto-applied.

The snapshot records the OPT offer regardless of which fields you accept, so a *declined* field stays a conflict on the next check rather than silently re-adopting.

Only the value fields are managed (color, secondary colors, density, the OPT-carried temps, drying temp/time, Shore D, transmission distance, tags). **Identity (name/vendor/type) and diameter are never re-synced**; spools/calibrations/usage history are never touched.

## Surface area

- `src/lib/optResync.ts` — pure, DB-free, unit-tested (diff / classify / snapshot / whitelisted `$set`).
- `GET /api/filaments/{id}/openprinttag/check` — read-only diff.
- `POST /api/filaments/{id}/openprinttag/sync` — CSRF-guarded; applies the accepted subset, rejects unknown field keys with 400, refreshes the snapshot under `runValidators`.
- `/api/openprinttag/import` now writes the snapshot on **both** the create and `$set` update paths, so provenance exists from day one (not just after a first manual sync).
- `OptResyncDialog` — focus-trapped, Escape-to-close, adopt-vs-conflict checkboxes. 31 new `resync.*` i18n keys (EN/DE in lockstep — the parity invariant passes).

## Tests
- `tests/optResync.test.ts` — 16 pure-helper cases (classification matrix, snapshot, whitelist, order-sensitive arrays, color sentinel).
- `tests/opt-resync-route.test.ts` — 11 route cases: check (unlinked / slug-gone / conflict+adopt / up-to-date), sync (apply-subset + snapshot-refresh, unknown-field 400, not-linked 400, bad body 400, **CSRF 403 before mutating**), and import-snapshot on **both** create and update paths.
- `npm test` → 1870 passing; lint + tsc clean.

## Docs
CLAUDE.md (feature section + project layout), `docs/api.md`, `docs/usage.md` (a "Checking for Community Updates" subsection), and the OpenAPI spec.

## Note on local verification
I couldn't drive the browser preview to confirm the button end-to-end: a packaged Filament DB app was running and squatting `[::1]:3456` (IPv6 localhost) with the old bundle + working DB, while `next dev` on `127.0.0.1:3456` had the new code but its `.env.local` DB wasn't reachable. The logic is fully covered by the deterministic tests above. **Holding this open for you to test on your own build before merge, as requested** — quit the running app (or use `127.0.0.1` explicitly) when you do.


hyiger:
@codex review

hyiger:
Addressed Codex's P2 (surface explicit OPT clears) in 9858718. The `hasIncoming` guard was dropping any field whose current OpenPrintTag value was null/`[]`, so an explicit upstream clear never surfaced. Introduced `OPT_CLEARABLE_FIELDS = {color, secondaryColors, optTags}` — for those, an explicit null/`[]` is now diffed as a real change (covers single-color → coextruded with `color: null` + populated secondaries, plus secondaries/tags being dropped). Scalar fields OPT merely lacks (density, temps, …) still skip, so a sparse upstream entry can't wipe good local data. Also: the gray `#808080` sentinel is never pushed as a color, `valuesEqual` treats null ≈ `[]` to avoid a spurious no-tags-vs-empty diff, and `classify` adopts an empty-array local value as a gap-fill. Fixed a latent bug found in passing — `getPath` stringified array elements, which would have written string tags into the `optTags: [Number]` schema on sync. 7 new unit cases; full suite 1877 green.

hyiger:
@codex review

hyiger:
Addressed Codex's 2nd P2 (object in settings crashes the render) in d7de164. Moved the provenance snapshot out of `settings.openprinttag_snapshot` to a dedicated top-level `openprinttagSnapshot` field (`Schema.Types.Mixed`) — import (create + update), check, and sync all read/write the new field, so nothing structured lands in the `settings` bag that the detail-page settings table renders as a raw React child (or that rides into slicer exports). Belt-and-braces: that table now `JSON.stringify`s any non-scalar value instead of rendering it raw, so a future Mixed-bag entry can't reintroduce the crash. Tests updated to assert the snapshot on the new field (and that it's NOT in settings); docs updated. Full suite 1877 green.

hyiger:
@codex review

hyiger:
Addressed Codex's 3rd P2 (sync could clear a field OPT isn't offering) in 32c2fd8. The sync route now recomputes `diffOptFields` and rejects (400) any requested field that isn't in the live changelist before writing — so a stale/crafted POST of `fields:["density"]` against an upstream material with `density: null` can no longer wipe the user's local density (the sparse-data guard `diffOptFields` already enforces is now enforced on the write path too). Clearable clears (`color: null`, secondaries/tags `[]`) still apply because they DO appear in the diff. New route test: `density: null` upstream + `["density"]` → 400, local value untouched. Full suite 1878 green.

hyiger:
@codex review

hyiger:
Addressed Codex's 4th P2 (dead button on inheriting variants) in 834e6f5. The button was gated on the *resolved* `settings.openprinttag_slug`, which `resolveFilament` shallow-merges from the parent — so every variant of an OPT-linked parent showed a button that the check/sync routes (reading the raw child row) would answer with `linked:false`/400. The detail GET route now computes `_hasOwnOptLink` from the RAW doc before the settings merge, and the button gates on that. New route test confirms it's true for the linked root and false for an inheriting variant (whose resolved settings still carry the inherited slug). Full suite 1879 green.

hyiger:
@codex review

PR #644, opened 2026-06-09
fix(api): make print-history delete refund idempotent across retries

hyiger:
## Summary

`DELETE /api/print-history/{id}` refunds spool weight per filament (`await filament.save()` inside the loop) and only writes the `_deletedAt` tombstone after the loop. The refund was unconditional — not keyed to the matching `jobId` usageHistory entry still being present. If filament #2's save threw mid-loop (VersionError → the route's documented 409 "Please retry"), filament #1 was already refunded with its usageHistory entry removed while the job stayed active, so the advertised retry refunded filament #1 **again** — unbounded when `netFilamentWeight` is null, since the gross-capacity clamp only applies when capacity is known.

Fixes #621

## Fix

Each loop iteration now locates the usageHistory entry the usage row pays back **before** touching any weight, and applies the refund only when an entry is actually removed in this pass. Match preference (exactly one entry consumed per usage row):

1. **jobId + grams** — a job can carry multiple usage rows against the same spool (POST pushes one entry per row, all sharing the jobId), so each row must consume its own entry. The old code swept *every* jobId match at once; under refund-only-if-removed conditioning that sweep would have left later rows with nothing to remove and skipped their refunds, so the sweep became a one-entry-per-row match. End state for data POST creates is identical (N rows ↔ N entries).
2. **jobId alone** — grams drifted; jobId is still unambiguous.
3. **Legacy `(grams, startedAt)`** — pre-v1.12.x entries without a jobId, still restricted to `source: "job" | "slicer"` (manual logs sharing both fields stay protected) and still first-match-only.

No match → `continue` (no refund, no save). On retry, filament #1's entry is already gone → nothing removed → no second refund; the retry finishes filament #2 and lands the tombstone. The GH #228/#229 gross-capacity clamp logic and the parent-inheritance lookup are unchanged — only the ordering/conditioning moved.

## Judgment call: refund-without-entry removed

A usage row whose spool has **no** matching usageHistory entry at all used to get a refund anyway. This change removes that refund — deliberately:

- Refunding without removing a ledger entry is precisely the unbounded double-refund shape from #621 (every repeat of the same delete-shaped call inflated the weight again).
- In the orphan-PrintHistory-vs-manual-entry case (the existing "does not remove a manual entry even when fallback runs" fixture), the old refund drifted the spool weight out of sync with the surviving manual entry that analytics still counts.
- I checked `tests/print-history*` and `tests/api-validation-and-semantics.test.ts`: no existing test pinned the spool **weight** in that scenario (the manual-entry test asserts only the surviving entry). A new test now pins the corrected behavior (weight unchanged when nothing is removed).

## Test plan

- New: `retry after a mid-loop save failure refunds each filament exactly once (GH #621)` — two filaments in one job, second `save()` throws VersionError via the same prototype-patch technique as `tests/print-history-concurrency.test.ts`, first DELETE returns 409, retry returns 200; filament A ends at exactly its pre-job weight (pre-fix: +50g extra), filament B is refunded, tombstone lands.
- New: `refunds every usage row when a job carries multiple rows against the same spool` — pins the one-entry-per-usage-row semantics (a remove-all sweep would under-refund the second row).
- New: `does not refund a usage row whose spool has no matching usageHistory entry (GH #621)` — pins the judgment call above.
- Existing happy-path, legacy-fallback, manual-protection, gross-clamp, variant-inheritance, and tombstone-idempotency (repeat DELETE → 404) tests all pass unchanged.
- `npm run lint` clean; full suite: **111 test files / 1882 tests passed**.

Note: CI's "Security audit" step fails pre-existing (#617, separate PR).

### DIFF
--- a/src/app/api/print-history/[id]/route.ts
+++ b/src/app/api/print-history/[id]/route.ts
@@ -159,6 +159,11 @@ export async function PUT(
  * This handles the "print failed, undo this entry" case from issue #92. The
  * refund is best-effort: if a spool has since been deleted we log the refund
  * loss but still remove the history entry.
+ *
+ * GH #621: each usage row's refund is conditioned on its matching
+ * usageHistory entry actually being removed in this pass, so a retry after a
+ * mid-loop failure (the 409 path) can't refund an already-refunded spool a
+ * second time. See the comment block inside the refund loop.
  */
 export async function DELETE(
   request: NextRequest,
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #387, opened 2026-05-25
Add Bambu Studio filament-preset importer

hyiger:
Completes the round-trip for Bambu Studio users: export from the app → calibrate in Bambu → re-import the values without retyping. Inverse of the existing `GET /api/filaments/{id}/bambustudio` export.

## Endpoints

- **`POST /api/filaments/bambustudio`** — bulk import, upserts by name. Accepts `multipart/form-data` (UI flow) or `application/json` (script flow).
- **`POST /api/filaments/{id}/bambustudio`** — sync into a specific filament. Pinned by id; the parsed `name` is ignored. Powers the "Sync from Bambu Studio" button on the filament detail page.

## Calibration auto-detect

Bambu calibration values live IN the filament preset. The route parses `printer_settings_id` (`"Vendor Model 0.4 nozzle"`), looks up a Printer doc whose name or `manufacturer + printerModel` matches, and picks the unique installed nozzle at the parsed diameter:

- ✅ match → `calibrations[]` row written tagged with `(printer, nozzle)`
- ❌ no printer match → top-level update carries `maxVolumetricSpeed`; response has `calibrationUnresolved: true` so the UI can prompt
- ❓ >1 nozzle at that diameter (e.g. Brass + ObXidian both 0.4 on the same printer) → punt rather than guess

## UI

- `/import-export` page → new **Bambu Studio (.json)** tile next to the existing INI/CSV/XLSX importers.
- Filament detail page → **Sync from Bambu Studio** button next to the existing Export-for-slicer dropdown. Page re-fetches after import so the new values appear immediately.

## Architecture

- `src/lib/bambuStudioImport.ts` — pure parser, DB-free, inverts `filamentToOrcaSlicerKeys`. Bambu Studio = OrcaSlicer JSON + `from: "User"`, so the schemas are identical.
- `src/lib/bambuStudioApply.ts` — applier helper shared by both routes (DB lookups for printer/nozzle matching). Split so the parser stays unit-testable without `mongodb-memory-server`.

## Tests

- `tests/bambuStudioImport.test.ts` — **12 parser cases** including the round-trip invariant (export via `generateOrcaSlicerProfiles` → re-parse → every field matches).
- `tests/bambustudio-route.test.ts` — **11 route cases** covering multipart + JSON dispatch, upsert by name, id-pinned target, calibration auto-detect, unresolved flag, required-field validation.

## i18n

10 new keys added to both `en.json` and `de.json` (count parity preserved at 1211).

## Test plan
[...]

PR #645, opened 2026-06-09
fix(api): enforce validators and atomic calibration writes on slicer per-id sync

hyiger:
## Summary

Three related fixes in the slicer per-id sync paths.

### 1. PrusaSlicer sync write skips validators
`POST /api/filaments/{id}` wrote `{ $set: update }` with no `runValidators`, so the GH #337 numeric validators never fired — `filament_cost = "-3"` parseFloat'd clean and persisted a negative cost the regular PUT would have rejected. The write now passes `{ runValidators: true, context: "query" }` (the same options the Bambu sync route adopted via Codex P2 on #387), and a `ValidationError` surfaces as a JSON 400 via `errorResponseFromCaught` instead of a raw 500.

### 2. OrcaSlicer sync had no type checks either
`POST /api/filaments/{id}/orcaslicer` accepted `cost: "abc"` (CastError → generic 500) and persisted negative numbers silently. Now:
- Numeric body fields (`density`, `cost`, `diameter`, `maxVolumetricSpeed`, and each `temperatures.*` sub-field) are coerced through a finite-number check — numeric strings are accepted (matching the cast Mongoose applies on save), anything else 400s naming the field.
- The write runs with `{ runValidators: true, context: "query" }` so the #337 `min`/`max` validators fire.
- The inverted-nozzle-range guard the regular PUT enforces (#574/#577) now runs here too, reusing `effectiveNozzleRangeForUpdate` / `inheritNozzleRangeFromParent` / `isInvertedNozzleRange` — covering the merge-with-stored-endpoint case and a variant inheriting an endpoint from its parent. Gated on the sync actually touching a range endpoint so pre-existing bad data can't 400 an unrelated sync.

### 3. Own-calibrations whole-array read-modify-write race
The own-calibrations branch of the PrusaSlicer sync spread `[...calTarget.calibrations]`, mutated, and `$set` the whole array — two concurrent syncs for different nozzles could drop each other's entries. GH #265 already made the *parent*-targeted branch atomic per-entry; this PR unifies both branches onto that construction: an `$elemMatch`-filtered positional `$set` for the existing-entry case, a conditional `$push` for the new-entry case, and the step-3 fallback `$set` for the lost-insert race. Per-nozzle calibration context and `high_flow` disambiguation are unchanged.

Fixes #618

## Test plan
- [x] New regression tests in `tests/api-route-correctness.test.ts` (#618 describe block):
  - PrusaSlicer sync with negative `filament_cost` → 400, nothing persisted
  - OrcaSlicer sync with `cost: "abc"` → 400 (was 500), `cost: -3` → 400 via schema validators
  - OrcaSlicer inverted nozzle range → 400 (direct, lone-min vs stored max, lone-min vs inherited parent max)
  - OrcaSlicer happy path still applies valid values incl. numeric-string coercion and stored-temp merge
  - Sequential per-nozzle calibration syncs: append for a new nozzle, in-place update on re-sync, both entries present (pins the `$elemMatch`/`$push` shape)
- [x] `npm run lint` clean
- [x] `npm test` — 111 files, 1887 tests, all passing

Note: the failing "Security audit" CI step is pre-existing (#617, separate PR).

### DIFF
--- a/src/app/api/filaments/[id]/orcaslicer/route.ts
+++ b/src/app/api/filaments/[id]/orcaslicer/route.ts
@@ -9,6 +9,11 @@ import {
 import { errorResponse, errorResponseFromCaught } from "@/lib/apiErrorHandler";
 import { assertSameOriginRequest } from "@/lib/requestGuard";
 import { mergeSlicerSettings } from "@/lib/slicerSettings";
+import {
+  isInvertedNozzleRange,
+  effectiveNozzleRangeForUpdate,
+  inheritNozzleRangeFromParent,
+} from "@/lib/temperatureRange";
 
 /**
  * Top-level body keys that map to structured Filament DB fields.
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

PR #647, opened 2026-06-10
fix(ui): SSR-safe locale/currency init, stale edit drafts, fetch error handling

hyiger:
### #639 — hydration mismatch for locale/currency on web
`TranslationProvider` and `useCurrency` read localStorage inside `useState` initializers, so SSR rendered `en`/`USD` while the first client render produced the stored value — a React 19 hydration error + full client re-render on every page load for any non-default web user. Both now default during SSR and sync in a post-mount effect (the documented CollapsibleSection pattern); the Electron electron-store hydration is async IPC and still resolves after, so desktop behavior is unchanged. The brief default-locale flash on web is the accepted trade-off — no loading gate introduced.

### #640 — four state/error-handling fixes
- **Inventory weight editor stale draft**: reseeds the draft from the current row value when opening the editor (GH #263 SpoolCard pattern) — a sync/edit from elsewhere can no longer be silently overwritten with a stale value.
- **PrinterForm dirty latch**: dirty is now derived from a baseline snapshot (GH #286 FilamentForm pattern), so reverting an edit no longer triggers the unsaved-changes prompt. Scoped to the dirty logic only — the conflict-modal region is untouched (a sibling PR works there).
- **Fetch handlers without error handling**: `handleIniFile`/`handleCloneFetch` (new-filament page), trash `handleRestore`/`handleEmptyTrash`/`handleRestoreAll` (bulk loops now report partial-failure counts via a new `trash.bulkFailures` key, en+de), share `handlePublish`/`handleUnpublish`, and `refreshPrinters` now surface failures through the pages' existing toast/error patterns instead of rejecting silently.
- **Post-Prusament-import refresh**: gates on `r.ok` so a non-2xx response can't write error JSON into `filament` state and crash the detail-page render.

Fixes #639
Fixes #640

## Test plan
- `npm run lint` → clean
- `npm test` → 111 files / 1879 tests passed
- i18n key parity maintained (1 new key in both en + de)

### DIFF
--- a/src/app/filaments/[id]/page.tsx
+++ b/src/app/filaments/[id]/page.tsx
@@ -272,10 +272,16 @@ function FilamentDetail() {
   }, [params.id]);
 
   // GH #607: re-pull the filament after an OpenPrintTag re-sync applied
-  // changes, so the detail view reflects the adopted values.
+  // changes, so the detail view reflects the adopted values. GH #640:
+  // never throws — a failed refresh keeps the current data rather than
+  // surfacing an unhandled rejection from fire-and-forget callers.
   const refetchFilament = useCallback(async () => {
-    const r = await fetch(`/api/filaments/${params.id}`);
-    if (r.ok) setFilament(await r.json());
+    try {
+      const r = await fetch(`/api/filaments/${params.id}`);
+      if (r.ok) setFilament(await r.json());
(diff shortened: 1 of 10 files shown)


hyiger:
@codex review

PR #648, opened 2026-06-10
fix(api): soft-delete consistency in aggregations, delete guards, and resolveFilament

hyiger:
### #625 — hasVariants probe missed legacy variants lacking `_deletedAt`
In aggregation `$expr`, `{ $eq: ["$missingField", null] }` is FALSE — missing is its own BSON type (the documented v1.32.2 quirk). Pre-v1.15 variant docs that never gained the `_deletedAt` field made their parent report `hasVariants: false`, dropping the composite parent swatch while find()-based routes still treated the variant as active. The probe now wraps in `$ifNull` (the `spools/by-location` pattern). Regression test inserts a raw variant doc with the field `$unset` and asserts the parent's `hasVariants`.

This is the likely root cause of the cross-hatch swatch reports — Refs #597, Refs #605 (left open for the reporter to confirm on real data after this merges).

### #629 — delete guards ignored trashed referrers; OPT sync write race
- The location/nozzle/printer/bed-type delete guards now count **trashed** filament referrers too (only `_purged: { $ne: true }` tombstones are excluded), so deleting an entity referenced by a trashed filament is refused instead of leaving a dangling ref that restore resurrects. Error messages updated; tests cover the trashed-referrer-blocks and purged-tombstone-doesn't-block cases.
- The OPT sync route's final write re-filters `_deletedAt: null` and 404s when the filament was trashed between the initial read and the write (mirrors the Bambu per-id sync fix).

### #633 — resolveFilament allowlist drift
`lowStockThreshold` joins `INHERITABLE_FIELDS` (cost/density posture); `openprinttagSnapshot` and `_purged` join `VARIANT_ONLY_FIELDS`. A new drift-guard test diffs the Filament schema's top-level paths against the allowlist union (with an explicit exclusion list for specially-handled fields), so the next schema addition fails loudly instead of silently vanishing from resolved variants.

Fixes #625
Fixes #629
Fixes #633
Refs #597
Refs #605

## Test plan
- `npm run lint` → clean
- `npm test` → 111 files / 1897 tests passed (+18 new; one unrelated toast.test.ts flake under parallel-suite load passed in isolation and in the clean re-run)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #649, opened 2026-06-10
fix(import): prusament error handling, row caps, per-row isolation, CSV/XLSX round-trip fidelity

hyiger:
### #622 — Prusament import robustness
The route's DB writes were entirely unguarded (only `dbConnect` and the body parse had try/catch) and several spool fields were never validated. Now: handler-wide try/catch via `errorResponseFromCaught` (ValidationError → 400), validation for `priceUsd` / nozzle+bed temp ranges / `pageUrl` (http(s) only, tdsUrl posture), a trashed-row resurrect phase, and E11000 create-race recovery — a lost race recovers as add-spool against the winner; an unrecoverable duplicate maps to 409. New `tests/prusament-import-route.test.ts` (the route had zero tests).

### #627 — import/export robustness sweep
- **Row caps**: 10k on filament CSV/XLSX import (mirroring the INI importer's #297 cap); 1k id cap on import-atlas.
- **Per-row isolation**: `upsertImportRows` wraps each row's create/update — duplicate-key errors land in `skippedRows` with a named reason instead of aborting the batch mid-commit with rows already written (GH #370 spool-importer pattern).
- **Formula-prefix round-trip**: `unsanitizeCsvCell` now runs on name/vendor/colorName/spoolType/parentName, so an exported `'+95A TPU` re-imports as `+95A TPU` instead of creating a corrupted duplicate.
- **Bare CR quoting**: `csvCell` quotes cells containing `\r` (RFC 4180) — `parseCsv` treats bare CR as a row terminator, so an unquoted CR split the row on round-trip. One pre-existing assertion updated to the new (correct) prefix+quote output.
- **XLSX parity**: XLSX export string cells get the same formula-prefix neutralisation as CSV (defense-in-depth — ExcelJS already stores these as shared strings, not formulas).

### #628 — variant inheritance preserved on re-import
Export flattens variants through `resolveFilament` (correct), but re-importing an existing variant `$set` every value, silently severing the GH #106 live-inheritance link. The update path now compares each inheritable field against the parent's resolved value: equal → skip (keep inheriting), reverting-to-parent → `$unset`, genuinely different → `$set`. Mirrors `setIfNotInherited` semantics from the Bambu applier, including `temperatures.*` dot-keys and the empty-string-is-missing rule. Round-trip test: export parent+variant → re-import → parent edits still propagate.

Fixes #622
Fixes #627
Fixes #628

## Test plan
- `npm run lint` → clean
- `npm test` → 112 files / 1916 tests passed (+37 new)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

Issue #503, opened 2026-06-02
Filament `color` field lacks the hex validator that `secondaryColors` has


hyiger:
`src/models/Filament.ts:223` declares the primary color as `color: { type: String, default: "#808080" }` with no validator, while `secondaryColors` immediately below (lines 228-242) has a strict per-entry `^#[0-9A-Fa-f]{6}$` validator plus a max-5 cap. The schema docstring at lines 218-222 claims `Hex shape is validated on writes via the route handlers (matches the existing posture for free-form strings)` — that claim does not hold.

**Verified at HEAD (72a4e05):**
- POST handler at `src/app/api/filaments/route.ts:189-275` strips meta fields then calls `Filament.create(body)` without inspecting `body.color`.
- PUT handler at `src/app/api/filaments/[id]/route.ts:151-228` passes `runValidators: true` to `findOneAndUpdate`, but since the schema has no validator on `color`, nothing fires.
- CSV importer at `src/lib/importFilaments.ts:342` writes `doc.color = row.color` unchanged.
- The only `pre()` hooks on the model (`src/models/Filament.ts:418-444`) cover `instanceId` generation and `tdsUrl` scheme validation; no `color` guard.
- No `#[0-9A-Fa-f]{6}` / `isValidHex` / `validateColor` reference exists under `src/app/api/filaments/` or `src/lib/` outside the `secondaryColors` parser line.

The browser UI path is shielded by `<input type="color">` in `FilamentForm.tsx`, so end users with browsers can't trigger this. The bug is reachable via direct API calls, CSV import, Atlas import, snapshot restore, and the hybrid-sync engine's cross-side writes. There is no XSS risk (React escapes inline style attribute values), but arbitrary strings persist in `color` and round-trip into:
- PrusaSlicer / OrcaSlicer / Bambu Studio exports as the `filament_colour=` / `default_filament_colour=` lines.
- NFC OpenPrintTag writes (`src/lib/openprinttag.ts` parses `color` as RGBA — a non-hex string will encode as garbage bytes).
- The `displayColor()` helper at `src/lib/filamentColors.ts` will return the bad string as-is when it's truthy, never falling through to the secondary or sentinel.

**Recommended fix:** add a schema-level validator on `color` that mirrors `secondaryColors` but permits null per the OpenPrintTag spec, and pull the docstring claim back to match reality:

```ts
color: {
  type: String,
  default: "#808080",
  validate: {
[...]

PR #650, opened 2026-06-10
fix(openprinttag): hex/integer validation on encoder and import update paths

hyiger:
### #632 — color validators skipped update paths; rgbaToHex accepted non-hex
- `rgbaToHex` validated only string **length**, so a malformed `color_rgba` in a community YAML (e.g. `"zzzzzzzz"`) became `"#zzzzzz"` and persisted on re-import-update. It now requires `^[0-9a-fA-F]{6}([0-9a-fA-F]{2})?$` before slicing.
- The OPT import route's existing-row update now passes `runValidators: true, context: "query"` so the GH #503 hex validators actually run on that path.
- New `validateColorsInUpdate` pre-hooks on `updateOne`/`updateMany`/`findOneAndUpdate` mirror the existing tdsUrl coverage, so **no** bare update path can persist an invalid `color`/`secondaryColors`. The coextruded null primary remains valid; `$set` nesting and direct-field shapes both handled.

### #634 — encoder input-validation gaps (GH #274 covered temps only)
- `optTags` gains schema validation (entries must be non-negative integers), and the **encoder filters invalid entries instead of throwing** — a legacy stored bad tag can no longer 500 the `.bin` export route or break every NFC write for that filament. Fractional tags are dropped, not rounded (rounding would silently change the tag's meaning).
- `dryingTemperature`/`dryingTime`/`shoreHardnessA`/`shoreHardnessD` are rounded and capped before CBOR uint encoding (`dryingTime` bounded well below 2³²).
- Print temps enforce min ≤ max with the same `Math.min`/`Math.max` ordering bed temps already had — a first-layer temp ≥ nozzle+20 no longer emits an inverted range.
- `parseHexColor` accepts exactly 6 or 8 hex chars — 7-char values no longer silently drop their last char.
- The documented alpha-byte truncation behavior is unchanged (intentional spec gap).

Fixes #632
Fixes #634

## Test plan
- `npm run lint` → clean
- `npm test` → 111 files / 1891 tests passed (+12 new across Filament model, encoder, and browser-parser tests)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #651, opened 2026-06-10
fix(electron): pin powershell path, device-path allowlist, gate status getters, isolate sync closes; value-level CSP parity test

hyiger:
### #623 — four Electron hardening items
- **PATH-resolved `powershell` (Windows binary planting)**: both label-printer call sites now resolve `powershell.exe` via `%SystemRoot%\System32\WindowsPowerShell\v1.0\` through a shared helper — `CreateProcess` searches the application directory and cwd *before* System32, the exact threat `electron/main.ts` already closed for `sc.exe`. Unix CUPS tools stay bare-name (execvp doesn't search cwd; install paths vary across distros) with a comment saying so.
- **Device-path allowlist**: `label-printer-set-device-path` accepts only `usb://` URIs (the one scheme the CUPS lister emits) or installed-queue names without `/` or a scheme; `printCups` refuses to forward non-usb schemes to `lpadmin`, closing the `file://`/`ipp://` queue-binding gadget. Legacy serial-path rejection unchanged.
- **Status getters gated**: `get-sync-status`, `nfc-get-status`, and `update-get-status` now carry `assertTrustedSender` like every sibling handler — closes the undocumented asymmetry (and the `tagUid` read from sub-frames).
- **Sync close isolation**: `sync()`'s `finally` uses `Promise.allSettled([local.close(), remote.close()])` so a local close failure can no longer skip the remote close and leak the Atlas client pool every 5-minute cycle.

### #635 — CSP parity test now diffs values, not names
`tests/csp-parity.test.ts` parsed each CSP into directive **name** sets only — adding `https://cdn.example` to one side's `img-src` would have passed (the very #371 asymmetry the test was built against). It now parses directive → token sets and diffs tokens per directive, with carve-outs limited to the two documented variances: `connect-src` (web tokens must be a subset of Electron's; the extras must be exactly the localhost pair) and the dev-gated `script-src 'unsafe-eval'`.

Fixes #623
Fixes #635

## Test plan
- `npm run lint` → clean
- `npm run electron:compile` → bundles cleanly (electron/ is outside tsconfig)
- `npm test` → 111 files / 1893 tests passed (+14, incl. new label-printer transport cases for the allowlist and absolute-path expectations)

### DIFF
--- a/electron/auto-updater.ts
+++ b/electron/auto-updater.ts
@@ -75,7 +75,13 @@ export function initAutoUpdater(win: BrowserWindow) {
   // "No handler registered". In dev, the mutating handlers short-circuit
   // with a "dev-mode" error instead of touching the real updater, which
   // electron-updater refuses to run when app.isPackaged is false.
-  ipcMain.handle("update-get-status", () => currentState);
+  // GH #623: read-only, but gated like its siblings (#434) — the state
+  // carries version/release-notes/error text, and an open getter was an
+  // undocumented asymmetry vs. the trusted-sender posture everywhere else.
+  ipcMain.handle("update-get-status", (event) => {
+    assertTrustedSender(event, "update-get-status");
+    return currentState;
+  });
 
   // GH #434: update-check and update-download were gated only on
   // `app.isPackaged`. In a release build a sub-frame could drive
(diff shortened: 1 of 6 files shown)


hyiger:
@codex review

PR #653, opened 2026-06-10
fix(ui): dialog a11y, keyboard access, and translated aria labels

hyiger:
### #637 — four a11y / markup fixes
- **PrinterForm nozzle-conflict modal**: declared `role="dialog" aria-modal="true"` but had no keyboard mechanics. Added a focus trap (Tab/Shift-Tab cycle), Escape-to-close (inert mid-resolution), initial focus, and focus restore — matching every sibling dialog.
- **OpenPrintTag browser rows**: the expand target was a plain `<div onClick>`, so keyboard/SR users could tick the checkbox but never open `MaterialDetail`. Now `role="button"` + `tabIndex={0}` + Enter/Space + `aria-expanded`; the checkbox is a sibling that stops propagation, so selection still doesn't toggle expansion.
- **colorName suggestion listbox**: replaced the invalid `<span>`-wrapping-`<li>` structure (inside `role="listbox"`) with a keyed `<Fragment>` and a `role="presentation"` section header, so AT enumerates only the options.
- **MultiColorEditor secondary-color rows**: were keyed by array index, jumping focus in the hex input on a mid-list delete (the exact trap the file's `makeUid` docblock describes). Now keyed by a parallel `uids` list held in `useState`, kept in lockstep by the add/remove handlers, with a render-time length reconcile for external resizes (NFC/clone prefill) via React's adjust-state-during-render pattern.

### #638 — hardcoded English strings → t()
Routed through `t()` with 10 new keys in both locales (parity preserved):
- `FilamentSwatch` color-group / coextruded / gradient / swatch labels (component gained `useTranslation`; the context's en-based default keeps it safe outside the provider)
- color-swatch `aria-label`s on the filament list, detail header, color-variants list, and all four NFC-dialog swatches
- spool photo `alt`, CSV-import row label, Prusament input placeholder
- compare + new-filament Suspense fallbacks (reuse existing `common.loading`)
- share-page nozzle/bed temperature suffixes

Fixes #637
Fixes #638

## Test plan
- `npm run lint` → clean; `npx tsc --noEmit` → clean
- `npm test` → 111 files / 1879 tests passed, including `i18n-key-coverage` (en/de parity at 1477 keys)


hyiger:
@codex review

hyiger:
@codex review

PR #655, opened 2026-06-10
fix(openprinttag): respect variant inheritance in the re-sync diff (#607)

hyiger:
## Summary

Fixes the bug reported on #607 (last comment): running **"Check for updates"** on a *variant* filament offered fields that are actually inherited from its parent as "updates".

### Root cause
`GET /api/filaments/{id}/openprinttag/check` diffed the **raw** variant document (`.lean()`, no inheritance resolution) against the upstream OpenPrintTag material. A field the variant leaves unset to inherit reads as `null` on the raw doc, so `diffOptFields` classified it as an empty local value and offered it as a spurious **`adopt`** gap-fill — even though the resolved (inherited) value already matched OPT.

The sibling `.bin` download route (`openprinttag/route.ts`) already resolves inheritance for exactly this reason; the check route just never got the same treatment. Inheritance was handled for *link detection* (so the button shows on a variant) but not for the *field diff*.

### Fix
Resolve `variant → parent` before diffing, then diff the **effective** values. The slug and provenance snapshot still come from the raw doc (`settings` isn't carried through `resolveFilament`; `openprinttagSnapshot` is variant-only, so it survives resolution). Root filaments have no `parentId` and diff exactly as before — no behavior change for them.

### Note on the report
The reporter cited `diameter`, but `diameter` is hardcoded to `1.75` in `mapToFilamentPayload` and deliberately excluded from `OPT_MANAGED_FIELDS`, so it can't actually be offered — the affected field was an inheritable one (density / a temperature). The fix covers all of them.

Closes the Phase 1 inheritance bug on #607. Phase 2/3 (weekly background check, bulk "updates available" surface, per-filament ignore, global toggle) remain on the roadmap.

## Test plan
- `npm run lint` + `npx tsc --noEmit` → clean
- `npm test` → 114 files / 2007 tests passed (+2 new):
  - a variant inheriting OPT-matching values → **no** spurious changes
  - a variant inheriting a *stale* value → the genuine upstream change is still surfaced, with `current` showing the resolved (inherited) value, not `null`
- Both new tests fail on the pre-fix code.

### DIFF
--- a/src/app/api/filaments/[id]/openprinttag/check/route.ts
+++ b/src/app/api/filaments/[id]/openprinttag/check/route.ts
@@ -6,7 +6,7 @@ import {
   mapToFilamentPayload,
 } from "@/lib/openprinttagBrowser";
 import { diffOptFields } from "@/lib/optResync";
-import { resolveFilament } from "@/lib/resolveFilament";
+import { resolveEffectiveFilament } from "@/lib/resolveEffectiveFilament";
 
 /**
  * GET /api/filaments/{id}/openprinttag/check  (GH #607, Phase 1)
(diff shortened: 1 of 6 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #659, opened 2026-06-11
feat(spool): view a spool's logged usage history (#608)

hyiger:
## Summary

Implements [#608](https://github.com/hyiger/filament-db/issues/608) (crzykidd): the spool card let you *add* a usage entry and showed a count, but never let you *see* the log.

The "Usage history (N)" line on each spool is now an expandable disclosure (only when entries exist). Expanding it lists each logged usage, newest first:

`6/10/26 · 42g · preview-test-print · Manual`

— date · grams · job label · source badge (Manual / Slicer / Job / NFC). The data was already returned on the detail-page GET (full spools, no projection); this just surfaces it. No new endpoint, no schema change.

This pairs with his note on the issue: his Spoolman→filament-db bridge writes usage entries, so being able to view them matters.

## Implementation
- `SpoolCard` gains a `showUsageHistory` toggle; the count line becomes a `▸`/`▾` disclosure button (`aria-expanded`) when `usageHistory.length > 0`, else stays a plain label.
- Entries render in a scrollable list, sorted by date descending.
- 5 new i18n keys (en + de): `detail.spool.usageNoLabel` + `detail.spool.usageSource.{manual,slicer,job,nfc}`.

## Test plan
- `npm run lint` + `npx tsc --noEmit` → clean
- `npm test` (run earlier with this change present) → 114 files / 2012 pass; i18n parity holds (1484 keys each)
- **Preview-verified** against real data: created a usage entry via the API, navigated to the detail page, expanded the disclosure → entry rendered correctly with date/grams/label/source, no console errors. (The test entry was removed from the DB afterward.)


hyiger:
@codex review

PR #669, opened 2026-06-12
fix(ci): keep x64 as the single Windows auto-update channel (#586)

### LINKED ISSUE #586: Restart & Install not working for last few updates - have to do manually v1.34.8
Simply put, for the last weekish of updates, if I do the download, then restart & install and go through the wizard for install.  Which it finishes to launch the app it gives this error as if it didn't get installed:

Image

If I install from a direct download of github (the same x64 installer for windows, it never has an issue)

hyiger:
## Summary

Fixes [#586](https://github.com/hyiger/filament-db/issues/586) (reported by @tonysurma): Windows **Restart & Install** auto-updates intermittently fail with Windows' **"Problem with Shortcut — 'Filament DB.exe' has been changed or moved"** dialog, ~every other release. A manual install of the GitHub x64 `.exe` always works.

## Root cause (verified against the v1.38.0 release logs)

The Windows updater metadata file is **`latest.yml` with no arch suffix** (unlike `latest-mac.yml` / `latest-linux-arm64.yml`). The release matrix builds Windows **x64** and **arm64** in two separate jobs (each must rebuild `pcsclite` for its own arch), and both upload `latest.yml` to the same GitHub release — **last write wins**.

On the v1.38.0 release, the arm64 job uploaded last, so the published `latest.yml` referenced:

```
path: FilamentDB-1.38.0-windows-arm64-setup.exe
```

An **x64** user's `electron-updater` reads that, downloads the **arm64** installer, and installs an arm64 binary on x64 Windows — which can't run → the exe the shortcut points to is broken → the "changed or moved" dialog. **"Every other release"** is simply the two jobs' upload order alternating by CI timing. (electron-builder's own implicit publish is skipped — `existingType=release publishingType=draft` — so softprops is the effective uploader and the collision is there.)

## Fix

Keep **x64 as the single Windows update channel**. In the `win-arm64-cross` job:
- run electron-builder with **`--publish never`** (so its arch-less `latest.yml` can't be uploaded in any draft-timing window), and
- **delete `dist-electron/latest.yml`** before the softprops upload.

x64 installers run on arm64 Windows via the OS x64 emulation layer, so **arm64 users still auto-update** (to the emulated x64 build); the **native arm64 installer stays on the release** for manual download. The arm64 `latest.yml` is the only thing dropped.

## Trade-off / alternatives
- Native-arm64 Windows users auto-update to the emulated-x64 build instead of a native arm64 one. Given native arm64 Windows is a small segment, x64 emulation works, and the current behavior is *broken for x64 users every other release*, this is a strict improvement. The "proper" alternative (a post-build job that merges both jobs' `latest.yml` into one multi-arch file) is noted in CLAUDE.md as a future option.
- The mac jobs have the same latent `latest-mac.yml` collision, but it's **not user-impacting**: unsigned macOS builds can't auto-install, so the updater only ever prompts a manual download regardless of which arch's metadata wins. Left as-is.

## Test plan
- ⚠️ **Not exercisable by PR CI** — `release.yml` only runs on `v*` tags. YAML validated locally (js-yaml).
- **Needs an on-Windows release test** (per the issue): cut a release with this change, then auto-update an **x64** install to it and confirm `latest.yml` references the x64 installer and Restart & Install completes without the shortcut error. (You can tag a pre-release off this branch to test before merging.)

Closes #586


hyiger:
@codex review

PR #694, opened 2026-06-13
ci(release): sign + notarize macOS builds (opt-in via secrets)

hyiger:
## Summary

Wires **Developer ID code signing + notarization** into the macOS release builds, so distributed `.dmg`s open without the Gatekeeper "can't be checked for malware" block — and the macOS auto-updater starts working (it can't auto-install unsigned builds today).

**Safe to merge before the secrets exist.** Signing is gated on the secrets: when they're absent the workflow sets `CSC_IDENTITY_AUTO_DISCOVERY=false` and builds an **unsigned** `.dmg` exactly as today. Notarization is opt-in per build via `--config.mac.notarize=true`, which the workflow appends only when the Apple key secrets are present. Windows/Linux jobs are untouched (the mac signing env never reaches them).

## Changes
- `assets/entitlements.mac.plist` — Hardened Runtime entitlements (V8 JIT + `disable-library-validation` so the signed app can load the `@pokusew/pcsclite` native module). In `assets/` because that's electron-builder's `buildResources` dir here (`/build` is gitignored).
- `electron-builder.yml` `mac:` — `hardenedRuntime: true`, `gatekeeperAssess: false`, `entitlements`/`entitlementsInherit`.
- `.github/workflows/release.yml` — a mac-only **Prepare** step exports `CSC_LINK`/`CSC_KEY_PASSWORD` + the App Store Connect API key (`APPLE_API_KEY` path / `_ID` / `_ISSUER`) from secrets; the mac arm64 build is split into its own bash step (the shared step also runs on Windows, where the notarize-flag shell logic wouldn't parse); both mac steps notarize only when enabled.

## To enable (repo secrets — Settings → Secrets and variables → Actions)
| Secret | Value |
|---|---|
| `MAC_CSC_LINK` | base64 of your **Developer ID Application** `.p12` |
| `MAC_CSC_KEY_PASSWORD` | the `.p12` export password |
| `APPLE_API_KEY` | base64 of the App Store Connect **`.p8`** key |
| `APPLE_API_KEY_ID` | the key's **Key ID** |
| `APPLE_API_ISSUER` | the **Issuer ID** |

Once those are set, the next `v*` tag produces signed + notarized macOS installers. Verify a build with `codesign -dvvv` and `spctl -a -vvv` (expect "source=Notarized Developer ID").

> ⚠️ Untestable in CI without the secrets + a tag build — the logic is reviewed and the YAML validated, but the first signed run should be watched.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,11 +100,52 @@ jobs:
           npm run electron:fixlinks
           npm run electron:compile
 
+      # macOS code signing + notarization. Exports the signing env (via
+      # $GITHUB_ENV) consumed by the mac packaging steps below. When the secrets
+      # aren't configured, it disables auto-discovery so the build is UNSIGNED —
+      # exactly today's behavior — instead of failing. Notarization is opt-in via
+      # FILAMENTDB_MAC_NOTARIZE, set here only when the Apple key secrets exist.
+      - name: Prepare macOS code signing & notarization
+        if: startsWith(matrix.platform, 'mac')
+        env:
+          MAC_CSC_LINK: ${{ secrets.MAC_CSC_LINK }}
+          MAC_CSC_KEY_PASSWORD: ${{ secrets.MAC_CSC_KEY_PASSWORD }}
+          APPLE_API_KEY_BASE64: ${{ secrets.APPLE_API_KEY }}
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

PR #698, opened 2026-06-13
fix(updater): enable correct-arch macOS auto-update (zip + multi-arch latest-mac.yml)

hyiger:
## Why

After signing/notarization landed (v1.39.1), the macOS update banner still only offered **View release** — no in-app Download/Install. Signing was necessary but not sufficient. Two gaps remained, both confirmed against the installed `electron-updater@6.8.3` source:

1. **No `.zip` artifact.** `MacUpdater` calls `findFile(files, "zip", ...)` and throws **`ZIP file not provided`** for a dmg-only release — it can't auto-update from a `.dmg`.
2. **`latest-mac.yml` arch collision.** Both macOS jobs emit `latest-mac.yml`; the later softprops upload wins, so every Mac was served one arch (the macOS analogue of the Windows `latest.yml` collision fixed in #586). `MacUpdater` *filters the yml's `files` by the running arch* (keeps `arm64`-in-URL files on Apple Silicon / Rosetta, the rest on Intel), so a single yml listing **both** arches lets it pick the native build.

The `UpdateBanner` `isMac` gate existed precisely because of #1/#2 and routed Mac users to the release page.

## Changes

- **`electron-builder.yml`** — `mac.target: [dmg, zip]`. The global `artifactName` embeds `${arch}`, so zips land as `…-mac-arm64.zip` / `…-mac-x64.zip` (the `arm64` substring is what the updater filters on).
- **`release.yml`** — each mac job stashes its single-arch `latest-mac.yml` as an artifact and removes it from the upload set; a new **`merge-mac-metadata`** job merges them into one multi-arch `latest-mac.yml` and uploads it.
- **`scripts/merge-mac-latest-yml.mjs`** — pure merge helper (uses the `yaml` dep the app already parses OpenPrintTag YAML with; throws if either arch is missing so we never ship a yml that strands one platform). Unit-tested in **`tests/mergeMacLatestYml.test.ts`** (7 cases).
- **`UpdateBanner.tsx`** — drop the `isMac` gate; Mac gets the Download button like every platform. **View release** stays as a manual fallback.
- **`auto-updater.ts`** — refresh the stale "macOS can't auto-install" docblock.

## Verified locally

`tsc` ✓ · `eslint` ✓ · `mergeMacLatestYml` (7) + `auto-updater` (3) tests ✓ · both YAML files parse · `mac.target=[dmg,zip]` · `merge-mac-metadata` job present.

## ⚠️ Validation caveat (CI-only path)

The merge job + zip target can't be exercised without cutting a release. **It takes effect on the NEXT release**: that build ships the zip + multi-arch `latest-mac.yml`. Recommended check after the next tag: download `latest-mac.yml` from the release and confirm its `files` array lists both `-mac-arm64.zip` and `-mac-x64.zip`. The v1.40.0 → next transition still needs **one manual install** (the running app predates this change); auto-update is native-arch thereafter.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -209,6 +209,28 @@ jobs:
         shell: bash
         run: rm -f dist-electron/*.blockmap
 
+      # macOS auto-update needs ONE multi-arch latest-mac.yml so electron-updater
+      # serves the native arch. Each mac job emits its own single-arch
+      # latest-mac.yml; uploading both would let the later one win and serve one
+      # arch to every Mac (the macOS analogue of the Windows #586 collision
+      # below). Stash each arch's yml as an artifact and drop it from the upload
+      # set; the `merge-mac-metadata` job combines them into the multi-arch file.
+      - name: Stash mac update metadata for the multi-arch merge
+        if: startsWith(matrix.platform, 'mac')
+        shell: bash
+        run: |
+          mkdir -p mac-update-yml
+          cp dist-electron/latest-mac.yml "mac-update-yml/latest-mac-${{ matrix.arch }}.yml"
+          rm -f dist-electron/latest-mac.yml
+
+      - name: Upload mac update metadata artifact
+        if: startsWith(matrix.platform, 'mac')
+        uses: actions/upload-artifact@v4
+        with:
+          name: mac-update-yml-${{ matrix.arch }}
+          path: mac-update-yml/latest-mac-${{ matrix.arch }}.yml
+          retention-days: 1
+
       # GH #586: Windows update metadata is named `latest.yml` with NO arch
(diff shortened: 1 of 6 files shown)


hyiger:
@codex review

hyiger:
Addressed in 621d817d. Both macOS `electron-builder` commands now pass `--publish never` (mirroring the Windows arm64 job from #586), so electron-builder's `publish: github` config no longer uploads each job's single-arch `latest-mac.yml` on the tag build. Installers still upload via softprops; `merge-mac-metadata` is now the sole writer of `latest-mac.yml`. This also makes the failure mode safer — a skipped merge leaves *no* `latest-mac.yml` (the update check simply no-ops) rather than a wrong-arch one.

hyiger:
@codex review

PR #714, opened 2026-06-14
fix(web): dark-mode <select> popups rendering light-on-white (#711)

hyiger:
## Summary
Fixes #711 — in dark mode the native option popup of several `<select>` controls rendered **light-on-white** (unreadable), reported on the Spool Inventory **Kind** filter.

## Root cause
`globals.css` sets `color-scheme: dark` on `html.dark` (GH #598), which makes OS-drawn popups follow dark mode. **But** a `<select>` with an author-set *transparent* background (`bg-transparent`, no `dark:bg-*`) makes Chromium/Electron draw the option popup with light defaults *regardless* of `color-scheme`. The inherited light text then lands on a white popup → the screenshot in the issue.

The home-page filter selects never had this bug because they carry an explicit `bg-white dark:bg-gray-900 text-gray-900 dark:text-gray-100`.

## Fix
Apply that proven pattern to every affected select:
- **4 inline selects** updated directly: inventory Kind filter + per-row move-to; spool location + assign-slot on the filament detail page.
- **5 shared-`inputClass` forms** (Nozzle, Filament `spoolType`, Printer ×2, Location, BedType) get a derived `selectClass = inputClass.replace("bg-transparent", "bg-white dark:bg-gray-900")`, so **text/number inputs keep their transparent fill** (they have no popup) — only `<select>` changes.

## Scope / verification
- Audited **all 20 `<select>` elements** app-wide — the other 11 already had explicit dark backgrounds; **none missed**.
- Audited date/color/datalist controls (also `bg-transparent`): their OS pickers follow `color-scheme`, **not** the host background, so they're not the #711 failure mode and need no change.
- `tsc` clean, `eslint` clean. No test asserts on these classNames. Pure presentational change.

### DIFF
--- a/src/app/bed-types/BedTypeForm.tsx
+++ b/src/app/bed-types/BedTypeForm.tsx
@@ -82,6 +82,10 @@ export default function BedTypeForm({ initialData, onSubmit, onDirtyChange }: Pr
 
   const inputClass =
     "w-full px-3 py-2 border border-gray-300 dark:border-gray-600 rounded text-sm dark:text-gray-100 bg-transparent";
+  // GH #711: a <select> needs an explicit dark background so its native option
+  // popup follows dark mode — a transparent bg makes the popup render
+  // light-on-white. Text inputs keep bg-transparent (they have no popup).
+  const selectClass = inputClass.replace("bg-transparent", "bg-white dark:bg-gray-900");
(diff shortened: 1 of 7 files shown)


hyiger:
@codex review

PR #756, opened 2026-06-17
feat(openprinttag): create/link variants from OpenPrintTag (#753)

hyiger:
## What & why

Closes the gap from [#753](https://github.com/hyiger/filament-db/issues/753): you couldn't create a variant *from* OpenPrintTag (or link an existing filament to OPT) without the manual "import standalone → set parent → delete every inherited field by hand" dance. The OPT import path and the variant/inheritance model were built independently and never wired together.

Two of the issue's three proposed approaches, per the triage decision (**C + A**, strict-equality auto-prune; **B** skipped as the only data-mutation-risky option):

### C — Link an existing filament to OPT
- **`POST /api/filaments/{id}/openprinttag/link`** (CSRF-guarded): writes *only* `settings.openprinttag_slug`/`_uuid` + the `openprinttagSnapshot` provenance via the shared `buildOptLinkUpdate()`. **No field value is touched**, so a variant's inherited values are never clobbered — the issue's explicit "if a variant, don't overwrite inherited identical values" requirement falls out for free (the check route already diffs the variant's *effective* values).
- Detail page: **"Link to OpenPrintTag"** button (shown when `!_hasOwnOptLink`, the inverse of the re-sync gate) → new `OptLinkDialog` material picker → link → refetch. The teal "Check for updates" button then appears.

### A — Import an OPT material as a variant
- **`POST /api/openprinttag/import`** gains an optional `parentId`: imports exactly one slug as a variant of the parent, pruning every OPT-managed field **equal to the parent's effective value** (strict equality) so it inherits — pure, tested `pruneOptPayloadAgainstParent()`. The snapshot is built from the **full** offer *before* pruning (so a pruned-then-edited field still classifies correctly). Refuses a name collision with **409** rather than re-parenting an existing row. `vendor`/`type` (required) and `color`/`name` (variant-only) are never pruned.
- OPT browser: when exactly one material is selected, an **"Import as variant of…"** parent picker (lazily-fetched root filaments) drives the variant import.

## Notes
- DRY: the existing bulk import route now uses `buildOptLinkUpdate` for its link write.
- OpenAPI documents the new endpoint + the `parentId` import param. EN/DE i18n in lockstep (parity invariant passes).

## Testing
- `npm test` — **2247 pass** (full suite). `npm run lint` clean, `tsc --noEmit` clean.
- New: `pruneOptPayloadAgainstParent` + `buildOptLinkUpdate` (pure helpers) and link-route + variant-import route cases — variant-aware classification, strict-equality prune, collision / parent-validation / malformed-id guards, CSRF.

Fixes #753


hyiger:
@codex review

hyiger:
Addressed the P2 in `db32b0b1` → `93e091d2`.

**Empty variant arrays inheriting parent data** — this is a genuine consequence of the intentional empty=inherit inheritance model (`resolveFilament`, GH #106/#477): a variant has no representation for "explicitly empty, don't inherit" arrays, so `[]` always resolves to the parent's array when the parent is non-empty. It's the **same** limitation already documented on the resync side in this exact file (`OPT_CLEARABLE_ARRAY_FIELDS`). No stored value can make the variant resolve to empty while the parent's array is non-empty, so it can't be fixed in the prune without a model-level "no-inherit" array sentinel — which would ripple through `resolveFilament`, the resync diff, and exports (out of scope for this PR).

The prune correctly does **not** fabricate the parent's data onto the variant, and the user can override after import by editing the variant. I've documented the limitation explicitly in `pruneOptPayloadAgainstParent` and pinned it with a test. Common case (parent is a base catalog entry without secondaries/tags) is unaffected.

@codex review

PR #760, opened 2026-06-18
fix(label-printer): commit the Windows spool job so it doesn't reprint (#759)

hyiger:
Fixes [#759](https://github.com/hyiger/filament-db/issues/759) — on Windows (USB Brother PT-P710BT) a label prints correctly but the spooler job never reports "completed", so it re-spools on reboot/re-plug (printing twice) and blocks the next print (2nd sticks at "unknown status", 3rd errors). macOS/Linux (CUPS) are unaffected.

## Root cause
The winspool RAW path (`electron/label-printer.ts`) drove the Win32 sequence in the correct order, but **discarded the bool returns of `EndPagePrinter`/`EndDocPrinter`** — the two calls that *commit* the spool job as printed. The bytes reach the head (so the label prints), but if `EndDocPrinter` returns false the C# still returns normally → PowerShell exits 0 → the renderer sees `{ ok: true }` while the spooler holds an uncommitted job. That uncommitted job re-spools on reboot and keeps the port busy.

A too-aggressive 15s subprocess timeout (shared with *listing*) can also `SIGKILL` PowerShell mid-`EndDocPrinter` once the spooler is slow/busy — the amplifier behind "2nd sticks, 3rd errors".

## Fix (Windows branch only — CUPS path untouched)
- **Check `EndPagePrinter`/`EndDocPrinter` returns and throw on failure** — a job that can't be committed now surfaces a real error instead of a false success. `pageOk`/`docOk` flags guard the throw so a teardown after a genuine `WritePrinter` failure doesn't mask the original error; the End*/Close calls still **always** run in their finallys, tearing a failed job down instead of leaking it open.
- **Dedicated `WINDOWS_PRINT_TIMEOUT_MS` (25s, under the 30s IPC wrapper)** instead of the 15s listing timeout, so a slow USB-label spooler isn't killed mid-commit.

## Testing
- `electron/` is out of the Vitest/tsconfig scope and the win32 path can't run without a Windows host + printer, so `tests/label-printer.test.ts` gains **string-level guards** on the generated PowerShell (End* failure checks present, RAW datatype, the timeout bounds) to pin the fix against regression. Full `npm test`/`lint` green.
- ⚠️ **On-hardware confirmation is OWED** (this is a blind patch, as discussed on the issue): verify on Windows that the job completes in the queue, doesn't reprint on reboot, and that 2nd/3rd prints succeed.

## Fallback (if needed)
If hardware testing shows `EndDocPrinter` returns *true* yet the job still hangs, the deeper fix is printing through the installed Brother driver queue (OS owns lifecycle, like CUPS) rather than raw winspool — a separate, larger change to decide after testing.

### DIFF
--- a/electron/label-printer.ts
+++ b/electron/label-printer.ts
@@ -38,10 +38,19 @@ const PRINTER_PATTERN = /pt-?p710bt|p-?touch|brother/i;
  *  letters, digits and underscores. */
 const MANAGED_QUEUE = "FilamentDB_Label";
 
-/** Per-subprocess timeout. Listing and a single 24mm label both complete
- *  well under this; the IPC handler wraps the whole print in its own 30s
- *  timeout on top. */
+/** Per-subprocess timeout for listing + the CUPS `lp` print. Listing and a
+ *  single 24mm label both complete well under this; the IPC handler wraps the
+ *  whole print in its own 30s timeout on top. */
 const EXEC_TIMEOUT_MS = 15_000;
+
+/** GH #759 — the Windows winspool print gets a LONGER subprocess timeout than
+ *  the 15s listing timeout. `EndDocPrinter` blocks until the spooler drains the
+ *  RAW bytes to the (slow, USB) label printer, which can take several seconds;
+ *  at 15s a busy/recovering spooler could be SIGKILLed mid-`EndDocPrinter`,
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #782, opened 2026-06-19
fix(ui): settings order, compare cost row, analytics copy (#778, #779, #781)

hyiger:
Three independent QA-pass fixes from the v1.53.0 hands-on review, grouped (all small, no overlap).

### #778 — Danger Zone rendered mid-page
"Delete All Data" sat before the Label printer / Label format sections. Moved the Danger Zone block to be the **last** section (after the label sections, before the version footer) so the irreversible action is at the bottom. Pure JSX reorder; `LabelFormatEditor` renders on web too, so Danger Zone stays last in both web + Electron.

### #779 — Compare redundant cost rows
`cost` is stored **per-kg**, so "Cost" and "Cost / kg" always showed the same number. Dropped the plain "Cost" row, kept a single per-kg row, and removed its spurious `density == null` guard (per-kg needs no conversion). Removed the now-orphaned `compare.row.cost` key from en + de.

### #781 — Analytics empty-state jargon
Softened `analytics.noData` (en + de) so it no longer surfaces the raw `POST /api/print-history` dev path — now "…have your slicer integration log a print job." API detail stays in the docs.

`tsc`, `lint`, `tests/i18n-key-coverage.test.ts`, and `tests/compare-route.test.ts` all green.

Fixes #778
Fixes #779
Fixes #781


hyiger:
@codex review

hyiger:
@codex review (HEAD is the main-merge c4f29896cb — content unchanged from the reviewed commit)

PR #797, opened 2026-06-23
fix(docker): install tzdata so the TZ env var works (#796)

hyiger:
Fixes [#796](https://github.com/hyiger/filament-db/issues/796).

The `node:22-alpine` runtime stage shipped without a zoneinfo database, so `TZ=America/Los_Angeles` (or any zone) couldn't be resolved — musl **silently falls back to UTC**, and the documented env var had no effect.

- **Dockerfile**: `RUN apk add --no-cache tzdata` in the runner stage, as root, before the `USER nextjs` switch. The container stays UTC unless `TZ` is set; ~3MB image bump.
- **docs/setup.md**: added a `TZ` row to the environment-variable table.

**Verification** (no Docker build in CI): `docker run --rm -e TZ=America/Los_Angeles  date` now shows local time; `docker run --rm  date` (no TZ) stays UTC.

Fixes #796

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,12 @@ ENV NODE_ENV=production
 ENV PORT=3000
 ENV HOSTNAME=0.0.0.0
 
+# GH #796: the node:22-alpine base ships no zoneinfo DB, so a `TZ=America/...`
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #812, opened 2026-06-23
test: retry MongoMemoryServer startup on random-port collision (#808)

hyiger:
Fixes [#808](https://github.com/hyiger/filament-db/issues/808).

A full `npm test` run could fail **before any suite executed** when `mongodb-memory-server` picked a random port already held by another local process (the report saw `steam_osx` on `:57343`), throwing `StdoutInstanceError: Port "57343" already in use`. The target suite passed on its own when rerun — a classic flaky, unrelated red build.

### Fix
`tests/setup.ts`'s shared `beforeAll` now starts Mongo through a new [`startWithRetry`](tests/mongoRetry.ts) helper:
- up to **5 attempts**, a **fresh `MongoMemoryServer`** (new random port) per attempt;
- retries **only** on the port-in-use / `EADDRINUSE` error — genuine startup failures (missing binary, launch timeout) rethrow immediately so they stay visible;
- `console.warn` on each retry so CI shows the flake happened;
- port collisions fail fast, so the retries fit comfortably inside the existing 120s start budget.

The helper is dependency-injected (`create` + `sleep`), so it's unit-tested without spawning real mongod — `tests/mongoRetry.test.ts` covers retry-then-succeed, immediate rethrow of a non-port error, give-up after `maxAttempts`, and the `isPortInUseError` classifier.

`tsc` + `lint` green; the retry tests pass and DB-backed suites still start correctly through the new wiring.

Fixes #808

### DIFF
--- a/tests/mongoRetry.test.ts
+++ b/tests/mongoRetry.test.ts
@@ -1,0 +1,69 @@
+import { describe, it, expect, vi } from "vitest";
+import { isPortInUseError, startWithRetry } from "./mongoRetry";
+
+/**
+ * GH #808 — the shared Mongo startup must survive a random-port collision by
+ * retrying with a fresh server, but must not mask genuine startup failures.
+ */
+describe("isPortInUseError", () => {
+  it("matches the mongodb-memory-server port-in-use message", () => {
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

PR #836, opened 2026-06-24
fix(mobile): throttle QR re-scan + gate NFC button on device availability

hyiger:
From the audit triage. Two mobile-scanner fixes in `packages/mobile`.

## #815 — QR scanner request storm (P2)
`scan-qr.tsx` re-armed its once-only `handled` guard **synchronously** on both failure paths, while `expo-camera` fires `onBarcodeScanned` on every frame a code is visible. Holding an unmatched QR in frame (or scanning while the server is unreachable) produced a continuous stream of `GET /api/filaments/match` requests — one per round-trip — plus a flickering error overlay.

**Fix:** re-arm on a **1.5 s cooldown** instead of synchronously, with same-code dedupe (`lastScan` ref) so a held code is tried once. Clearing `lastScan` in the cooldown lets the *same* code retry after the window (e.g. once the server is back). `useEffect` cleanup clears a pending timer on unmount.

## #824 — NFC button not gated on device availability (P3)
The "Scan NFC tag" button was gated only on the build-time `NFC_ENABLED` flag; `isNfcAvailable()` was unused dead code. On a device with no NFC (e.g. iPad) or NFC turned off, tapping surfaced a raw native error.

**Fix:** probe real device NFC availability on mount (lazy-init `false` when `NFC_ENABLED` is off, to satisfy `react-hooks/set-state-in-effect`), disable the button + show an "NFC unavailable on this device" hint when absent/off, and guard `scanNfc()` as defense-in-depth.

## Verification
- `packages/mobile`: `tsc --noEmit` clean, `eslint` clean on both files.
- No automated harness exists for the mobile screens (`packages/` is excluded from root Vitest/tsc), so behavior was verified by reading the camera/handler flow. On-device manual checks (request storm throttled, NFC button disabled without hardware) are owed.

Fixes #815
Fixes #824

### DIFF
--- a/packages/mobile/src/app/index.tsx
+++ b/packages/mobile/src/app/index.tsx
@@ -1,11 +1,11 @@
 import { Link, useRouter } from 'expo-router';
-import { useState, type ReactNode } from 'react';
+import { useEffect, useState, type ReactNode } from 'react';
 import { ActivityIndicator, Alert, Pressable, StyleSheet, Text, View } from 'react-native';
 import { SafeAreaView } from 'react-native-safe-area-context';
 
 import { ApiError, createApi } from '@/lib/api';
 import { NFC_ENABLED } from '@/lib/features';
-import { readOpenPrintTag } from '@/lib/nfc';
+import { isNfcAvailable, readOpenPrintTag } from '@/lib/nfc';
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

hyiger:
@codex review

PR #848, opened 2026-06-24
fix(list): show all filaments under "All" when a catalog has no spools (#847)

hyiger:
From the triage of #847 (bug 2 of 2 — the filter behaviour).

## Problem
A catalog with filaments but **0 spools** rendered "No filaments match the current filters" under the selected **`All (120)`** chip. The #712 "hide out-of-stock by default" feature hides filaments with no active spool — with 0 spools *nothing* is in stock, so it hid all 120 — while the chip count is computed **pre-hide**, giving the contradiction "All (120)" with zero rows. The same 120 appeared under "Show out of stock (120)" / "No calibration (120)".

## Fix
- When the unfiltered `All` view would hide **everything** (the in-stock list is empty), fall back to showing all filaments (`src/app/page.tsx`). The #712 declutter still applies whenever ≥1 filament is stocked, so this only ever shows **more** rows — it can't regress the stocked case.
- Suppress the now-no-op "Show out of stock (N)" toggle when nothing is in stock (the items are already shown), gating it on a new `hasAnyInStock`.

The triage's secondary "toggle-aware empty state" message is **intentionally not added**: this fix makes `All` non-empty whenever filaments exist, so that branch is unreachable for `All`, and showing an out-of-stock hint on a genuine search/filter no-match would mislead.

## Verification
- `tsc --noEmit` + eslint clean.
- Reproduction is data-dependent (filaments + 0 spools); the change is a small, self-contained conditional (show all when the in-stock subset is empty). The live preview shares the production DB, so I didn't reproduce the 0-spool state there.

Part of **#847** — the false-tag-detection half is a separate PR; #847 stays open until both land.

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -494,6 +494,12 @@ export default function Home() {
     }).length;
   }, [filaments, inStock]);
 
+  // #847: when NOTHING is in stock (e.g. a catalog with 0 spools), the "all"
+  // view falls back to showing every filament, so the "Show out of stock"
+  // toggle would be a no-op — suppress it unless at least one filament is
+  // actually in stock (i.e. the hide is genuinely hiding something).
+  const hasAnyInStock = useMemo(() => filaments.some(inStock), [filaments, inStock]);
+
   const visibleFilaments = useMemo(() => {
     // The "all" view keeps parents in the dataset so the list renders them as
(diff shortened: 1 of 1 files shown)


hyiger:
@codex review

PR #849, opened 2026-06-24
fix(nfc): ignore virtual/contact PC/SC readers + surface reader name (#847)

hyiger:
From the triage of #847 (bug 1 of 2 — false tag detection).

## Problem
On Windows 11 with **no NFC hardware installed**, the app showed "Tag detected" and popped an unknown-filament dialog. The NFC service trusted **every** PC/SC reader the OS enumerates (`electron/nfc-service.ts:118`, no name/type filter), and Windows routinely enumerates *virtual* readers (Windows Hello, UICC/eSIM, TPM virtual smart cards) and built-in **contact** smart-card slots. A virtual/idle card's "present" bit (or a driver phantom) flipped `tagPresent` → the header pill + the present-edge auto-read → the popup.

## Fix (chosen option: denylist + surface reader name)
- **`src/lib/nfcReaderFilter.ts`** (new, pure, unit-tested) — `isLikelyContactlessReader()` rejects virtual/system/contact reader names (Windows Hello / UICC / TPM / VSC / Microsoft virtual / `Contacted`) and lets everything else through (denylist, not allowlist, so an unlisted real reader still works).
- **`electron/nfc-service.ts`** — skips a non-contactless reader at enumeration (logs the name); it never sets `readerConnected` or fires an auto-read.
- **`NfcStatus.tsx`** — surfaces the connected reader's **name** in the tooltip (the reporter's explicit ask + the diagnostic to extend the denylist).

## Scope note (decode-gate deliberately deferred)
The chosen option also listed a "decode-gate" (don't show "Tag detected"/popup until a successful decode). The code analysis showed it's **redundant after filtering** (the reader is dropped at enumeration, so nothing downstream fires) and a decode-gate on the deliberate present-edge read would **contradict the #572 design** (a deliberate placement is *meant* to surface errors) and regress real-reader feedback — which I can't verify without Windows NFC hardware. If an unlisted reader still trips it, the now-visible reader name pins exactly which pattern to add. Happy to add a decode-gate if you'd rather.

## Verification
- `tsc` + **electron typecheck** (0 errors) + eslint clean.
- `tests/nfcReaderFilter.test.ts` (5 cases: accepts ACS/Sony/Identiv/Omnikey, rejects Windows virtual + `Contacted`, rejects empty); i18n parity 1593/1593; coverage test passes.
- **On-hardware (Windows) confirmation owed** — no Windows NFC host in CI.

Part of **#847** — the All-filter half is #848; #847 stays open until both land.

### DIFF
--- a/electron/nfc-service.ts
+++ b/electron/nfc-service.ts
@@ -20,6 +20,7 @@ import {
   classifyNfcError as classifyNfcErrorImpl,
   type NfcErrorCode,
 } from "../src/lib/nfcErrorClassify";
+import { isLikelyContactlessReader } from "../src/lib/nfcReaderFilter";
 
 export type { NfcErrorCode };
 /** Re-export the classifier so the existing call-sites in this file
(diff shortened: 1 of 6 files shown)


hyiger:
@codex review

PR #855, opened 2026-06-25
fix(api): 400 (not 500) on a malformed id for the OpenPrintTag binary route (#854)

hyiger:
Follow-up to #818/#839 — that PR added a `mongoose.isValidObjectId` 400 guard to the OpenPrintTag `check`/`sync`/`link` routes but **missed the binary download route** `GET /api/filaments/{id}/openprinttag`.

A non-ObjectId `id` (e.g. `/api/filaments/not-an-object-id/openprinttag`) made Mongoose throw a `CastError` before the 404 branch, which the generic catch mapped to **500** with the raw driver detail in `detail`.

**Fix:** add the identical guard after `const { id } = await params;` — returns **400 "Invalid filament id"** before the query, byte-identical to the sibling routes. Plus a route test asserting 400 (not 500) for a malformed id.

## Verification
- `tsc` + eslint clean; `tests/openprinttag-route.test.ts` → 7 passing (added the malformed-id case).

Fixes #854

### DIFF
--- a/src/app/api/filaments/[id]/openprinttag/route.ts
+++ b/src/app/api/filaments/[id]/openprinttag/route.ts
@@ -1,4 +1,5 @@
 import { NextRequest, NextResponse } from "next/server";
+import mongoose from "mongoose";
 import dbConnect from "@/lib/mongodb";
 import Filament from "@/models/Filament";
 import "@/models/Nozzle";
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #856, opened 2026-06-25
ci: full gate before release/docker publish + pin reference content (#852, #853)

hyiger:
From the triage of the new review-filed issues. Two release-pipeline hardening fixes.

## #852 (P1) — tag releases skipped the full CI gate
`test.yml` (push-to-main/PRs only) runs the full gate — lint, root `tsc`, **`electron:typecheck`**, `test:coverage`, `npm audit --audit-level=moderate`, build, standalone smoke — but the tag-triggered `release.yml`/`docker.yml` re-implemented only **lint + `npm test`**. So a `v*` tag could publish installers/images from a commit branch CI would reject. Sharpest gap: `electron/` is excluded from the root tsconfig and `electron:compile` is esbuild (strips types), so ~5.6k lines of Electron code went **type-unchecked into installers**.

**Fix:** new reusable **`ci-gate.yml`** (`workflow_call`) with the full gate. `release.yml` adds a `gate` job that runs **once** before the platform matrix (build legs `needs: gate`); `docker.yml`'s build `needs: gate`. Dropped release.yml's redundant per-leg lint/test.

**Why `test.yml` stays inline (chosen with the maintainer):** a reusable call renames its checks to `gate / test (20)` etc., which would break `main`'s branch protection (required contexts `test (20)`/`test (22)`/`smoke`) and stall every open PR. So `test.yml` keeps its inline jobs and `ci-gate.yml` **mirrors** them — two copies with cross-referenced `KEEP IN SYNC` comments.

## #853 (P2) — per-platform reference-content drift
`release.yml` ran `fetch-reference.mjs` (best-effort) **inside each of the 6 platform legs**, so one tag could bundle different reference content + `REFERENCE_WIKI_SHA` per platform.

**Fix:** remove the per-build fetch — every artifact (desktop + Docker, which already used the committed snapshot) now ships the committed `src/content/` **identical by construction**. New scheduled **`reference-drift.yml`** (weekly + `workflow_dispatch`) regenerates from the wiki and **fails on drift** as a visible nudge to refresh via a PR (not a PR gate). CLAUDE.md documents the pre-tag `npm run fetch:reference` ritual.

## Validation
- **All 5 workflows pass `actionlint` 1.7.12** (it validates `workflow_call` + `needs:` wiring).
- `test.yml`'s diff is **only a comment** — the `test`/`smoke` jobs are byte-identical, so the required check names are untouched.
- **Inherent limitation (this is the #852 bug itself):** the tag-path wiring (`release.yml`/`docker.yml` `needs: gate`) can't be exercised by PR CI — the first real run is the **next `v*` tag**. The reusable `ci-gate.yml` content is proven by being a verbatim mirror of the test.yml jobs that pass on this PR.

Fixes #852
Fixes #853


hyiger:
@codex review

PR #860, opened 2026-06-25
fix(api): PrusaSlicer sync no longer 400s on a legacy out-of-range stored temp (#859)

hyiger:
From the triage of #859. **Most likely** root cause, empirically reproduced (mongoose + mongodb-memory-server).

## Root cause (a regression)
#645 (commit `80de848d`) added `runValidators: true` to the slicer per-id sync write — **but** the handler re-`$set`s the **whole `temperatures` object** (`{ ...existing, ...incoming }`, `route.ts:463-467`), and PrusaSlicer always sends temp keys. So the filament's **stored** temps ride into the validated `$set`, and a single **legacy out-of-range value** (e.g. a `temperatures.nozzle` > 600 saved before the validators existed) fails validation → **400 → nothing writes**, silently dropping EM *and* max-volumetric-flow. Before #645 the write had no validators, so legacy data was tolerated — hence *"no longer syncs back."*

## Fix
Write only the temperature keys PrusaSlicer actually sent, as **dotted paths** (`temperatures.bed`, …), instead of `$set`-ing the whole object. Update-validators then check **only the incoming values** — a bad incoming value is still rejected (preserves #645's intent), but a legacy value the sync isn't touching is left alone and unvalidated.

## Verification
- New regression test: a legacy-out-of-range doc (seeded raw to bypass validators) now syncs EM + `maxVolumetricSpeed` (**200**, was **400** on the old route — confirmed by reverting); a genuinely-bad *incoming* temperature still **400s**.
- Full `api-route-correctness.test.ts` → **28 passing** (incl. the existing #265 sync tests). `tsc` + eslint clean.

## Why "Addresses" not "Fixes"
This is the most likely cause but **medium confidence for this user's exact case** — the decisive evidence (the sync POST request/response) isn't in the issue. I'm asking the reporter to capture it; a **403** would instead point at the CSRF guard (a separate, security-sensitive change), a **404** at a renamed preset (client-side). The fix here is a genuine robustness bug worth landing regardless. **#859 stays open until confirmed.**

Addresses #859

### DIFF
--- a/src/app/api/filaments/[id]/route.ts
+++ b/src/app/api/filaments/[id]/route.ts
@@ -460,10 +460,18 @@ export async function POST(
     if (config.filament_soluble) update.soluble = config.filament_soluble === "1";
     if (config.filament_abrasive) update.abrasive = config.filament_abrasive === "1";
 
-    // Merge temperatures into existing
-    if (Object.keys(temps).length > 0) {
-      const existing = (filament.temperatures as Record<string, unknown>) || {};
-      update.temperatures = { ...existing, ...temps };
+    // #859: write ONLY the temperature keys PrusaSlicer actually sent, as
+    // dotted paths — never a $set of the whole `temperatures` object. #645 added
+    // `runValidators` to the sync write below; Mongoose update-validators check
+    // every path in the $set, so the previous `{ ...existing, ...temps }` merge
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #862, opened 2026-06-25
fix(api): PrusaSlicer EM/PA sync falls back to the global nozzle catalog (#859)

hyiger:
Found while live-debugging a PrusaSlicer 2.9.5 (Filament-Edition 1.9.2) ↔ Filament DB 1.57.3 sync.

## Root cause
The sync endpoint is **healthy** — replaying PrusaSlicer's exact request (name-based `POST /api/filaments/{name}?nozzle_diameter=0.4&high_flow=1`, no `Origin`) returns **200** and the top-level `maxVolumetricSpeed` persists. But the **extrusion multiplier never landed**: EM/pressure-advance write to a **per-nozzle calibration**, and the matching only searched the filament's `compatibleNozzles` with **no fallback** when that list is empty (`if (compatIds.length > 0) { … }`, no `else`).

A filament synced/imported from Filament DB typically has `compatibleNozzles: []`, so the block was skipped and **EM was silently dropped** — even though a unique matching nozzle (e.g. a `0.4mm HF`) was sitting in the global catalog. The **Bambu and OrcaSlicer** routes already fall back to a unique catalog nozzle; the **PrusaSlicer** route didn't.

## Fix
When no compatible nozzle matches, fall back to the **global nozzle catalog** and attach the calibration to the **unique** nozzle at this diameter (+ `high_flow`) — only when exactly one matches, so it never guesses. Mirrors the Bambu/Orca logic.

## Verification
- New tests: a no-compatible-nozzle filament now lands EM on the unique `0.4mm-HF` catalog nozzle; an ambiguous case (two matching catalog nozzles) writes no calibration. `tests/api-route-correctness.test.ts` → **30 passing**. `tsc` + eslint clean.
- Confirmed live against the running 1.57.3 server that the *endpoint* returns 200 for the real request shape; the running app predates this fix, so the test suite is the validation.

Addresses **#859** (the EM-sync half — kept open until verified end-to-end).

### DIFF
--- a/src/app/api/filaments/[id]/route.ts
+++ b/src/app/api/filaments/[id]/route.ts
@@ -578,6 +578,33 @@ export async function POST(
             };
           }
         }
+
+        // #859 follow-up: when the filament has no matching COMPATIBLE nozzle
+        // (commonly because `compatibleNozzles` is empty on a filament imported
+        // / synced from Filament DB), fall back to the GLOBAL nozzle catalog and
+        // attach the calibration to the UNIQUE nozzle at this diameter (+
+        // high_flow). Mirrors the Bambu/OrcaSlicer sync routes; only when
+        // EXACTLY ONE catalog nozzle matches, so we never guess. Without this,
+        // PrusaSlicer's EM / pressure-advance edits were silently dropped for any
+        // filament with no compatible nozzles set.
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #874, opened 2026-06-26
feat(ui): distinguish same-Ø nozzles + hardened gate for abrasive (#872 item 4)

hyiger:
Part of #872 (item 4 of 6).

## Problem
You defined nozzles like '0.4mm' (Brass) and '0.4mm' (Diamondback) — **named by diameter** — so the compatible-nozzle picker (which showed only `nozzle.name`) rendered two identical '0.4mm' options you couldn't tell apart. (Confirmed against your data: two 0.4mm nozzles, one Brass/soft, one Diamondback/hardened.)

## Fix
- **Always show `· <Ø>mm <Type>`** after the name, so same-diameter nozzles are distinguishable regardless of how they're named (e.g. '0.4mm · 0.4mm Brass' vs '0.4mm · 0.4mm Diamondback').
- **Hardened gate**: when the filament's `abrasive` tag is set, the picker hard-filters to **hardened** nozzles. *Select all* respects it, and a note reports how many were hidden. An already-selected soft nozzle stays visible with a 'not hardened' flag so it can still be removed.
- Added `hardened` to the form's `NozzleOption` type (the `/api/nozzles` response already includes it).

## Verified (preview)
- The two '0.4mm' nozzles now read 'Brass' vs 'Diamondback'.
- Toggling **Abrasive** drops the picker 8 → 6 nozzles with the note *'Abrasive filament — showing hardened nozzles only (2 hidden).'*
- tsc + lint + i18n-parity green.

i18n: `form.nozzle.hardenedOnly` + `form.nozzle.notHardened` (en + de).

### DIFF
--- a/src/app/filaments/FilamentForm.tsx
+++ b/src/app/filaments/FilamentForm.tsx
@@ -620,14 +620,19 @@ export default function FilamentForm({ initialData, onSubmit, onDirtyChange }: P
   }, [printers, nozzles, form.compatibleNozzles]);
 
   // #872: an abrasive filament needs a hardened nozzle, so the compatible-nozzle
-  // picker hard-filters to hardened ones when `abrasive` is set. Already-selected
-  // nozzles stay visible (even if soft) so the user can still deselect them —
-  // they're flagged in the list; only NEW soft nozzles are hidden.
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

hyiger:
@codex review

PR #876, opened 2026-06-27
feat(prusaslicer): multi-nozzle filaments → one preset per nozzle, with sync-back routing (#872 item 5)

hyiger:
Part of #872 (item 5, **export side**).

## What
PrusaSlicer has no parent/child for *user* filament presets, so a filament with **≥2 distinct nozzle calibrations** now exports as **N flat presets**, named `<name> <Ø> <type>` (e.g. `Inslogic PA12-CF 0.4 Diamondback`). Each preset:
- **bakes that nozzle's filament-level calibration** — extrusion_multiplier, filament_retract_*, filament_max_volumetric_speed, temps, fan;
- carries a **nozzle-scoped** `compatible_printers_condition = nozzle_diameter[0]==<Ø>` (dovetails with item 6);
- carries a `filamentdb_nozzle` hint for sync-back routing;
- shares the one `filamentdb_id` (#867).

**Single-nozzle filaments are unchanged** (one preset; calibration applied dynamically via the fork's `/calibration` endpoint).

## Known limitations (by design)
- **Pressure advance is printer-scoped** in PrusaSlicer, so it is NOT baked into the flat preset — it stays dynamic via `/calibration`.
- **Sync-back routing is item 5b (follow-up):** reading `filamentdb_nozzle` to route a synced preset's calibration values to the right per-nozzle entry (instead of top-level fields). Until 5b lands, syncing a per-nozzle preset writes EM/PA/retraction to the calibration entry via the existing `?nozzle_diameter=` path (the fork sends it), while temps/max-vol land on the top-level default — bounded, not data-loss.

## Tests
Multi-nozzle expansion (suffixed names, baked keys, scoped condition, nozzle hint) + single-nozzle stays-one-preset. Full bundle suite + tsc + lint green.

Per your 'proceed' on the recommended approach. Item 5b (sync-back) next.


hyiger:
@codex review

hyiger:
@codex review

hyiger:
Addressed both:

**P1 (sync-back / round-trip):** the exported per-nozzle presets carry the base `filamentdb_id`, so a name-addressed sync of `PLA 0.4 Brass` 409'd against the stored name `PLA` — not round-trippable. The sync route now reads the `filamentdb_nozzle` hint (added to `STRUCTURED_KEYS`), recognizes a per-nozzle preset (`<base> <Ø type [HF]>`), **suppresses the name mismatch** for it (a genuine mismatch with no hint still 409s), and **routes the calibration** to the matching per-nozzle entry — diameter from the hint, plus **type + high-flow to disambiguate same-Ø nozzles** (Brass vs Diamondback), across both the compatible-nozzle and global-catalog paths. Also suppresses the Phase 2 ObjectId-URL rename when a hint is present (an id-addressed per-nozzle sync can't overwrite the base name).

**P2 (high-flow grouping):** export groups by high-flow too (`0.4 Brass` std + `0.4 Brass HF` are distinct), suffix `<Ø> <type> HF`. Extracted a shared `nozzleSuffix()` so the export name, the hint, and the sync-back parse all agree.

This makes #876 the **complete item 5** (export + sync-back), not export-only. +4 sync-back route tests + high-flow/single/multi export tests.

@codex review

hyiger:
Addressed both:

**P1 (baked max-vol/temps/fan overwriting the filament-wide default):** a recognized per-nozzle sync now routes the full set — EM / PA / retraction **+ max-vol / temps / fan** — onto the *matching calibration entry* and leaves the top level untouched. Fan (which has no top-level home) is pulled out of the settings bag only when a calibration actually captured it. If no nozzle resolves, max-vol + temps fall back to the top level so nothing is lost.

**P2 (hinted ObjectId syncs not treated as per-nozzle):** `isPerNozzlePreset` now also covers `matchedByUrlObjectId`, so an id-addressed per-nozzle sync (ObjectId URL + `filamentdb_nozzle`, no `?nozzle_diameter=`) routes calibration via the hint's diameter/type/HF.

+3 route tests (routing-to-calibration with the top level untouched & fan not leaked; id-addressed routing; no-nozzle top-level fallback). 50 route + bundle tests green, tsc + lint clean.

@codex review

hyiger:
Addressed in 0b418452 — per-nozzle EM/pressure-advance/retraction are now excluded from the shared `settings` bag under the same `routeToCalibration && calibrationWrite` gate as the fan keys, so they land only on the resolved calibration entry. Single-nozzle syncs are unaffected. Strengthened the #872 per-nozzle test to assert EM/retraction don't leak into settings.

@codex review

hyiger:
Addressed the PA-routing P2 in `96f04c6a`, plus four related export/read/sync consistency gaps an adversarial self-review surfaced in the same area:

**The flagged P2** — PA stays dynamic (it's printer-scoped in PrusaSlicer, correctly not baked), so the `/calibration` read path needed the same type disambiguation as sync-back. Added an optional `nozzle_type` param to `GET /api/filaments/{id}/calibration` (soft filter: type match wins, else fall back to diameter matches so a type/data miss never 404s).

**Self-review follow-ups (same diff):**
- **Sync-back type match was case-SENSITIVE** while the new read path is case-insensitive — my own comment claimed symmetry that didn't hold. Made the sync-back nozzle-type query case-insensitive (anchored `escapeRegex`), so a lowercased/user-edited `0.4 diamondback` hint resolves the same nozzle on both paths.
- **EM/PA/retraction leak**: a per-nozzle preset that resolved *no* nozzle still dropped those keys into the shared `settings` bag (the exclusion was gated on `calibrationWrite`). Now gated on `routeToCalibration` alone — unresolved per-nozzle keys are dropped (no top-level home) rather than becoming the base filament's default.
- **Export grouping case-fold**: the `byNozzle` key used raw `nz.type`, so a casing variant ("Brass"/"brass") would split into two presets that both resolve to the same calibration. Case-folded the key (the human-readable suffix/hint keeps original casing).
- Reworded the read-path comment to stop overstating fallback symmetry (read soft-falls-back to a same-diameter entry; sync-back writes nothing per-nozzle on a type miss — intentional divergence).

**OrcaSlicer/Bambu (deliberately not flattened):** the per-nozzle flat-preset workaround is PrusaSlicer-specific because PrusaSlicer has no parent/child for *user* filament presets (issue #872). OrcaSlicer/Bambu support user-preset inheritance, so they keep one preset + dynamic `/calibration` — and the new `nozzle_type` param serves them too (the endpoint backs `format=orcaslicer`).

Tests: new `tests/calibration-route.test.ts` (type disambiguation both ways, case-insensitivity, unmatched-type fallback, high_flow, 404 diagnostics), extended #872 sync tests (case-insensitive routing + no-leak on the unresolved path), and a new export case-fold grouping test. `tsc` + lint clean; 89 tests across the three affected suites pass.

@codex review

hyiger:
Fixed in `f9bdcab7`. The bulk re-import no longer spawns suffixed orphan records — plus a self-review caught the same class of issue on two more surfaces, so all three `parseIniFilaments` callers are now consistent.

**The flagged P2** — new shared pure helper `collapsePerNozzleImportSections()` folds Filament DB's own per-nozzle suffixed sections (detected by the `filamentdb_nozzle` hint) back into their base filament: de-suffixes the name, groups siblings by `filamentdb_id`, strips routing hints + per-nozzle baked keys, and omits temps/max-vol so an update can't clobber the base's shared values.

**Wired into all three callers** (the other two had the same latent bug):
- `POST /api/filaments/import` (bulk)
- `POST /api/filaments/prusaslicer` (bulk) — its doc builder now spreads the collapsed entry, so omitted keys are uniformly not `$set` (it was gating temps but still clobbering `color`/`cost`/`density`/`diameter` with parse defaults)
- `POST /api/filaments/parse-ini` (new-filament prefill) — was leaking the suffixed name + baked per-nozzle values into the form

**Scalar non-clobber (P3):** a collapsed section now carries a shared scalar (`cost`/`density`/`color`/`diameter`) only when the section actually supplied it, so a partial/hand-crafted section can't `$set` parse defaults (`null` / `#808080` / `1.75`) over the base's real values. The normal export bakes these from the base, so they still round-trip.

**Lossiness (P3):** the per-nozzle calibration model is intentionally not reconstructed from a flat bundle — a fresh import of a multi-nozzle export lands the base filament without baked temps/calibrations by design; Backup & Restore is the lossless path. Documented in code on both bulk-import paths.

Tests: pure-helper cases (collapse, pass-through, sibling grouping, scalar-gate both ways) + route round-trips on all three endpoints (no orphans, base temps/scalars preserved, hints not stored). `tsc` + lint clean; full suite green (2426 tests).

@codex review

hyiger:
Both addressed in `48d49481`, plus a related self-review fix.

**P2 — validate per-nozzle calibration before persisting:** the calibration writes are atomic `updateOne` `$set`/`$push` that don't run schema validators. Now a probe sub-document is validated against the schema (`validate(["calibrations"])`) BEFORE any write, so an out-of-range baked value (`temperature=900`, `max_fan_speed=150`) rejects the whole sync with **400** and nothing persists — symmetric with the top-level `runValidators` path. Confirmed the calibration validators (nozzleTemp 0–600, fans 0–100, etc.) actually fire.

**P3 — gate vendor/type when collapsing a partial section:** `vendor`/`type` are now carried only when the suffixed section actually supplied `filament_vendor`/`filament_type` (same treatment as cost/density/diameter/color), so a partial/hand-crafted section can't `$set` `"Unknown"` over an existing base filament's identity. The normal export bakes them from the base, so round-trips are unaffected.

**Self-review P2 — per-row import resilience:** `/api/filaments/prusaslicer` had no per-row `try/catch`, so one bad section (e.g. a partial collapsed section failing required-field validation on create) would 500 the whole bundle. Wrapped each row to collect per-row errors into `errors[]` and return them — mirroring `/api/filaments/import`. This also makes the "surfaced as a per-row error" code comment accurate for both routes.

Tests: out-of-range temp + fan → 400 (nothing persisted); collapse omits unsupplied vendor/type; a hint-only section preserves base vendor/type/cost/density/color; a bad partial section → 200 + `errors[]`, not 500. `tsc` + lint clean; full suite green (2429).

@codex review

PR #879, opened 2026-06-28
fix(proxy): raise proxyClientMaxBodySize to 52mb so large snapshot restores aren't truncated (#878)

hyiger:
Fixes #878.

## Root cause
`src/proxy.ts` matches every `/api/*` request, so Next 16 buffers the request body to allow reads in both the proxy and the route. That buffer defaults to **10MB**, and per [the Next docs](https://nextjs.org/docs/app/api-reference/config/next-config-js/proxyClientMaxBodySize), over the limit it **keeps only the first 10MB and lets the request continue with a partial body — it does not error**. `next.config.ts` set no `experimental.proxyClientMaxBodySize`, so the 10MB default was in effect.

`POST /api/snapshot` accepts up to **50MB** (`MAX_SNAPSHOT_SIZE`), and the backup UI posts a multipart snapshot there. A valid 10–50MB backup was therefore truncated *before* the handler: `file.size` read ~10MB (< 50MB, so the route's 413 never fired), `file.text()` returned truncated JSON, `JSON.parse` threw, and the user saw a misleading **"Invalid JSON in snapshot file"**. The 10MB import routes (`MAX_UPLOAD_SIZE`) sat right at the same boundary.

## Fix
Set `experimental.proxyClientMaxBodySize = "52mb"` — above the 50MB route cap with headroom for the multipart envelope, so legitimate bodies reach the handler and each route's own size guard stays authoritative (a 50–52MB file now gets the route's real 413 instead of a truncation-induced 400).

## Tests
Added `tests/next-config-proxy-size.test.ts` pinning the invariant **proxy cap ≥ `MAX_SNAPSHOT_SIZE`** — the truncation is a proxy-layer behavior route-handler unit tests can't reproduce, so this guards against a future change silently dropping the cap or raising the snapshot cap past it. `tsc` + lint clean.

## Notes
- Affects desktop too — Electron embeds the standalone Next server, so this is the primary backup/restore UX.
- A full end-to-end check (POST a 10–50MB snapshot to a running server) would belong in the CI `smoke` job; flagging that as a possible follow-up rather than bundling a large fixture here.

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -11,6 +11,20 @@ const nextConfig: NextConfig = {
   env: {
     APP_VERSION: pkg.version,
   },
+  experimental: {
+    // GH #878: `src/proxy.ts` runs on every `/api/*` request, and Next buffers
+    // the request body so it can be read in both the proxy and the route. That
+    // buffer defaults to 10MB — over which Next silently keeps only the first
+    // 10MB and lets the request continue with a PARTIAL body (it does not error).
+    // `POST /api/snapshot` accepts up to 50MB (MAX_SNAPSHOT_SIZE), so a valid
+    // 10–50MB backup was truncated before the handler, parsed as partial JSON,
+    // and rejected with a misleading "Invalid JSON in snapshot file" instead of
+    // restoring (or returning the route's real 413). Raise the cap above the
+    // largest accepted route body (50MB) with headroom for the multipart
+    // envelope, so legitimate bodies reach the handler and the route's own size
+    // guard stays authoritative. Keep this >= MAX_SNAPSHOT_SIZE.
+    proxyClientMaxBodySize: "52mb",
+  },
   async headers() {
     // GH #225 — Content-Security-Policy:
     //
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

hyiger:
Addressed in `1cbe380c` — the invariant test now derives `MAX_SNAPSHOT_SIZE` from `src/app/api/snapshot/route.ts` and asserts the proxy cap >= that, so raising the route cap without raising the proxy cap fails the test instead of silently passing. Throws loudly if the literal can't be parsed.

@codex review

hyiger:
Addressed in `3ea30032` — the invariant now requires `proxyCap >= MAX_SNAPSHOT_SIZE + 1 MiB` headroom (for the multipart boundary + part headers), so equality / a route-cap-raised-to-match fails the test instead of silently truncating a near-limit snapshot. 52mb keeps 2 MiB of real margin over the 50mb route cap.

@codex review

PR #881, opened 2026-06-28
fix(ui): align the "show out of stock" pill with the quick-filter chips

hyiger:
## Problem
On the filament list header, the **"Show out of stock (N)"** pill was visibly larger than the quick-filter chips (All / Low stock / Has spools / No calibration) and sat slightly off their baseline.

Two causes:
- It rendered as a **sibling** next to `<QuickFilterChips>` with bigger classes — `text-sm` / `px-3` vs the chips' `text-xs` / `px-2.5`.
- The chips row carries `mb-2`; under the wrapper's `items-center`, that bottom margin pushed the chips up relative to the sibling button, so they didn't line up.

## Fix
Render the toggle **inside** the chip row via a new `trailing?: ReactNode` prop on `QuickFilterChips`, and match the chip size classes. Now it's a true peer of the chips — same flex row, gap, height, padding, and vertical baseline. Its gray active/inactive styling is unchanged (it's a toggle, not a filter chip).

## Verification
In-browser (dev server), all five pills measure identically:

| pill | height | font-size | padding-x | top / bottom |
|------|--------|-----------|-----------|--------------|
| All / Low stock / Has spools / No calibration / **Show out of stock** | 26px | 12px | 10px | 113 / 139 |

No component-render test added — the repo has no `@testing-library/react` setup, and this is a presentational/CSS-only change (verified visually + by measured DOM metrics).

### DIFF
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -1265,28 +1265,28 @@ export default function Home() {
       )}
 
       <div className="flex items-center justify-between gap-3 mb-2 flex-wrap">
-        <div className="flex items-center gap-2 flex-wrap">
-          <QuickFilterChips
-            active={quickFilter}
-            onChange={setQuickFilter}
-            counts={quickFilterCounts}
-          />
-          {quickFilter === "all" && !debouncedSearch && !typeFilter && !vendorFilter && outOfStockCount > 0 && hasAnyInStock && (
-            <button
-              onClick={() => setShowOutOfStock((s) => !s)}
-              aria-pressed={showOutOfStock}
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

hyiger:
Fixed in `9f39e5ca` — the tablist is now nested so it owns only `role="tab"` children, and the `aria-pressed` toggle is rendered as a sibling outside it (still in the same flex row, so size + baseline are unchanged). Verified in-browser: tablist children are all tabs, the toggle is outside the tablist and still measures 26px / 12px / 10px-pad aligned with the tabs.

@codex review

PR #906, opened 2026-06-28
fix(optresync): compare hex colors case-insensitively (#894)

hyiger:
Fixes #894.

## Root cause
`diffOptFields` / `classify` in `src/lib/optResync.ts` compared `color` and `secondaryColors` with strict `===` (via `valuesEqual`). Hex is case-insensitive, so a casing-only difference between the stored value and the OPT offer (or the provenance snapshot) read as a real change → a permanent `conflict` that re-surfaced on every "Check for updates".

## Fix
- Added `COLOR_FIELDS` + `canonicalizeColor` + `valuesEqualForField`, and used the color-aware comparison in both the diff (`current` vs `incoming`) and `classify` (`current` vs the snapshot).
- `buildOptSnapshot` now stores hex colors lower-cased, so the recorded offer is in a stable case. The diff also case-folds, so **pre-existing mixed-case snapshots compare correctly with no data backfill** (deferred per the triage).
- Non-color fields are unchanged (still strict structural equality).

## Tests
3 new cases in `tests/optResync.test.ts`: casing-only `color` and `secondaryColors` differences are not surfaced; an upper-cased local color equal to a lower-cased snapshot classifies as `adopt`, not `conflict`. Full `optResync` (38) + `opt-resync-route` (34) suites pass; `tsc` + lint clean.

### DIFF
--- a/src/lib/optResync.ts
+++ b/src/lib/optResync.ts
@@ -152,6 +152,26 @@ function valuesEqual(a: OptValue, b: OptValue): boolean {
   return a === b;
 }
 
+/** The managed fields whose values are hex colors. Hex is case-insensitive
+ *  (`#AABBCC` === `#aabbcc`), so comparing them with raw `===` (GH #894) made
+ *  a casing-only difference look like a real change → a permanent spurious
+ *  `conflict` re-surfaced on every re-check. Compare these case-folded. */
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #907, opened 2026-06-28
fix(print-history): validate only modified paths on spool-debit save() (#905)

hyiger:
Fixes #905.

## Root cause
A print-history job debits a spool via `filament.save()` (`src/app/api/print-history/route.ts:288/336/373`), which re-validates the **whole** document. A legacy out-of-range field elsewhere on the filament (e.g. a `temperatures.nozzle` stored before the max-600 validator existed) throws a `ValidationError` and blocks the unrelated spool-weight debit. The refund (`print-history/[id]` DELETE) and per-spool usage log have the identical exposure.

## Fix
Pass `{ validateModifiedOnly: true }` to the spool-mutation `save()` calls (POST debit + sequential-fallback + rollback, DELETE refund, usage log). They operate on docs **loaded from the DB** (all required fields present), so validating only the changed paths is safe and lets the debit proceed.

### Deviation from the triage's suggested option (worth a look)
The triage recommended a **schema-level** `validateModifiedOnly`. I implemented it, but it **broke required-field enforcement on create** — an omitted `name`/`vendor`/`type` is "unmodified" so its `required` validator is skipped (the existing "fails without required name/vendor/type" model tests caught this). So I reverted to the **route-scoped** form: creates still validate in full; only the loaded-doc debit saves relax re-validation of untouched fields.

## Tests
- `tests/print-history.test.ts`: a job debits a spool successfully even when the filament carries a legacy out-of-range field (was 5xx pre-fix).
- `tests/Filament.test.ts`: `save({validateModifiedOnly:true})` skips an unmodified legacy field but still rejects a *modified* out-of-range field.
- Full Filament-model (incl. the required-field tests) + print-history suites pass; `tsc` + lint clean.

### DIFF
--- a/src/app/api/filaments/[id]/spools/[spoolId]/usage/route.ts
+++ b/src/app/api/filaments/[id]/spools/[spoolId]/usage/route.ts
@@ -91,7 +91,9 @@ export async function POST(
     if (spool.usageHistory.length > MAX_SPOOL_HISTORY) {
       spool.usageHistory = spool.usageHistory.slice(-MAX_SPOOL_HISTORY);
     }
-    await filament.save();
+    // GH #905: usage logging only mutates this spool — validate modified paths
+    // only so a legacy out-of-range field elsewhere can't block the log/debit.
+    await filament.save({ validateModifiedOnly: true });
     return NextResponse.json(filament.toObject(), { status: 201 });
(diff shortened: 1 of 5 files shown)


hyiger:
@codex review

PR #908, opened 2026-06-28
fix(db): serialize cold-start migrations behind one in-flight promise (#898)

hyiger:
Fixes #898.

## Root cause
`dbConnect()`'s migration block ran inline with no concurrency guard. Two concurrent first-connects could both pass the all-flags short-circuit and **both** execute it; the GH#232 nozzle-split pass mints clones, so a concurrent double-run created **duplicate `Name #N` nozzle clones**.

## Fix
Serialize the block behind a single `cached.migrationsPromise` — the first connect runs it, concurrent callers await the same promise. The per-migration success flags still make each step idempotent + independently retryable; the promise is cleared in a `finally` so a future connect retries any step whose flag didn't flip. **In-process only** — a multi-process deploy (two desktops / desktop + Docker on one Atlas) would still need a DB-level lock (out of scope per the issue).

## Tests
`tests/mongodb.test.ts`: two concurrent `dbConnect()` calls with a re-armed nozzle-split run it **once** (original + exactly one clone, not two). Existing 16 migration tests still pass; `tsc` + lint clean.

### DIFF
--- a/src/lib/mongodb.ts
+++ b/src/lib/mongodb.ts
@@ -4,6 +4,10 @@ interface MongooseCache {
   conn: typeof mongoose | null;
   promise: Promise<typeof mongoose> | null;
   uri: string | null;
+  /** GH #898 — the in-flight migration run, so two concurrent first-connects
+   *  can't both execute the migration block (the nozzle-split pass would mint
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

PR #909, opened 2026-06-28
fix(import): use the shared parseCsv for filament CSV import (#888)

hyiger:
Fixes #888.

## Root cause
`src/app/api/filaments/import-csv/route.ts` split the file on `/\r?\n/` **before** quote-parsing, so a quoted cell containing an embedded newline (a multi-line value) was shredded across multiple rows and mis-columned. The local per-line parser also bypassed the shared `parseCsv`'s row cap.

## Fix
Parse with `parseCsv(content, { header: false })` (positional `string[][]`):
- embedded-newline-aware (the quoted newline stays in the cell);
- enforces `maxRows` itself — `CsvRowLimitExceededError` → 400;
- positional mode sidesteps the object-key `__proto__` vector (keys come from the header `mapping`, not the file);
- blank rows are filtered (non-header `parseCsv` returns them verbatim) to match the old `l.trim() !== ""`.

## Tests
`tests/filaments-import-export-route.test.ts`: a CSV with a quoted multi-line cell followed by another row imports **two** filaments (no shredded extra), the embedded newline is preserved, and the following row is intact. Existing import-csv + import-validation tests pass; `tsc` + lint clean.

### DIFF
--- a/src/app/api/filaments/import-csv/route.ts
+++ b/src/app/api/filaments/import-csv/route.ts
@@ -1,40 +1,10 @@
 import { NextRequest, NextResponse } from "next/server";
 import { mapHeaders, rowToImport, upsertImportRows } from "@/lib/importFilaments";
+import { parseCsv, CsvRowLimitExceededError } from "@/lib/parseCsv";
 import { assertMultipartFormData, getErrorMessage, errorResponse, checkFileSize } from "@/lib/apiErrorHandler";
 import { assertSameOriginRequest } from "@/lib/requestGuard";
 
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

hyiger:
Fixed in `874832e1` — parseCsv now gets a generous physical/DoS bound (2× the data cap) and the strict 10,000 **data-row** cap is enforced after blanks are filtered, so a capped export with a trailing blank line is no longer falsely rejected. Added a regression test. @codex review

PR #910, opened 2026-06-28
fix(spools): clear AMS slot before the mutating write on spool DELETE/PUT (#886)

hyiger:
Fixes #886.

## Root cause
`src/app/api/filaments/[id]/spools/[spoolId]/route.ts` ran the destructive `$pull` FIRST, then cleared the spool from printer AMS slots. If the slot-clear threw, the spool was gone but `Printer.amsSlots[].spoolId` still referenced it — and every retry 404s before reaching the clear (the `spools._id` filter no longer matches), so the dangling ref was uncleanable. PUT-retire had the same after-the-write ordering. This inverts the deliberate #261/#333 filament-level clear-before-write pattern.

## Fix
- **DELETE**: a precondition `Filament.exists(...)` read (keeps the 404 contract), then `assignSpoolToSlot(..., null)` BEFORE the `$pull`.
- **PUT-retire**: `assignSpoolToSlot(..., null)` BEFORE the retire `$set`.

`assignSpoolToSlot(..., null)` is an idempotent, no-match-safe `updateMany`, so a clear failure now leaves the spool present and the whole operation retryable.

## Tests
`tests/spool-subroute.test.ts`: a mocked slot-clear failure leaves the spool present (pre-fix it would be removed with a dangling slot ref), and a retry then succeeds + removes it. Existing DELETE tests pass; `tsc` + lint clean.

### DIFF
--- a/src/app/api/filaments/[id]/spools/[spoolId]/route.ts
+++ b/src/app/api/filaments/[id]/spools/[spoolId]/route.ts
@@ -119,17 +119,6 @@ export async function PUT(
       );
     }
 
-    // GH #268: retiring a spool excludes it from inventory, so it must not stay
-    // loaded in a printer AMS slot (the assignment route already refuses to
-    // *assign* a retired spool). GH #886: clear it BEFORE the retire write — the
-    // same clear-before ordering as DELETE — so a slot-clear failure leaves the
-    // spool active and the op retryable rather than retired-but-still-slotted.
-    // The clear is an idempotent no-match-safe updateMany, so pre-clearing for a
-    // spool that turns out not to exist (the 404 below) is harmless.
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

hyiger:
Both addressed in `b2e71dee`:
- **DELETE** now keeps the pre-clear (retryability) AND does a best-effort clear *after* the `$pull` — closing the window where a concurrent assignment slots the spool between pre-clear and delete.
- **PUT retire** reverted to clearing *after* the `$set` — the `spools._id` filter proves ownership first, so we never clear another filament's spool on a stale/cross-filament spoolId (PUT keeps the spool findable, so it doesn't need clear-before for retryability).

@codex review

PR #911, opened 2026-06-28
fix(import): round-trip calibration chamber temp on Bambu/Orca import (#891)

### LINKED ISSUE #891: Bambu/Orca import drops calibration chamber temp into the settings bag (round-trip break, same class as #508)
## Problem

The OrcaSlicer/Bambu Studio exporter emits a calibration's chamber temperature, but the import path has no inverse mapping for it. As a result, `export → calibrate-in-Bambu/Orca → re-import` does NOT restore the value to `calibrations[].chamberTemp`; it strands in the filament's `settings` passthrough bag instead. The per-nozzle calibration silently loses its chamber temp, and a subsequent calibration read returns `chamberTemp: null` for that nozzle.

This is the exact same class of bug as GH #508 (`bridge_fan_speed`), which was fixed and pinned — but `chamberTemp` was never added to the round-trip coverage.

## Evidence

- **Export emits it** — `src/lib/orcaSlicerBundle.ts:160-165`: `calibrationToOrcaSlicerKeys()` writes both `activate_chamber_temp_control` (`"1"`) and `chamber_temperature` (`String(chamberTemp)`) when `calibration.chamberTemp != null`. This is what the slicer fork reads from `GET /api/filaments/{id}/calibration`.
- **Field is first-class in the model** — `src/models/Filament.ts:368`: `chamberTemp: { type: Number, default: null, min: 0, max: 300 }` on the `calibrations[]` subdocument.
- **Import never inverts it**, all in `src/lib/bambuStudioImport.ts`:
  - `CalibrationHints` (lines 218-237) has no `chamberTemp` field.
  - `chamber_temperature` / `activate_chamber_temp_control` are in neither `STRUCTURED_KEYS` (61-99) nor `CALIBRATION_KEYS` (135-151).
  - `parseBambuStudioProfile`'s `calibrationHints` builder (313-341) never extracts `num(json.chamber_temperature)`.
  - `hasAnyHint` (351-359) doesn't reference it.
  - The settings-bag loop (365-378) only skips `STRUCTURED_KEYS`/`CALIBRATION_KEYS`, so both keys fall through into `filament.settings`.
- **Applier never writes it** — `src/lib/bambuStudioApply.ts:302-318`: `resolveAndApplyCalibration` builds the calibration `row` field-by-field and never sets `row.chamberTemp`.
- **Test gap** — `tests/bambuStudioImport.test.ts:228-246`: the calibration round-trip fixture omits `chamberTemp` (only the #508 `fanBridgeSpeed` case was added).

## Impact

A user who sets a chamber-temperature calibration loses it from the calibration row on any Bambu/Orca re-import. The value lands in the settings bag rather than `calibrations[].chamberTemp`, so the per-nozzle calibration silently degrades and the calibration API returns a null chamber temp. (Bounded to one calibration field, so P2.) A secondary effect: on the next export, `chamber_temperature` is re-emitted from the (now stale) settings copy while `calibrationToOrcaSlicerKeys` no longer emits it from the cleared `chamberTemp`, drifting the round-trip further.

## Suggested fix

Mirror the #508 (`fanBridgeSpeed`) fix:
1. Add `chamberTemp?: number` to `CalibrationHints`.
2. Add `chamber_temperature` and `activate_chamber_temp_control` to `CALIBRATION_KEYS` so they don't double-land in the settings bag.
3. Extract `num(json.chamber_temperature)` into `calibrationHints.chamberTemp` in `parseBambuStudioProfile`.
4. Include `chamberTemp != null` in the `hasAnyHint` computation.
5. Write `if (hints.chamberTemp != null) row.chamberTemp = hints.chamberTemp;` in `resolveAndApplyCalibration`.
6. Extend the round-trip test in `tests/bambuStudioImport.test.ts` to include `chamberTemp`, exactly like the `fanBridgeSpeed` case.

---
_Filed from an automated full-codebase audit; adversarially verified against the code before filing. Category: bug · domain: slicer-libs._

hyiger:
Fixes #891.

## Root cause
The exporter (`calibrationToOrcaSlicerKeys`, `src/lib/orcaSlicerBundle.ts:160-164`) emits chamber temp as `chamber_temperature` + the derived `activate_chamber_temp_control` flag, but `src/lib/bambuStudioImport.ts`'s `CALIBRATION_KEYS` and `CalibrationHints` had neither. So on import the chamber temp fell into the settings passthrough bag instead of `calibrations[].chamberTemp` — an export → calibrate → re-import cycle silently dropped it (same class as the #508 fan round-trip break).

## Fix
- Add `chamber_temperature` + `activate_chamber_temp_control` to `CALIBRATION_KEYS`. The latter is the single documented exception to the "every member maps to a hint" invariant — a derived enable flag re-emitted from `chamberTemp` on export.
- Add `chamberTemp` to `CalibrationHints`, extract it from `chamber_temperature`, include it in `hasAnyHint`, and write it to the calibrations[] row in `prepareBambuUpdate`.

## Tests
Extended the `calibrationToOrcaSlicerKeys` round-trip test: `chamberTemp` survives export → re-import as a hint and the chamber keys do **not** leak into the settings bag. bambu import + route suites (40) pass; `tsc` + lint clean.

### DIFF
--- a/src/lib/bambuStudioApply.ts
+++ b/src/lib/bambuStudioApply.ts
@@ -312,6 +312,7 @@ export async function resolveAndApplyCalibration(
   if (hints.fanMinSpeed != null) row.fanMinSpeed = hints.fanMinSpeed;
   if (hints.fanMaxSpeed != null) row.fanMaxSpeed = hints.fanMaxSpeed;
   if (hints.fanBridgeSpeed != null) row.fanBridgeSpeed = hints.fanBridgeSpeed;
+  if (hints.chamberTemp != null) row.chamberTemp = hints.chamberTemp; // GH #891
   if (parsed.temperatures.nozzle != null) row.nozzleTemp = parsed.temperatures.nozzle;
   if (parsed.temperatures.nozzleFirstLayer != null) row.nozzleTempFirstLayer = parsed.temperatures.nozzleFirstLayer;
   if (parsed.temperatures.bed != null) row.bedTemp = parsed.temperatures.bed;
--- a/src/lib/bambuStudioImport.ts
+++ b/src/lib/bambuStudioImport.ts
@@ -148,6 +148,14 @@ const CALIBRATION_KEYS = new Set<string>([
   "fan_min_speed",
   "fan_max_speed",
   "bridge_fan_speed",
+  // GH #891: the exporter emits chamber temp under these keys; pull them out of
+  // the settings bag so the round-trip lands chamberTemp on the calibration row.
+  "chamber_temperature",
+  // `activate_chamber_temp_control` is the ONE documented exception to the
+  // "every member maps to a hint" invariant above: it's a derived enable flag
+  // (no independent datum) that the exporter re-emits from chamberTemp, so it's
+  // dropped from settings on import and regenerated on the next export.
+  "activate_chamber_temp_control",
 ]);
 
 /**
@@ -230,6 +238,8 @@ export interface CalibrationHints {
   fanMinSpeed?: number;
   fanMaxSpeed?: number;
   fanBridgeSpeed?: number;
+  /** Chamber temp (chamber_temperature) → calibrations[].chamberTemp (#891). */
+  chamberTemp?: number;
   /** True when at least one calibration-relevant value was present. The
    * route uses this to decide whether to upsert a calibrations[] row vs
    * leave the filament's calibration data alone. */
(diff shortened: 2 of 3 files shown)


hyiger:
@codex review

hyiger:
Closing without merging — #891 turns out to be a **false positive**.

Verified against `main`: `chamber_temperature` was never in `CALIBRATION_KEYS` or `STRUCTURED_KEYS`, so the Bambu/Orca parser already routed it through the **settings passthrough bag**, and `filamentToOrcaSlicerKeys` re-emits the settings bag on export → it round-trips **losslessly**. The audit conflated it with #508 (the *fan* keys, which WERE excluded-from-settings-and-never-extracted → genuinely lost). This change (moving chamber into `CALIBRATION_KEYS`) is what *introduced* the two regressions Codex flagged (lost when no calibration resolves; the disabled `activate_chamber_temp_control` flag re-enabled on export). Reverting is the correct outcome — no fix needed.

PR #912, opened 2026-06-28
fix(nozzles): enforce one-printer-per-nozzle on the nozzle PUT printerIds path (#897)

hyiger:
Fixes #897.

## Root cause
`PUT /api/nozzles/{id}` syncs `Printer.installedNozzles` from the nozzle side via a `printerIds` body, but never enforced the #232 one-printer-per-nozzle invariant the printer routes guard with `findNozzleConflicts`. A `printerIds: [A, B]` body `$addToSet` the nozzle into **both** printers → one physical nozzle installed in multiple printers (the corruption #232 was built to prevent).

## Fix (AUTO-MOVE — per the chosen resolution)
The nozzle-side assignment is authoritative; the existing `$pull` already displaces the nozzle from its previous printer on a single-printer reassign (no conflict prompt needed — the nozzle form has no resolution modal). Added:
- reject `printerIds.length > 1` with **400** — one nozzle can't be in two printers;
- validate the single target is a valid, **active** printer (mirrors the printer routes' existence check) so the assignment can't silently no-op while the `$pull` still strips the old assignment.

**Deviation note:** the issue suggested a `findNozzleConflicts` **409**; the chosen **auto-move** keeps the reassign A→B flow working without a dead-end conflict toast.

## Tests
`tests/nozzles-route.test.ts`: multi-printer → 400 (neither printer mutated); single-printer reassign auto-moves A→B; non-existent target → 400; empty `printerIds` clears the assignment. 21 nozzle-route tests pass; `tsc` + lint clean.

### DIFF
--- a/src/app/api/nozzles/[id]/route.ts
+++ b/src/app/api/nozzles/[id]/route.ts
@@ -1,4 +1,5 @@
 import { NextRequest, NextResponse } from "next/server";
+import mongoose from "mongoose";
 import dbConnect from "@/lib/mongodb";
 import Nozzle from "@/models/Nozzle";
 import Filament from "@/models/Filament";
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

hyiger:
Fixed in `28416a9b` — printerIds validation (multi-printer reject + target existence) now runs BEFORE `Nozzle.findOneAndUpdate`, so a rejected assignment no longer partially renames the nozzle. Added a regression test. @codex review

hyiger:
Fixed in `f4aae9e2` — printerIds is now deduped (`new Set(map(String))`) before the one-printer check, so `[id, id]` (a duplicated selection/retry payload targeting one printer) is accepted instead of falsely rejected. Added a test. @codex review

hyiger:
Updated past main (the other 6 audit-fix PRs landed). @codex review

PR #913, opened 2026-06-28
fix(slicer): don't clobber a coextruded filament's null primary on sync-back (#883)

hyiger:
Fixes #883.

## Root cause
A coextruded filament is stored spec-pure as `color: null` + populated `secondaryColors[]`. `slicerExportColor` gives the slicer ONE representative color (`secondaryColors[0]`) under the single colour key. On sync-back the slicer echoes that hex and the naïve mapping wrote it onto the null primary → a hybrid (real primary + secondaries) that breaks the spec-pure shape. All three sync-back paths were affected: PrusaSlicer (`route.ts`), OrcaSlicer single-preset (`orcaslicer/route.ts`), and Bambu apply (`bambuStudioApply.ts`). The export already guards against inventing gray (#485) — the import lacked the symmetric guard.

## Fix
New pure helper `resolveSyncBackColor(stored, incomingHex)` (in `prusaSlicerBundle.ts`): returns `undefined` (don't write) when stored is coextruded **and** incoming == `secondaryColors[0]` (case-insensitive — the exported echo); otherwise returns the incoming hex (a genuine user edit, or a normal single-color filament). Wired into all three sync-back color writes; the Bambu routes' `existing` shape now carries `color` + `secondaryColors`.

**Known limitation** (documented, follow-up): a *variant* inheriting its parent's coextruded colors has its own `color` null + own `secondaryColors` empty, so it isn't detected as coextruded; resolving effective (parent-merged) values would close that — out of scope here.

## Tests
- `tests/prusaSlicerBundle.test.ts`: helper unit cases (echo suppressed incl. case-insensitive; real edit written; single-color written; null stored = create branch; absent incoming).
- `tests/api-route-correctness.test.ts`: PrusaSlicer sync-back of a coextruded filament preserves the null primary + secondaries; a genuinely different color IS applied.
- 120 tests across the affected suites pass; `tsc` + lint clean.

### DIFF
--- a/src/app/api/filaments/[id]/bambustudio/route.ts
+++ b/src/app/api/filaments/[id]/bambustudio/route.ts
@@ -186,6 +186,10 @@ export async function POST(
       // the unset branch unreachable in this route.
       type: existing.type ?? null,
       vendor: existing.vendor ?? null,
+      // GH #883: color + secondaryColors let resolveSyncBackColor detect the
+      // coextruded shape and suppress the exported-secondary echo on sync-back.
+      color: existing.color ?? null,
+      secondaryColors: existing.secondaryColors ?? null,
       diameter: existing.diameter ?? null,
       density: existing.density ?? null,
       cost: existing.cost ?? null,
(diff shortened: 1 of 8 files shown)


hyiger:
@codex review

hyiger:
Fixed in `3dd9fab9` — `resolveSyncBackColor` now takes the resolved parent and compares against EFFECTIVE secondaryColors, so a variant inheriting its parent's coextruded colors (own array empty) is detected and the echoed parent color isn't written onto the variant's primary. All three callers pass the parent (bambu apply already had it; prusa/orca routes load the parent's secondaryColors for a variant). Added a helper test for the inherited-coextruded case. @codex review

PR #914, opened 2026-06-28
fix(electron): await in-flight start in stopLocalMongo to avoid orphaning mongod (#900)

hyiger:
Fixes #900.

## Root cause
`stopLocalMongo()` only stops when `mongod` is already assigned, but `mongod` is set only AFTER `MongoMemoryServer.create()` resolves. If the app quits while a start is in flight (the `starting` promise pending), `stop()` sees `mongod === null`, does nothing, and the `create()` then resolves into an **orphaned mongod** holding the `dbPath` lock — with no handle to stop it, and the lock blocks the next launch.

## Fix
Await the in-flight `starting` promise before stopping, then stop whatever it produced. A failed start already rolls `mongod` back to null in `doStartLocalMongo`'s catch, so there's nothing to stop in that case.

## Verification
- `electron:typecheck` passes.
- **No unit test**: `electron/local-mongo.ts` imports `electron` + `mongodb-memory-server-core` and isn't in the vitest harness. **On-device confirmation of the quit-mid-boot path is owed** (per the issue, mongodb-memory-server's detached `mongo_killer` watchdog usually reaps the orphan as a backstop, but the explicit await is the correct fix).

### DIFF
--- a/electron/local-mongo.ts
+++ b/electron/local-mongo.ts
@@ -86,6 +86,19 @@ async function doStartLocalMongo(): Promise<string> {
  * Stop the embedded MongoDB instance.
  */
 export async function stopLocalMongo(): Promise<void> {
+  // #900: if a start is still in flight, `mongod` hasn't been assigned yet —
+  // returning here would let MongoMemoryServer.create() resolve AFTER we return,
+  // orphaning a mongod that holds the dbPath lock with no handle to stop it
+  // (the app quitting mid-boot). Wait for the in-flight start to settle first,
+  // then stop whatever it produced. A failed start rolls `mongod` back to null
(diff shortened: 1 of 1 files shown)


hyiger:
@codex review

hyiger:
Fixed in `4002cfd4` — added a `shuttingDown` flag set by `stopLocalMongo`; `doStartLocalMongo` checks it immediately after `create()` and self-stops the just-created server. So even if the await outlasts main.ts's 5s quit timeout and `app.quit()` fires first, the in-flight start can't leave an orphaned mongod holding the dbPath lock. Reset on a fresh start. (Electron module — `electron:typecheck` clean; on-device verification still owed.) @codex review

PR #915, opened 2026-06-28
fix(nfc): serialize transport ops to avoid auto-read racing a user op (#903)

### LINKED ISSUE #903: NfcService doesn't serialize transport ops — present-edge auto-read can race a user write/read on the shared PC/SC connection
## Problem

`NfcService` shares a single PC/SC connection handle (`this.activeReader`) across all high-level operations (`readTag`, `writeTag`, `formatTag`, `setReadOnly`), but nothing serializes those operations. Two overlapping ops race on `this.activeReader` and the one physical reader connection: one op's `disconnect()` (run from `withConnection`'s `finally`) can fire while another is mid-`transmit()`, or one op's `connect()` can reassign `activeReader` out from under another.

The automatic present-edge auto-read makes overlap realistic in normal use: placing a tag on the reader fires the auto-read AND is exactly the moment a user clicks "Write" on the detail page — so the auto-read interleaves with the IPC-driven write/probe.

## Evidence

- Single shared handle: `private activeReader: CardReader | null` declared at `electron/nfc-service.ts:113`; assigned in `connect()` at `electron/nfc-service.ts:395`; read in `transmit()` at `electron/nfc-service.ts:445-446` (`if (!reader) throw new Error("No active reader connection")`); torn down in `disconnect()` at `electron/nfc-service.ts:436-437`, called from `withConnection`'s `finally` at `electron/nfc-service.ts:480`.
- No mutex/queue anywhere: all ops just call `withConnection` (`readTag` `electron/nfc-service.ts:766,769,780,786`; `writeTag` `:792`; `formatTag` `:875,885`; `setReadOnly` `:954,962`). `withConnection` (`electron/nfc-service.ts:463-482`) is plain `connect → fn → finally{disconnect}` with no re-entry guard. Note `readTag` calls `withConnection` up to 3× sequentially (Bambu / NTAG / ISO-15693 paths) and `formatTag` twice, so even a single logical op has gaps where `activeReader` is reconnected/torn down.
- Auto-read fires automatically: `electron/main.ts:1593-1599` (`statusChange` → `triggerAutoRead(false)` on the present edge) and `:1609` (`presentAtConnect`). Its only guard is `lastAutoReadAt` + `AUTO_READ_COOLDOWN_MS = 4000` (`electron/main.ts:1537-1553`), which debounces auto-read against auto-read **only**.
- IPC handlers call the service directly with no shared lock: `nfc-read-tag` (`electron/main.ts:1158-1162`), `nfc-write-tag` (`:1164-1204`), `nfc-format-tag` (`:1206+`), `nfc-set-readonly` (`:1221`). The post-write read-back at `electron/main.ts:1194-1201` is a raw `nfcService.readTag()` that bypasses the cooldown entirely — a third uncoordinated path.

## Impact

Intermittent failed/corrupted NFC reads and writes when an auto-read overlaps a user-initiated op (a common timing). Symptoms: spurious `"No active reader connection"` errors, an in-flight op's connection torn down by another op's `finally`. Worst case: a `writeTag` interrupted mid-block by an interleaved `disconnect()` can leave a **partially-written tag** — corrupted data on physical media.

## Suggested fix

Serialize all `NfcService` transport ops behind a single in-class promise chain / mutex so `readTag`, `writeTag`, `formatTag`, `setReadOnly`, and the post-write read-back can never run concurrently (the second op awaits the first's `withConnection` `finally`). Additionally, suppress / defer the present-edge auto-read while a user-initiated op is in flight so a scan can't preempt a deliberate write.

---
_Filed from an automated full-codebase audit; adversarially verified against the code before filing. Category: bug · domain: electron-nfc-sync-label._

hyiger:
Fixes #903.

## Root cause
`NfcService` shares one PC/SC connection across all transport operations. The present-edge auto-read in `electron/main.ts` (one-shot silent verification fired on tag arrival) could run `readTag()` while a user-triggered `writeTag`/`formatTag`/`setReadOnly` was mid-flight — interleaving APDUs on the same reader.

## Fix (QUEUE-AND-WAIT, per the chosen resolution)
Serialize every public transport op through a single promise chain (`runExclusive`): a second op runs only after the first finishes; a failed op doesn't reject the next caller. Each public method now delegates to a private `*Impl` — no internal call goes through a public (serialized) entry point, so there's no re-entrant deadlock.

## Verification
- `electron:typecheck` passes.
- **No CI test**: `NfcService` instantiates `pcsclite()` (native module) and isn't in the vitest harness. **On-device confirmation is owed** (ACR1552U: an auto-read overlapping a user write no longer interleaves/errors).

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -643,13 +643,25 @@ const MAX_NFC_PAYLOAD_BYTES = 4096;
  * operations should report progress through their own status channel
  * instead.
  */
-function withIpcTimeout<T>(fn: () => Promise<T>, label: string, timeoutMs: number = IPC_TIMEOUT_MS): Promise<T> {
-  return Promise.race([
-    fn(),
-    new Promise<T>((_, reject) =>
-      setTimeout(() => reject(new Error(`IPC timeout: ${label} took longer than ${timeoutMs}ms`)), timeoutMs)
-    ),
-  ]);
+function withIpcTimeout<T>(
+  fn: (signal: AbortSignal) => Promise<T>,
+  label: string,
+  timeoutMs: number = IPC_TIMEOUT_MS,
+): Promise<T> {
+  // GH #915: abort on timeout so an operation that's still QUEUED behind a stuck
+  // op (e.g. an NFC write waiting on a stalled auto-read) can be dropped before
+  // it runs — without this the renderer promise rejects but the queued work
+  // still executes later, possibly against a different tag. Callers that don't
+  // need cancellation simply ignore the signal arg.
+  const controller = new AbortController();
+  let timer: ReturnType<typeof setTimeout>;
+  const timeout = new Promise<T>((_, reject) => {
+    timer = setTimeout(() => {
+      controller.abort();
+      reject(new Error(`IPC timeout: ${label} took longer than ${timeoutMs}ms`));
+    }, timeoutMs);
+  });
+  return Promise.race([fn(controller.signal), timeout]).finally(() => clearTimeout(timer));
 }
 
 /**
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

hyiger:
Fixed in `5a589bf7` — `withIpcTimeout` now aborts an `AbortController` on timeout and passes the signal to `fn`; the NFC handlers thread it to `nfcService`, and `runExclusive` checks it at **dequeue**, so a mutation whose 15s IPC timeout already fired while still queued is dropped before it touches the tag. An op already in progress is left to finish (the hazard is the queued one). Existing `withIpcTimeout` callers ignore the new signal arg (backward-compatible). Electron module — `electron:typecheck` clean; on-device confirmation owed. @codex review

PR #918, opened 2026-06-28
fix(filaments): clear stale colorName on color sync-back (#885) + correct permanent-delete guard comment (#884)

hyiger:
Two filament `[id]` route fixes, batched (same file).

### #885 — sync-back leaves stale `colorName`
The sync-back handler wrote the primary hex but never touched the paired `colorName`, so after a slicer color change the stored name kept describing the old hex (name chip + CSV/XLSX exports). Now clears `colorName` when the synced hex actually changes the color. **Gated on `resolvedColor !== undefined`** so the #883/#913 coextruded-echo suppression path (which intentionally leaves `color` unwritten) doesn't clear the name. Cleared rather than reverse-looked-up — an arbitrary slicer hex won't reliably map to a named color.

### #884 — permanent-delete variant guard comment vs query
The `?permanent=true` guard's comment claimed it counts only *trashed* variants, but the query has no `_deletedAt` clause and counts **all** non-purged variants (active or trashed). No functional bug — the guard is stricter than the comment — but a maintainer "fixing" the query to match could let a purge orphan an active variant. Reworded the comment to match the (safe, broader) query + warn against narrowing it; reworded the error string to drop the inaccurate "in the trash". No query change.

**Tests:** #885 — changed-color clears name, same-color keeps it, coextruded-echo leaves it untouched (3 cases). #884 — updated the existing orphan-refusal assertion to the new error text.

@codex review

### DIFF
--- a/src/app/api/filaments/[id]/route.ts
+++ b/src/app/api/filaments/[id]/route.ts
@@ -560,7 +560,15 @@ export async function POST(
         // because an arbitrary slicer hex won't reliably map to a named colour.
         // Gated on resolvedColor !== undefined above so the coextruded-echo
         // suppression path (which leaves color untouched) doesn't clear it.
-        if (resolvedColor !== filament.color) update.colorName = null;
+        // Codex P2 (#918): compare case-insensitively — the schema accepts
+        // mixed-case hex, so a `#ff0000` → `#FF0000` sync isn't a real colour
+        // change and must NOT drop the name.
+        if (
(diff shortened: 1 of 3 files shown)


hyiger:
Fixed in `a843506e` — the clear decision now normalizes both sides (`toLowerCase()`), so a case-only hex difference (`#ff0000` → `#FF0000`) is treated as a no-op and the `colorName` is preserved. Added a case-only test. @codex review

PR #919, opened 2026-06-28
fix(inventory): compute last-dried as MAX cycle date, not last array element (#887)

hyiger:
Fixes #887.

The dry-cycle POST honors an arbitrary client `date` and `$push`es with no `$sort`, so a **backdated** dry cycle lands last in the array. Two readers took the last element as 'newest', so a backfilled past cycle was reported as the most-recent dry:

- **/inventory** (`spools/by-location`): `$arrayElemAt[-1]` → replaced with `$max` over `dryCycles.date` (`$max` yields null on empty/missing).
- **dashboard** 'needs drying' tile: `cycles[length-1].date` → replaced with a max-by-time loop (so a recent dry stored before a backdated one no longer flags the spool as due).

**Tests:** both routes seed out-of-order `dryCycles` (recent first, older appended last) and assert the newest date wins.

@codex review

### DIFF
--- a/src/app/api/dashboard/route.ts
+++ b/src/app/api/dashboard/route.ts
@@ -179,8 +179,18 @@ export async function GET() {
       for (const s of f.spools || []) {
         if (s.retired) continue;
         const cycles = s.dryCycles || [];
-        const lastCycle = cycles.length > 0 ? cycles[cycles.length - 1].date : null;
-        const lastCycleMs = lastCycle ? new Date(lastCycle).getTime() : 0;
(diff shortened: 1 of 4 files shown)


hyiger:
Fixed in `8b1620a5` — switched to an explicit `$reduce` (`{$max:["$$value","$$this.date"]}` per element). The original `$max` was an expression nested inside `$push` (not a `$group` accumulator, so it did traverse the array — the out-of-order-cycles test passed), but the `$reduce` makes the per-element traversal unambiguous. @codex review

PR #920, opened 2026-06-28
fix(snapshot): cap raw body before buffering (#889) + field-gate array ObjectId coercion (#890)

hyiger:
Two snapshot-restore hardening fixes, batched (same file).

### #889 — raw body buffered before size check
The non-multipart restore branch buffered the entire body via `request.text()` *before* the 50 MB `text.length` check — a same-origin / non-browser client (curl, the desktop renderer) could force full allocation of a multi-GB body. Added `checkContentLength` (the GH #676 helper the sibling `prusaslicer` / `nfc-decode` raw-body routes already use) to short-circuit on `Content-Length` before buffering; the post-buffer check stays as belt-and-suspenders.

### #890 — array ObjectId coercion ignored the field name
`restoreTypes` coerced **every** 24-hex array element to an `ObjectId` regardless of key — asymmetric to the field-gated scalar path. A future (or `settings`-bag) string array whose elements happened to be 24 hex chars would have its type silently changed on restore. Added `OID_ARRAY_FIELDS = {compatibleNozzles, installedNozzles}` and gated the array coercion on it. Object-element recursion stays unconditional; docstring corrected.

**Tests:** #889 — oversized `Content-Length` → 413. #890 — `installedNozzles` 24-hex still coerces to ObjectId; a 24-hex string in a `settings` array round-trips as a string.

@codex review

### DIFF
--- a/src/app/api/snapshot/route.ts
+++ b/src/app/api/snapshot/route.ts
@@ -2,6 +2,7 @@ import { NextRequest, NextResponse } from "next/server";
 import mongoose from "mongoose";
 import dbConnect from "@/lib/mongodb";
 import { assertSameOriginRequest } from "@/lib/requestGuard";
+import { checkContentLength } from "@/lib/apiErrorHandler";
 import Filament from "@/models/Filament";
 import Nozzle from "@/models/Nozzle";
 import Printer from "@/models/Printer";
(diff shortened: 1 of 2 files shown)


hyiger:
Fixed in `9a256046` — the post-buffer fallback now uses `Buffer.byteLength(text, "utf8")` so the 50MB cap is byte-based even when Content-Length is absent/wrong, matching the sibling raw-body routes. @codex review

Issue #892, opened 2026-06-28
Bambu Studio import routes skip the inverted-nozzle-range (min>max) guard the OrcaSlicer sync enforces


hyiger:
## Problem

The Bambu Studio import routes can persist an inverted nozzle temperature range (`nozzleRangeMin > nozzleRangeMax`) that the OrcaSlicer sync route — operating on the same `temperatures` model — explicitly rejects with a 400. The validation is asymmetric across two routes that share one schema.

The schema's per-field validators (`min: 0`, `max: 600`) cannot express the cross-field relationship `min > max`, so `runValidators: true` alone does not catch it. The OrcaSlicer route added a dedicated cross-field guard for exactly this reason (#574/#577/#618); the Bambu routes never got the equivalent.

## Evidence

- The Bambu parser extracts the two range endpoints: `src/lib/bambuStudioImport.ts:278-279` (`nozzle_temperature_range_low/high` → `temperatures.nozzleRangeMin/nozzleRangeMax`).
- `buildStructuredUpdate` merges them into the written `temperatures` object: `src/lib/bambuStudioApply.ts:234-241`.
- Per-id Bambu sync writes that update with only `runValidators: true` and no range guard: `src/app/api/filaments/[id]/bambustudio/route.ts:248-260`.
- Bulk Bambu import writes via `prepareBambuUpdate` (→ `buildStructuredUpdate`) across all three phases, also with no range guard: `src/app/api/filaments/bambustudio/route.ts` (e.g. lines 133/170/224/249, all `runValidators: true`). The file imports no `temperatureRange` helpers.
- Schema validators are per-field only: `src/models/Filament.ts:330-331` (`nozzleRangeMin`/`nozzleRangeMax` each `min: 0, max: 600`).
- The OrcaSlicer sync route DOES guard this: `src/app/api/filaments/[id]/orcaslicer/route.ts:202-217`, using `isInvertedNozzleRange` + `effectiveNozzleRangeForUpdate` + `inheritNozzleRangeFromParent` (imported at lines 12-16), resolving parent inheritance for variants and returning 400.

## Impact

[...]

PR #921, opened 2026-06-28
fix(slicer): guard inverted nozzle range on Bambu import (#892) + dedup per-id Bambu route (#893)

hyiger:
Two Bambu-import fixes, batched (same area).

### #893 — per-id route duplicated the apply pipeline
The per-id sync route hand-inlined `buildStructuredUpdate` + `mergeSlicerSettings` + `resolveAndApplyCalibration` — the exact sequence the shared `prepareBambuUpdate` centralises and the bulk route already uses — including a misleading dead `new Set(Object.keys(...).filter(() => false))`. Replaced with `prepareBambuUpdate` so the two paths can't drift. Dropped the now-unused imports.

### #892 — Bambu routes skipped the inverted-nozzle-range guard
Both Bambu routes persisted an inverted range (`min > max`) that the OrcaSlicer sync route rejects (the per-field 0–600 validators can't express the cross-field relationship). Added a shared pure combinator **`isUpdateNozzleRangeInverted(body, ownTemps, parentTemps)`** (own effective range → parent inheritance → inversion test). `prepareBambuUpdate` now returns a `nozzleRangeInverted` flag that **both** Bambu routes map to 400 (via a shared per-phase `bambuPayloadError` gate on the bulk route). **Refactored the OrcaSlicer route to call the same combinator** so all three slicer-sync paths share one implementation.

**Tests:** unit tests for the combinator (incl. inheritance-induced inversion); per-id + bulk routes reject an inverted range with 400 and persist nothing; existing orca #618 range tests still pass.

@codex review

### DIFF
--- a/src/app/api/filaments/[id]/bambustudio/route.ts
+++ b/src/app/api/filaments/[id]/bambustudio/route.ts
@@ -11,10 +11,7 @@ import {
   parseBambuStudioProfile,
   type BambuParseResult,
 } from "@/lib/bambuStudioImport";
-import {
-  buildStructuredUpdate,
-  resolveAndApplyCalibration,
-} from "@/lib/bambuStudioApply";
+import { prepareBambuUpdate } from "@/lib/bambuStudioApply";
 import {
   assertMultipartFormData,
   checkFileSize,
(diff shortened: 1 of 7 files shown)


hyiger:
Fixed in `ab178075` — `nozzleRangeInverted` is now gated on whether the parsed profile actually carried `nozzle_temperature_range_low/high`, so a sync that only touches e.g. `nozzle_temperature` no longer 400s on legacy data with a pre-existing inverted range (matches the OrcaSlicer route's `touchesNozzleRange` gate). Added a regression test. @codex review

PR #924, opened 2026-06-28
docs(security): require proxies to preserve Host for the CSRF guard (#899)

### LINKED ISSUE #899: CSRF guard's Origin-vs-Host fallback ignores X-Forwarded-Host — silent 403 for some clients behind the docs-recommended reverse proxy
## Problem

`assertSameOriginRequest` (the trusted-origin / CSRF guard on every mutating API route) has a fallback branch that runs when `Sec-Fetch-Site` is absent but an `Origin` header is present (older browsers, embedded webviews, or an intermediary that strips Fetch Metadata). In that branch it compares the `Origin` authority against the **raw `Host` header** and never consults `X-Forwarded-Host` / `X-Forwarded-Proto`.

The docs explicitly recommend running a network-exposed instance behind an **authenticating reverse proxy** with the app bound to loopback (`docs/setup.md:128`: "put Filament DB behind an authenticating reverse proxy (nginx/Caddy/Authelia)… bind Filament DB itself to loopback… The proxy must be the only way in."). A proxy whose `Host` is not explicitly preserved (e.g. nginx's bare `proxy_pass`, which rewrites `Host` to the upstream address such as `127.0.0.1:3000`) presents an internal `Host` to the app while the browser's `Origin` is the public name. `originHostname !== hostParts.hostname` → 403 on every same-origin mutating request from those clients.

## Evidence

- `src/lib/requestGuard.ts:62` — `const hostHeader = request.headers.get("host");` (raw Host only)
- `src/lib/requestGuard.ts:89-91` — `const hostParts = splitAuthority(hostHeader); … if (originHostname !== hostParts.hostname || originPort !== hostPort) { return 403 }`
- No `X-Forwarded-Host` / `X-Forwarded-Proto` handling anywhere in `src/` (`grep -rni x-forwarded src/` → no hits).
- The in-code comment at `src/lib/requestGuard.ts:84-85` claims "no false-reject behind a proxy" — but that only addresses port normalization (`:443`/`:80`), not a rewritten Host.
- Reverse-proxy deployment is docs-recommended: `docs/setup.md:128`, `README.md:3`, `docs/desktop.md:52`, `docs/usage.md:689`, `docs/tutorial.md:615`.

## Impact

Scoped: modern browsers send `Sec-Fetch-Site` on same-origin mutations and pass via the earlier branch, so they are unaffected. The breakage is limited to the `Origin`-only fallback path — clients lacking Fetch Metadata — AND only when the proxy rewrites `Host`. For that subset, every first-party mutation (create/edit/delete filaments, spools, etc.) returns a hard-to-diagnose 403 in exactly the deployment the docs steer users toward. P3 footgun.

## Suggested fix

Two options, in order of preference:

1. **Document the requirement** (lowest risk): the reverse-proxy section in `docs/setup.md` should state that the proxy must preserve the original Host (`proxy_set_header Host $host` for nginx; Caddy preserves it by default). Add a code comment at `requestGuard.ts:84` correcting the "no false-reject behind a proxy" claim to scope it to ports only.

2. **Honor forwarded headers** in the fallback: also accept a match against `X-Forwarded-Host` (normalizing the port via `X-Forwarded-Proto`) when present. **Trust caveat:** these headers are spoofable by a direct (non-proxied) attacker, so this is only safe because the docs already require the app to be loopback-bound with the proxy as the only ingress. Gate it on a documented "trust forwarded headers" assumption and call that out next to the change.

---
_Filed from an automated full-codebase audit; adversarially verified against the code before filing. Category: footgun · domain: security-lib._

hyiger:
Fixes #899.

The CSRF guard's `Origin`-only fallback (older browsers / embedded webviews that don't send `Sec-Fetch` metadata) compares the `Origin` authority against the **raw `Host` header** and never consults `X-Forwarded-Host`. Behind a reverse proxy that rewrites `Host` to the upstream address (e.g. nginx's bare `proxy_pass` → `Host: 127.0.0.1:3000`), those clients' requests look cross-origin and get a `403` on every first-party mutation — in exactly the loopback-bound-behind-a-proxy deployment the docs steer users toward.

Shipped the **low-risk option** (the triage's recommendation): documentation + a comment correction, no behavior change.
- `docs/setup.md`: the reverse-proxy section now requires preserving `Host` (`proxy_set_header Host $host` for nginx; Caddy default), scoped to the non-Fetch-Metadata fallback.
- `requestGuard.ts`: corrected the "no false-reject behind a proxy" comment to scope it to ports only, and documented why forwarded headers are deliberately **not** trusted (spoofable by a direct client; the loopback-bind + sole-ingress requirement is the mitigation).

Honoring `X-Forwarded-Host` in code is the heavier alternative; left out because it only becomes safe under the same loopback-bind assumption the docs already require, and it adds a spoofable surface.

@codex review


hyiger:
Fixed in `8c11dd8a` — both corrections applied: (1) the comment + docs no longer claim modern browsers are unaffected (they send `Origin` on same-origin mutations, and the guard compares it to `Host` regardless of `Sec-Fetch`), and (2) the nginx snippet now uses `$http_host` (full authority incl. port) instead of `$host` (which drops the port and would still 403 on a non-default port like :3456). @codex review

PR #927, opened 2026-06-29
feat(opentag3d): tag-type-aware NFC write/erase/read-only for SLIX2 + NTAG215

hyiger:
Promotes **OpenTag3D writing** from the dev-only CLI harness to a real product feature, and makes the existing **Write NFC** button + **Settings → Devices → NFC Tools** (Erase, Set read-only / Make writable) **auto-detect the loaded tag** and act on the right standard:

- **SLIX2 → OpenPrintTag** (ISO-15693 `FF FB`) — unchanged, extracted to `writeSlix2Impl`.
- **NTAG213/215/216 → OpenTag3D** (Type-2 `FF B0`/`FF D6`) — new.

### How it works
- `NfcService.detectTag()` (IPC `nfc-detect-tag`) reports `{family, standard, formatted, readOnly}`. The renderer picks the encoder — `filamentToOpenTag3DFields`→`encodeOpenTag3D` (NTAG) vs `generateOpenPrintTagBinary` (SLIX2) — and sends a `standard` discriminator on `nfc-write-tag`; the service routes the transport and **refuses a chip/standard mismatch** (`TAG_TYPE_MISMATCH`).
- New NTAG transport: `writeNtagPage` (`FF D6`, the exact APDU the dev harness **proved on the ACR1552U + NTAG215**), `writeNtagImpl`/`formatNtagImpl`/`setReadOnlyNtagImpl`, capacity-checked, write-then-verify.
- The NFC Tools card now shows the loaded tag type (e.g. "NTAG215 · OpenTag3D").

### Safety (the key design decision)
- **NTAG read-only = the REVERSIBLE Type-2 CC byte-3 write nibble** (`0x0`⇄`0xF`). **The OTP static/dynamic lock bytes are NEVER written** — they're permanent and would brick the tag. Verified: every `writeNtagPage` call targets only page 3 (CC), page 4 (TLV), or page 5→last-user-page. Same "no permanent hardware lock, no bricking risk" posture as SLIX2; Erase / Make Writable clears it.
- **Blank-tag sizing** via `GET_VERSION` (NXP datasheet storage bytes `0x0F/0x11/0x13`); an unrecognized size **refuses** rather than guessing.

### Layers
- **L1 (`dc3466f7`):** pure `filamentToOpenTag3DFields` + `wrapOpenTag3DType2` + Type-2 CC helpers, unit-tested (round-trip).
- **L2/L3 (this):** Electron transport + IPC + renderer codec-selection + UI + i18n (en/de parity).

### Verification
tsc, electron:typecheck, lint, i18n parity, and 105 NFC/opentag3d/ndef/codec tests all green. SLIX2 path byte-for-byte unchanged.

### ⚠️ Hardware status
`FF D6` page write is proven on the ACR1552U+NTAG215 (harness). The **full product write/erase/read-only flow + GET_VERSION are pending on-device confirmation** — I'll validate on real NTAG215 hardware before merge. NTAG *read* is already verified.

@codex review


hyiger:
Fixed all three in `9679f72e`:
1. **NTAG213 core fallback** — `detectTag` now returns `ndefCapacity` (CC size, or GET_VERSION for a blank tag); the renderer emits the Core (112B) image when the Extended (214B TLV) won't fit, so NTAG213 writes succeed. Test pins that Core fits 144B and Extended doesn't.
2. **Capacity-before-format** — `writeNtagImpl` builds + capacity-checks the TLV before writing the page-3 CC, so a too-large payload no longer half-formats a blank tag.
3. **Byte-exact verify** — the read-back now compares the OpenTag3D record payload byte-for-byte against what was written, not just `tagSource`.

tsc + electron:typecheck + lint + tests green. @codex review

hyiger:
Both fixed in `df30d086`:
- **Erase clamp (safety)** — `writeNtagPage` now refuses pages outside [3,255] (never page 0-2 static lock, never an APDU-wrapped page); CC-derived capacity is clamped to the largest real NTAG (872B); and the erase zero-fill extent is bounded by the **authoritative GET_VERSION** size (conservative NTAG213 floor when unknown), so a corrupt/inflated CC can't drive the wipe into a smaller chip's config/lock pages.
- **serial → spoolUid** — the decoder now surfaces the OpenTag3D `serial` (where we store the spool instanceId) as `spoolUid`, so a scan of a tag we wrote keeps per-spool matching via `/api/filaments/match`; a commercial vendor serial falls through to name/vendor/type.

tsc + electron:typecheck + lint + 95 NFC-area tests green. @codex review

hyiger:
Fixed in `5e27a25f` — the formatted-NTAG write capacity (and detectTag's reported capacity, which drives the renderer's Core/Extended choice) now takes **min(CC, GET_VERSION)**: GET_VERSION is authoritative and catches an inflated/corrupt CC on a smaller chip, so an Extended TLV can no longer be written past a 213's user area into its config/lock pages. Falls back to the clamped CC only when GET_VERSION is unavailable (the common formatted-tag case, where the CC is correct). Matches the erase path's authoritative bound. @codex review

hyiger:
Fixed in `5385fbe2` — the weight-update path now passes `requireExtended`; if the chip forces a Core-only image (no `measured_filament_weight`/`serial`), it aborts with a clear "tag too small, not changed" message instead of a false success. The regular Write path keeps its lossy-but-intentional Core fallback + notice. @codex review

hyiger:
Fixed in `e781962b` — the Write path now surfaces an explicit Core-only notice when an NTAG213 can't hold the Extended fields (spool id + remaining weight), so the user knows a later scan may not match the spool. The weight-update path still hard-refuses. @codex review

hyiger:
Fixed in `dfff662f` — an over-length spool ID is now OMITTED (write the rest + a notice) rather than truncated into a non-round-trippable `serial`, so a scan never resolves to a wrong/truncated id. Normal 10-hex ids fit and round-trip. Test added. @codex review

hyiger:
Pushed `e033b77e` from on-device NTAG215 testing:
- **Read staleness** (read-only not clearing, post-write 'no record', erase-then-write still read-only): root cause is the ACR1552U serving Type-2 `FF B0` reads from an internal card buffer refreshed only on card re-activation — so a cross-connection read after an `FF D6` write returned stale pages. (SLIX2/`FF FB` is a direct pass-through, hence it worked.) NTAG mutating ops now disconnect with `SCARD_RESET_CARD` so the next read re-reads fresh.
- **Material base/mod split**: 'PA12-CF' now → base 'PA12' + modifier 'CF' instead of truncating to 'PA12-'.
- **NFC Tools subcaption** now says OpenPrintTag and OpenTag3D.

@codex review

hyiger:
Pushed `7ac25ede` — **NTAG read-only removed (hardware finding: the NTAG21x CC page is OTP).** On-device, an FF D6 write that clears a CC bit returns SW 9000 but the bit persists (`0x0→0xF` works, `0xF→0x0` doesn't), so a CC-nibble read-only is irreversible — it permanently stuck a test tag. The static/dynamic lock bytes are also OTP, so there's no reversible NTAG read-only. Read-only is now SLIX2/OpenPrintTag-only (its CC byte is rewritable EEPROM); NTAG `setReadOnly` throws `NTAG_READONLY_UNSUPPORTED` and the buttons disable for NTAG. The write path no longer honors the OTP nibble, which recovers any stuck tag (data pages are never locked). @codex review

hyiger:
Fixed in `e34bca1c` — readNtagTag now reports an empty NDEF message (zero records, e.g. a just-erased tag's `03 00 FE` TLV) as 'Blank or unformatted' rather than the foreign-tag 'No NDEF record' error, so ensureTagWritable's blank-bypass fires and writing to a freshly-erased NTAG no longer prompts-to-overwrite / fails the weight-update closed. @codex review

hyiger:
Already addressed in `e34bca1c` (this re-flags the same root cause from the `formatNtagImpl` side). I took the 'distinguish empty from foreign' option you suggested, on the READ side: `readNtagTag` now reports `records.length === 0` — an empty NDEF message, e.g. an erased tag's `03 00 FE` TLV — as "Blank or unformatted" instead of "No NDEF record". The detail-page probe's blank-bypass then treats a freshly-erased NTAG as blank, so the erase-then-write flow no longer shows the overwrite prompt. Doing it on the read side also covers any empty-NDEF tag, not just our own erase output. (Hardware-confirmed: erase→write works on NTAG215.)

hyiger:
@codex review

hyiger:
Fixed in `87141cad` — split writeNtagImpl so the byte-exact verify runs in a **fresh post-reset connection**: phase 1 writes (resetAfter re-powers the card on disconnect), phase 2 reconnects and reads back/compares. The verify now reads the actual post-write chip state instead of a possibly-stale same-connection READ BINARY buffer, so it can't falsely fail a good write; a mismatch is a real failure. Per-page FF D6 SW=9000 still confirms each write too. @codex review

hyiger:
Fixed in `f1890033`. Both detectType2Head and readNtagTag now require the NXP UID manufacturer byte (`head[0] === 0x04`) to classify a tag as a Type-2 NTAG, instead of only rejecting `0xE1`. A real NXP NTAG21x's page-0 byte-0 is always 0x04 (immutable, blank or formatted); a SLIX2's block-0 byte-0 is only ever 0xE1 (formatted) or 0x00 (blank) — so a blank SLIX2 now correctly defers to the ISO-15693 path and OpenPrintTag write/erase no longer misroutes to the NTAG path. Rides on the already-hardware-verified page-0..3 read; does not gate on the unverified GET_VERSION. @codex review

hyiger:
Fixed in `2a5ec259`. New `safeNtagNdefBytes` helper: when GET_VERSION can't authoritatively size a formatted NTAG, it caps to the conservative NTAG213 extent (144 B) instead of trusting a possibly-inflated CC — so a corrupt CC can't drive an Extended write into a smaller chip's lock/config pages. Applied to BOTH writeNtagImpl and detectTagImpl so detect/write stay in lockstep (unverifiable size → renderer encodes Core, which fits + shows the opentag3dCoreOnly notice, rather than an Extended payload the write would refuse). GET_VERSION works on the ACR1552U in practice (a blank-NTAG write requires it), so this conservative branch is the edge case. @codex review

hyiger:
Fixed in `df1442d9`. formatNtagImpl now routes the CC size through the same `safeNtagNdefBytes` helper as write/detect, so Erase writes the conservative NTAG213 extent (not the existing CC) when GET_VERSION can't verify the size — an inflated CC no longer survives the erase to mislead other tools. The wipe extent was already bounded; this makes the rewritten CC honest and keeps all three paths (write/detect/erase) in lockstep. GET_VERSION works on the ACR1552U in practice, so a real 215/216's CC is preserved. @codex review

hyiger:
Fixed in `14ebb3ce`. Erase now takes the AUTHORITATIVE GET_VERSION size directly (not `min(CC, verSize)`) for the rewritten CC + zero-fill, so a real NTAG215/216 previously mis-formatted as a 213 (or with MLEN 0) is restored to its true capacity. It still can't write back a lying CC — the size is GET_VERSION's, not the old CC's, so an inflated 213 CC is replaced with 144; a blank tag is still refused when unsizable. write/detect keep `min(CC, verSize)` because they write a payload to an EXISTING CC (must respect the declared NDEF area), whereas erase REPLACES the CC and so uses the physical size. @codex review

hyiger:
Fixed in `e6ed2e6d`. Added the GH #437 parity guard to BOTH formatNtagImpl (the path you flagged) and writeNtagImpl (the same gap): refuse a page-3 CC byte that's neither 0xE1 (NDEF) nor 0x00 (blank), so a proprietary/non-NDEF NXP NTAG isn't reformatted/overwritten and its data clobbered. The byte-0 NXP guard only proves it's an NXP Type-2 chip, not that it's ours to write. Mirrors writeSlix2Impl. @codex review

hyiger:
Fixed in `d163d627`. Both writeNtagImpl and formatNtagImpl now re-read pages 0–3 inside their own mutating connection (via detectType2Head) and derive ccMagic/capacity from that fresh head — so a tag swapped after the probe can't have the guards applied to the old tag while the writes hit the new one; a chip that's no longer an NTAG aborts with TAG_TYPE_MISMATCH. The probe in writeTagImpl/formatTagImpl now only selects the path. **Also caught a latent bug while here:** the verify-split (87141cad) left writeNtagImpl opening with `return this.withConnection(...)`, making the fresh-connection verify phase unreachable — switched to `await` so the byte-exact read-back actually runs now. @codex review

hyiger:
Fixed in `41e8a6eb` (both threads):

**P2 (MIFARE Classic):** writeNtagImpl + formatNtagImpl now require GET_VERSION to confirm the NTAG family before ANY page write — MIFARE Classic doesn't support GET_VERSION, so a non-Bambu Classic card with an NXP 0x04 UID + coincidental 0xE1/0x00 byte-12 is now refused (fail closed) instead of having its blocks clobbered. GET_VERSION works on the ACR1552U (a blank-NTAG write already required it), so only the unverifiable edge case is refused. Erase uses the verified size directly; write keeps min(CC, verSize).

**P3 (detect accuracy):** detectTag now parses the NDEF records (like readNtagTag) instead of trusting the CC byte — 0 records ⇒ blank, an opentag3d record ⇒ opentag3d, any other ⇒ foreign NDEF. The Settings label distinguishes the foreign case (new `settings.nfcLoadedNtagForeign` key, en+de). @codex review

hyiger:
Fixed in `90b69222`. The SLIX2 detect branch now parses NDEF records (via readNfcVImage, mirroring setReadOnlyImpl + the NTAG branch) instead of trusting the Type-5 CC byte: 0 records ⇒ blank, OpenPrintTag record ⇒ openprinttag, OpenTag3D record ⇒ opentag3d (read-only support, #864), any other ⇒ foreign NDEF. New `settings.nfcLoadedSlix2OpenTag3d` + `settings.nfcLoadedSlix2Foreign` labels (en+de). detect is now symmetric across NTAG + SLIX2. @codex review

hyiger:
Fixed in `ea0b4ea1`. The SLIX2 detect catch now treats the no-NDEF/blank error signals (an erased SLIX2's TLV is just the FE terminator → parseNdefRecords throws "No NDEF TLV …") as blank, rather than falling back to CC-only OpenPrintTag. A genuine read glitch of a confirmed-CC tag still falls back to OpenPrintTag. So a freshly-erased SLIX2 now reads as blank in Settings. @codex review

hyiger:
Fixed in `69317a8a`. Extracted a refreshDetected callback and call it after a successful erase + read-only toggle, so the Settings NFC Tools loaded-tag label + lock badge update immediately (those mutate the tag in place with no tagPresent/tagUid change, so the detect effect wouldn't re-run on its own). @codex review

PR #933, opened 2026-06-29
fix(openprinttag): fix OPT browse timeout on Win

dcode:
The OpenPrintTag DB browse could hang after download with partial repo copies accumulating in %TEMP% and a misleading "connection timed out". On Windows, unpacking the ~11k tiny YAML files is slow (AV real-time-scans each write + NTFS overhead), so a cold extract blew past 60s and aborted the still-open fetch even though the ~3 MB download had finished in ~600ms. This would cause repeated "extract timed out (60s limit)" failures and an empty OpenPrintTag browser.

Root causes, all in src/lib/openprinttagBrowser.ts:

- Cleanup used synchronous rmSync(recursive) on the ~11k-file extracted repo, blocking the embedded server's event loop, and force:true does not retry on Windows EBUSY/EPERM/ENOTEMPTY, so a locked delete threw and was silently swallowed -> partial dir left behind.
- Each of 3 retry attempts minted a new temp dir with no sweep of leftovers -> several partial copies.
- The 60s AbortSignal is armed on the whole body stream, so an extract/parse stall surfaced as a download-shaped timeout.

Changes:
- Add [openprinttag]-prefixed phase logging (download/extract/parse/ cleanup, with elapsed + counts). In the packaged app these flow to %APPDATA%/Filament DB/logs/main.log via the existing stdout mirror, so there is finally something to see when diagnosing a stall.
- Replace rmSync with an awaited async rm helper (maxRetries/retryDelay) at both call sites; log failures instead of swallowing.
- Remove stale openprinttag-* temp dirs (>1h old) at the start of a run.
- Re-label timeout/abort errors with the phase they struck so the surfaced message is honest.
- Buffer the tarball into memory first (bounded by existing download timeout + a 128 MB size cap), then extract from the memory buffer under a more generous timeout (180s).

Tests:
- Cleanup leaves no net temp dirs
- Stale cleanup removes >1h dirs but spares fresh ones
- Download-phase timeout surfaces a download-specific message.
- Download goes into buffer and then successfully extracts.

Fixes: #932

### DIFF
--- a/src/lib/openprinttagBrowser.ts
+++ b/src/lib/openprinttagBrowser.ts
@@ -583,6 +583,15 @@ async function runFetchWithRetries(): Promise<OPTDatabase> {
 // only once. Worst case: MAX_ATTEMPTS × DOWNLOAD_TIMEOUT_MS + backoff(3.2s) +
 // EXTRACT_TIMEOUT_MS = 3×45 + 3.2 + 120 ≈ 258s, comfortably under the client's
 // 300s (5 min) abort.
+//
+// Strictly: EXTRACT_TIMEOUT_MS bounds the gunzip→counter→tar.x PIPELINE only.
+// `clearTimeout(extractTimer)` fires right after the pipeline resolves, so the
+// YAML parse loop that follows has NO deadline. That's fine in practice —
+// parse is CPU-bound, yields to the event loop every 256 files, and runs once
+// per cold load — but a pathologically slow parse can still eat into the ~42s
+// margin between the 258s server window and the 300s client abort. If parse
+// ever takes long enough to matter, give it its own per-file or whole-loop
+// budget rather than widening EXTRACT_TIMEOUT_MS, which only governs unpack.
 const MAX_ATTEMPTS = 3;
 const DOWNLOAD_TIMEOUT_MS = 45_000;
 const EXTRACT_TIMEOUT_MS = 120_000;
@@ -591,6 +600,12 @@ const EXTRACT_TIMEOUT_MS = 120_000;
 // tarball is ~3 MB; 128 MB is generous headroom. (The decompressed-size /
 // file-count tar-bomb guard still runs during extraction below.)
 const MAX_DOWNLOAD_BYTES = 128 * 1024 * 1024;
+// Cap the DECOMPRESSED stream during extract. A counting Transform between
+// gunzip and tar.x trips this against bytes actually streamed (not the lyable
+// header `size`). Injectable into extractAndParse so the trip can be unit-
+// tested without allocating 256 MB; production uses the default.
+const MAX_TARBALL_EXTRACT_BYTES = 256 * 1024 * 1024;
+const MAX_TARBALL_FILES = 50_000;
 
 /**
  * Retry the network download (exponential backoff) and return the compressed
@@ -730,9 +745,10 @@ async function extractParseOnce(tarballBuffer: Buffer): Promise<OPTDatabase> {
   }
 }
 
-async function extractAndParse(
+export async function extractAndParse(
   tarballBuffer: Buffer,
   tmpDir: string,
+  opts?: { maxExtractBytes?: number },
 ): Promise<OPTDatabase> {
   try {
     // GH #258: bound the extraction so a hostile/compromised tarball (tar
@@ -748,9 +764,11 @@ async function extractAndParse(
     // EXTRACT_TIMEOUT_MS budget. Aborting the pipeline rejects with an
     // AbortError, which relabelTimeoutError recognises so the surfaced message
     // names the extract phase honestly.
+    //
+    // `maxExtractBytes` is injectable so the decompressed-size trip can be
+    // unit-tested with a tiny cap — production uses MAX_TARBALL_EXTRACT_BYTES.
+    const maxExtractBytes = opts?.maxExtractBytes ?? MAX_TARBALL_EXTRACT_BYTES;
     const extractStart = Date.now();
-    const MAX_TARBALL_EXTRACT_BYTES = 256 * 1024 * 1024;
-    const MAX_TARBALL_FILES = 50_000;
     let decompressedBytes = 0;
     let fileCount = 0;
     const extractController = new AbortController();
@@ -761,7 +779,7 @@ async function extractAndParse(
     const byteCounter = new Transform({
       transform(chunk: Buffer, _enc, cb) {
         decompressedBytes += chunk.length;
-        if (decompressedBytes > MAX_TARBALL_EXTRACT_BYTES) {
+        if (decompressedBytes > maxExtractBytes) {
           cb(
             new Error(
               "OpenPrintTag tarball exceeds extraction limits (possible tar bomb).",
--- a/src/lib/openprinttagBrowser.ts
+++ b/src/lib/openprinttagBrowser.ts
@@ -13,9 +13,9 @@
  */
 
 import { parse as parseYaml } from "yaml";
-import { mkdtempSync, rmSync } from "fs";
-import { readFile, readdir } from "fs/promises";
-import { join } from "path";
+import { mkdtempSync } from "fs";
+import { readFile, readdir, rm, stat } from "fs/promises";
+import { basename, join } from "path";
 import { tmpdir } from "os";
 import { Readable } from "node:stream";
 import { pipeline } from "node:stream/promises";
@@ -424,6 +424,81 @@ async function walkDir(dir: string): Promise<string[]> {
   return results;
 }
 
+// Prefix on every diagnostic line so the OPT fetch flow is greppable in the
+// packaged app's main.log (electron/main.ts mirrors the embedded server's
+// stdout/stderr to %APPDATA%/Filament DB/logs/main.log).
+const LOG = "[openprinttag]";
+
+// All temp dirs this module creates share this prefix; the stale-dir sweep
+// keys off it to reclaim partial copies left by an earlier interrupted run.
+const TMP_PREFIX = "openprinttag-";
+
+/**
+ * Remove a temp dir robustly. Uses async `rm` (yields the event loop instead
+ * of blocking it on a ~11k-file recursive delete — the reported "hang") with
+ * Node's built-in retry, which backs off on the Windows file-lock errors
+ * (EBUSY/EPERM/ENOTEMPTY) that antivirus / lingering handles trigger and that
+ * a plain `force: true` does NOT retry — the cause of the partial leftovers.
+ * Best-effort: logs the failure rather than throwing (cleanup must not mask
+ * the real result), but does NOT swallow it silently.
+ */
+async function cleanupTempDir(dir: string): Promise<void> {
+  try {
+    await rm(dir, {
+      recursive: true,
+      force: true,
+      maxRetries: 5,
+      retryDelay: 100,
+    });
+  } catch (err) {
+    console.warn(
+      `${LOG} cleanup failed for ${dir} — may leave a partial copy:`,
+      err instanceof Error ? err.message : err,
+    );
+  }
+}
+
+/**
+ * Best-effort sweep of stale `openprinttag-*` temp dirs from earlier runs that
+ * failed to clean up (e.g. an interrupted extract on Windows). Only removes
+ * dirs older than STALE_AGE_MS so it can't race a concurrent in-progress fetch
+ * (another tab/instance sharing %TEMP%). Never throws.
+ */
+async function sweepStaleTempDirs(): Promise<void> {
+  const STALE_AGE_MS = 60 * 60 * 1000; // 1 hour
+  const root = tmpdir();
+  let reclaimed = 0;
+  try {
+    for (const name of await readdir(root)) {
+      if (!name.startsWith(TMP_PREFIX)) continue;
+      const full = join(root, name);
+      try {
+        const info = await stat(full);
+        if (!info.isDirectory()) continue;
+        if (Date.now() - info.mtimeMs < STALE_AGE_MS) continue;
+        await rm(full, {
+          recursive: true,
+          force: true,
+          maxRetries: 5,
+          retryDelay: 100,
+        });
+        reclaimed += 1;
+      } catch {
+        // per-entry best-effort — a locked dir is reclaimed on a later sweep
+      }
+    }
+  } catch (err) {
+    console.warn(
+      `${LOG} stale temp-dir sweep failed:`,
+      err instanceof Error ? err.message : err,
+    );
+    return;
+  }
+  if (reclaimed > 0) {
+    console.log(`${LOG} swept ${reclaimed} stale temp dir(s) from ${root}`);
+  }
+}
+
 /**
  * Fetch the OpenPrintTag database from GitHub, parse all YAML files,
  * and return the structured result.
@@ -465,21 +540,31 @@ async function runFetchWithRetries(): Promise<OPTDatabase> {
   const MAX_ATTEMPTS = 3;
   let lastError: unknown = null;
 
+  // Reclaim any partial copies an earlier interrupted run left behind before
+  // we start adding more.
+  await sweepStaleTempDirs();
+
   for (let attempt = 1; attempt <= MAX_ATTEMPTS; attempt++) {
-    const tmpDir = mkdtempSync(join(tmpdir(), "openprinttag-"));
+    const tmpDir = mkdtempSync(join(tmpdir(), TMP_PREFIX));
+    console.log(
+      `${LOG} attempt ${attempt}/${MAX_ATTEMPTS} starting (tmpDir=${basename(tmpDir)})`,
+    );
 
     try {
       const result = await fetchAndParse(tmpDir);
+      console.log(
+        `${LOG} attempt ${attempt} succeeded — ${result.totalFFF} FFF / ${result.totalSLA} SLA materials`,
+      );
       return result;
     } catch (err) {
       lastError = err;
+      console.error(
+        `${LOG} attempt ${attempt}/${MAX_ATTEMPTS} failed:`,
+        err instanceof Error ? err.message : err,
+      );
       // Clean up the tmp dir from the failed attempt before retrying so
       // disk usage doesn't grow on a flaky network.
-      try {
-        rmSync(tmpDir, { recursive: true, force: true });
-      } catch {
-        // best-effort
-      }
+      await cleanupTempDir(tmpDir);
       if (attempt < MAX_ATTEMPTS) {
         // Exponential backoff: 800ms, 2400ms. Total worst-case wait
         // ~3.2s extra over the base 60s per attempt — still well under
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

hyiger (dismissed):
Thanks for the thorough diagnosis and write-up — this is a genuinely good fix. Decoupling the download (its own `AbortSignal.timeout`) from the extract (a separate `AbortController`/timer cleared in a `finally` around the pipeline) is exactly the right structure, the honest per-phase error messages are a real improvement over the old "connection timed out" misattribution, and the async `rm` with `{maxRetries:5, retryDelay:100}` correctly targets the Windows `EBUSY/EPERM/ENOTEMPTY` lock errors that `force:true` alone won't retry — the actual root cause of the leftover `%TEMP%` dirs.

I reviewed control-flow, the new in-memory buffering/DoS surface, the Windows/async behavior, and the tests. **No correctness blockers** — control flow, the dual-deadline phase labeling, timer cleanup, and temp-dir cleanup are all leak-free and correct. The requested changes below are test coverage on the two highest-value *new* branches plus a cheap memory tweak; once those land I'm happy to merge.

## Requested changes

1. **Cover the extract-phase timeout relabel — the core of this PR.** The whole point is that an extract/parse stall used to surface as a *download* timeout. But the only timeout test throws a `TimeoutError` from `fetch`, which exercises solely the `phase === "download"` branch — nothing drives the `extractController` abort or feeds an `AbortError` through `isTimeoutAbort`. Please add coverage for the `extract` branch + its `"…the download completed but the extract phase exceeded the deadline"` message. Cheapest path: `export` `isTimeoutAbort` and unit-test it over both `AbortError` and `TimeoutError`; or make `EXTRACT_TIMEOUT_MS` injectable and stall the pipeline.

2. **Cover the `MAX_DOWNLOAD_BYTES` (128 MB) cap.** This is a new OOM/DoS guard with no test. Mock `fetch` to enqueue chunks past the cap (or make the limit injectable) and assert the `"exceeds size limit"` rejection. The check ordering is correct (it throws before `chunks.push`, so the over-limit chunk isn't retained) — just lock it in.

3. **Isolate the test temp root.** `countTempDirs()` / the sweep tests scan the real `os.tmpdir()` for the literal `openprinttag-` prefix and assert `after === before`. But production runs `sweepStaleTempDirs()` against that same dir, so a leftover `>1h`-old `openprinttag-*` dir from a crashed earlier CI run makes `after < before` and fails the suite for unrelated reasons. Point the suite at its own root (set `TMPDIR`/`TMP`/`TEMP` to a `mkdtempSync` dir in `beforeAll`/`afterAll`), or assert presence/absence of the *specific* dirs each test created rather than a global count.

4. **Avoid the ~2× peak memory of the buffer.** `chunks` is a function-scoped `const` that's never cleared, and `Buffer.concat(chunks)` allocates a second full copy while the first stays GC-rooted through the entire extract (`Readable.from([tarballBuffer])`) and the ~11k-file parse — worst case the 128 MB cap allows is ~256 MB transient on the embedded server. Baseline streamed `response.body` straight into `tar.x`, so this is a genuinely new ceiling. One cheap fix: `Buffer.concat(chunks, downloadedBytes)` then `chunks.length = 0;` right after the concat.

## Nits (optional)

- **Dead `phase === "parse"` relabel branch.** The parse loop has no abort signal, so a relabel there would fabricate a `180s "limit"` for a phase that has none. Drop `"parse"` from the `phase` union (or guard the relabel) so it can't misreport.
- **`const buf = chunk as Buffer`** — `Readable.fromWeb` can yield `Uint8Array`; it's runtime-safe (`Buffer.concat`/`.length` accept it) but the cast is a type lie. `Uint8Array[]` would be honest.
- **Comment drift** on the `isTimeoutAbort` docblock — it still references "the 60s deadline" and "DOMException/TypeError shapes" from before the two-deadline split.
- The tar-bomb disk guard counts the attacker-declared `entry.size` rather than bytes actually written. **Pre-existing** (byte-identical to `main`, untouched here) — and your new 128 MB compressed cap actually adds a bound the baseline lacked — so not a gate, just a worthwhile follow-up someday.

Nice work overall — the structure is right and the diagnosis is spot-on. Mainly want the two new branches (extract-timeout relabel + the size cap) under test before this goes into the OPT browse path.

src/lib/openprinttagBrowser.ts
```diff
@@ -669,16 +835,40 @@ async function fetchAndParse(tmpDir: string): Promise<OPTDatabase> {
...
+    // instead of implying a download/connection failure for an extract/parse
+    // stall. Non-timeout errors keep their own message.
+    if (isTimeoutAbort(err)) {
+      const limitSec = Math.round(
+        (phase === "download" ? DOWNLOAD_TIMEOUT_MS : EXTRACT_TIMEOUT_MS) / 1000,
+      );
+      throw new Error(
+        phase === "download"
+          ? `OpenPrintTag download timed out (${limitSec}s limit)`
+          : `OpenPrintTag ${phase} timed out (${limitSec}s limit) — the download completed but the ${phase} phase exceeded the deadline`,
```
hyiger: **should-fix — the core of this PR is untested.** The whole point is that an extract/parse stall used to surface as a *download* timeout; this relabel is what fixes it. But the only timeout test (`tests/openprinttagBrowser.test.ts:998`) throws a `TimeoutError` from `fetch`, which exercises solely the `phase === "download"` branch — nothing drives the `extractController` abort or feeds an `AbortError` through `isTimeoutAbort`, so this `extract`-phase message has zero coverage. Cheapest fix: `export isTimeoutAbort` and unit-test it over both `AbortError` and `TimeoutError`; or make `EXTRACT_TIMEOUT_MS` injectable and stall the pipeline, then assert `/extract timed out/` + the "download completed" wording.

src/lib/openprinttagBrowser.ts
```diff
@@ -543,6 +651,34 @@ async function fetchAndParse(tmpDir: string): Promise<OPTDatabase> {
...
+    // independent deadline (EXTRACT_TIMEOUT_MS) and the network timeout can no
+    // longer misfire mid-unpack.
+    let downloadedBytes = 0;
+    const chunks: Buffer[] = [];
+    for await (const chunk of Readable.fromWeb(
+      response.body as Parameters<typeof Readable.fromWeb>[0],
+    )) {
+      const buf = chunk as Buffer;
+      downloadedBytes += buf.length;
+      if (downloadedBytes > MAX_DOWNLOAD_BYTES) {
```
hyiger: **should-fix — new OOM/DoS guard with no test.** The ordering here is correct (it throws *before* `chunks.push`, so the over-limit chunk isn't retained), but `MAX_DOWNLOAD_BYTES` is a new security cap and nothing exercises it. Mock `fetch` to enqueue chunks past 128 MB (or make the limit injectable) and assert the "exceeds size limit" rejection, so a future refactor can't silently drop the cap.

src/lib/openprinttagBrowser.ts
```diff
@@ -543,6 +651,34 @@ async function fetchAndParse(tmpDir: string): Promise<OPTDatabase> {
...
+        throw new Error(
+          "OpenPrintTag download exceeds size limit (possible malicious response).",
+        );
+      }
+      chunks.push(buf);
+    }
+    const tarballBuffer = Buffer.concat(chunks);
```
hyiger: **should-fix — ~2× peak memory.** `chunks` is a function-scoped `const` that's never cleared, and `Buffer.concat(chunks)` allocates a second full copy while the first stays GC-rooted through the entire extract (`Readable.from([tarballBuffer])` below) and the ~11k-file parse — so the worst case the 128 MB cap allows is ~256 MB held transiently on the embedded server. Baseline streamed `response.body` straight into `tar.x`, so this is a genuinely new ceiling. Cheap fix:
```js
const tarballBuffer = Buffer.concat(chunks, downloadedBytes);
chunks.length = 0; // release the per-chunk copies before extract/parse
```

tests/openprinttagBrowser.test.ts
```diff
@@ -921,4 +930,103 @@ type: Resin
     expect(globalThis.fetch).toHaveBeenCalledTimes(1); // joined, not duplicated
     expect(a).toBe(b);
   });
+
+  // Helper: count `openprinttag-*` temp dirs currently in tmpdir().
+  function countTempDirs(): number {
```
hyiger, line 978: **should-fix — fragile against the shared real `os.tmpdir()`.** `countTempDirs()` scans the real temp dir for the literal `openprinttag-` prefix and the test asserts `after === before`, but production runs `sweepStaleTempDirs()` against that same dir on every fetch — so a `>1h`-old `openprinttag-*` leftover from a crashed earlier CI run makes `after < before` and fails this suite for unrelated reasons. Point the suite at its own root (set `TMPDIR`/`TMP`/`TEMP` to a `mkdtempSync` dir in `beforeAll`/`afterAll`), or assert presence/absence of the *specific* dirs each test created instead of a global count.

src/lib/openprinttagBrowser.ts
```diff
@@ -504,11 +589,31 @@ async function runFetchWithRetries(): Promise<OPTDatabase> {
...
 /**
  * Single-attempt fetch + tarball extract + parse. Factored out so the
  * retry loop above can call it multiple times against fresh temp dirs.
  */
 async function fetchAndParse(tmpDir: string): Promise<OPTDatabase> {
+  // Track which phase is running so a timeout/abort (the 60s AbortSignal is
+  // armed on the whole body stream, so it can fire during extract, not just
+  // download) is reported HONESTLY instead of always reading "connection
+  // timed out" — the misleading message users saw when extract/parse stalled.
+  let phase: "download" | "extract" | "parse" = "download";
```
hyiger: **nit — dead `"parse"` branch.** `phase` can be set to `"parse"`, but the parse loop has no abort signal, so the relabel below (`:843-851`) would only ever fabricate a `180s "limit"` for a phase that has no deadline. Drop `"parse"` from the union (or guard the relabel to `extract` only) so a parse-time error can't be misreported as a timeout.

src/lib/openprinttagBrowser.ts
```diff
@@ -543,6 +651,34 @@ async function fetchAndParse(tmpDir: string): Promise<OPTDatabase> {
...
+    // writes) aborted the still-open fetch even though the bytes were already
+    // down. With the bytes fully in hand, extraction runs under its own,
+    // independent deadline (EXTRACT_TIMEOUT_MS) and the network timeout can no
+    // longer misfire mid-unpack.
+    let downloadedBytes = 0;
+    const chunks: Buffer[] = [];
+    for await (const chunk of Readable.fromWeb(
+      response.body as Parameters<typeof Readable.fromWeb>[0],
+    )) {
+      const buf = chunk as Buffer;
```
hyiger: **nit — type lie.** `Readable.fromWeb` can yield `Uint8Array`, not only `Buffer`. It's runtime-safe (`Buffer.concat` and `.length` both accept `Uint8Array`, and the test mock enqueues `Uint8Array`), but `chunk as Buffer` over-claims — `const chunks: Uint8Array[]` + dropping the cast would be honest.

src/lib/openprinttagBrowser.ts
```diff
@@ -669,16 +835,40 @@ async function fetchAndParse(tmpDir: string): Promise<OPTDatabase> {
...
+    throw err;
+  } finally {
+    // Clean up the temp directory. Async + retrying so it neither blocks the
+    // event loop nor leaves a partial copy on a Windows file lock.
+    await cleanupTempDir(tmpDir);
   }
 }
 
+/**
+ * True for the DOMException/TypeError shapes Node's fetch + AbortSignal.timeout
```
hyiger: **nit — stale comment.** This docblock still references "the 60s deadline" and the pre-split single-signal framing; after the download/extract split there are two deadlines (60s / 180s) and this helper now also catches the extract `AbortController` abort. Worth a quick refresh so it doesn't mislead the next reader.

src/lib/openprinttagBrowser.ts
```diff
@@ -552,37 +688,59 @@ async function fetchAndParse(tmpDir: string): Promise<OPTDatabase> {
...
+          cwd: tmpDir,
+          filter: (entryPath: string, entry: { size?: number }) => {
+            if (
+              entryPath.startsWith("/") ||
+              entryPath.split(/[\\/]/).includes("..")
+            ) {
+              throw new Error(`Unsafe tarball entry path: ${entryPath}`);
+            }
+            fileCount += 1;
+            extractedBytes += entry.size ?? 0;
```
hyiger: nit (pre-existing, **not a gate**): the tar-bomb disk guard sums the attacker-declared `entry.size` from the header rather than bytes actually written, so a lying header could under-declare while streaming more to disk. This is byte-identical to `main` (untouched by this PR), and your new 128 MB *compressed* cap actually adds a bound the baseline lacked — flagging only as a worthwhile someday follow-up.

hyiger (changes requested):
Reviewed control-flow, the new in-memory buffering/DoS surface, the Windows/async behavior, and the tests (builds + lints clean, all 62 tests pass). **No correctness blockers** — decoupling the download deadline from the extract deadline is exactly the right structure, the async `rm` retry correctly targets the Windows `EBUSY/EPERM/ENOTEMPTY` lock errors that `force:true` won't retry, and the size-cap ordering + stale-dir sweep window are sound. Requesting changes for the inline items: the two highest-value *new* branches (the extract-phase timeout relabel and the 128 MB download cap) ship untested, plus a test-isolation fragility and a cheap memory tweak. Nits are optional. Nice diagnosis and write-up overall.

src/lib/openprinttagBrowser.ts
```diff
@@ -504,11 +589,31 @@ async function runFetchWithRetries(): Promise<OPTDatabase> {
...
 
+// Download is network-bound and fast (the compressed tarball is ~3 MB and
+// completes in well under a second on a healthy connection). The extract,
+// by contrast, writes ~11k tiny YAML files to disk — and on Windows hosts
+// real-time antivirus (Defender) scans every one of those writes, which can
+// push a cold extract past a minute even when the download itself took 600ms.
+// So the two phases get SEPARATE, independently-armed deadlines: a tight one
+// for the network, a generous one for the disk-bound unpack.
+const DOWNLOAD_TIMEOUT_MS = 60_000;
+const EXTRACT_TIMEOUT_MS = 180_000;
```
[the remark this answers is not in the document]
dcode: Resolved by increasing UI timeout to 300s and reduced server side timeout to max of 258s.

tests/openprinttagBrowser.test.ts
dcode: Resolved by capturing any values of `TMPDIR`, `TMP`, and `TEMP` env vars and overriding them for the test before restoring them in `afterAll()`.

src/lib/openprinttagBrowser.ts
dcode: Reduced phase to only `download` or `extract`.

src/lib/openprinttagBrowser.ts
dcode: type changed as suggested. Good callout.

src/lib/openprinttagBrowser.ts
dcode: docblock updated to reflect changes.

src/lib/openprinttagBrowser.ts
dcode: Fixed by gunziping first, then using a `Transform` to count the decompressed bytes, throwing an error if it exceeds `MAX_TARBALL_EXTRACT_BYTES` (256 MB).

src/lib/openprinttagBrowser.ts
dcode: Done.

src/lib/openprinttagBrowser.ts
dcode: Made limit injectable and is now exercised by test `"rejects a download that exceeds the size cap (OOM/DoS guard)"` by setting `maxBytes` to 150 (no point in exceeding 256MB of random memory in CI for no reason).

src/lib/openprinttagBrowser.ts
dcode: Added tests to exercise download timeout, exported helper `isTimeoutAbort` and added test to exercise it, and added exported helper `relabelTimeoutError` to provide a correct message indicating whether the error was due to timed out download or timed out extraction and provides the respective timeout value to provide clarity to the user of what happened.

src/lib/openprinttagBrowser.ts
```diff
@@ -543,46 +679,136 @@ async function fetchAndParse(tmpDir: string): Promise<OPTDatabase> {
...
-    // — or a flood of small entries — trips the limit before the data
-    // is written. It also rejects absolute paths and `..` traversal as
-    // defence-in-depth over tar's own path sanitisation.
+    // The extract gets its OWN AbortController/timeout, independent of the
+    // network deadline — a slow disk (Windows + antivirus) gets the full
+    // EXTRACT_TIMEOUT_MS budget. Aborting the pipeline rejects with an
+    // AbortError, which relabelTimeoutError recognises so the surfaced message
+    // names the extract phase honestly.
+    const extractStart = Date.now();
     const MAX_TARBALL_EXTRACT_BYTES = 256 * 1024 * 1024;
```
hyiger: nit / optional follow-up — this is the one new guard without a direct test. You made `MAX_DOWNLOAD_BYTES` injectable + tested it trip; the symmetric move here would be to give `extractAndParse` an optional `maxExtractBytes` (defaulting to this const) and add a test that a tarball decompressing past a tiny injected cap rejects with the tar-bomb error. The mechanism is correct (I verified the gunzip teardown bounds it), and the happy path runs through it in the existing extract tests — so this is parity/regression-safety, not a correctness gap. Not blocking.

src/lib/openprinttagBrowser.ts
```diff
@@ -462,76 +542,132 @@ export async function fetchOpenPrintTagDatabase(): Promise<OPTDatabase> {
...
+
+// Budget math (kept in lockstep with the client abort in
+// src/app/openprinttag/page.tsx). The download is network-bound and fast (the
+// compressed tarball is ~3 MB), so it gets a tight deadline and is retried for
+// transient blips. The extract writes ~11k tiny YAML files to disk — and on
+// Windows hosts real-time antivirus (Defender) scans every one of those
+// writes, which can push a cold extract past a minute even when the download
+// itself took 600ms — so it gets a separate, more generous deadline but runs
+// only once. Worst case: MAX_ATTEMPTS × DOWNLOAD_TIMEOUT_MS + backoff(3.2s) +
+// EXTRACT_TIMEOUT_MS = 3×45 + 3.2 + 120 ≈ 258s, comfortably under the client's
```
hyiger, line 584: nit — the worst-case math (and the matching `page.tsx` comment) says "extract/parse ONCE (≤120s)", but `EXTRACT_TIMEOUT_MS` only bounds the extract **pipeline**: `clearTimeout(extractTimer)` fires right after it (`:807`), so the YAML parse loop that follows has no deadline. In practice it's fine — parse is CPU-bound, yields every 256 files, and the 300s client abort leaves ~42s of margin over the 258s — but strictly the bound covers extract only, not parse. Worth a word in the comment so the next reader isn't surprised that a pathologically slow parse could still exceed the budget.

hyiger (approved):
Re-reviewed the revision — **all 8 prior findings are genuinely resolved**, several handled better than I asked. Verified: tsc + eslint clean, all 71 tests pass (was 62), and I adversarially checked the riskiest new piece (the gzip-bomb bound).

**Resolved**
1. ✅ Extract-phase relabel — extracted to pure, exported `relabelTimeoutError` + `isTimeoutAbort`, tested for both phases incl. injected deadlines.
2. ✅ `MAX_DOWNLOAD_BYTES` cap — `downloadTarballToBuffer({maxBytes})` is exported + tested (over-cap rejects, under-cap returns, ordering checks before retain).
3. ✅ Test temp-root isolation — `TMPDIR`/`TMP`/`TEMP` repointed in `beforeAll`/`afterAll` (nice — `os.tmpdir()` reads env at call time, so it redirects both the module and the helpers).
4. ✅ ~2× memory — `Buffer.concat(chunks, downloadedBytes)` + `chunks.length = 0`.
5. ✅ Dead `"parse"` relabel branch removed.
6. ✅ `chunks: Uint8Array[]`, cast dropped.
7. ✅ Docblock refreshed.
8. ✅ Tar-bomb guard — went beyond the nit: replaced the lyable header `entry.size` with a counting `Transform` between `createGunzip()` and `tar.x({gzip:false})` that bounds **actual decompressed bytes**. I confirmed `pipeline()` tears down the upstream gunzip when the counter errors, so a gzip bomb stops at ~256 MB rather than fully decompressing.

The retry-budget rework (retry download only, extract/parse once; client abort 240s→300s to clear the ~258s worst case) is sound. **Approving** — only two optional follow-ups inline, neither blocking. Excellent, thorough response to the review.

hyiger:
@codex review

src/lib/openprinttagBrowser.ts
hyiger: Resolved in 275d742. Hoisted `MAX_TARBALL_EXTRACT_BYTES` (+ `MAX_TARBALL_FILES`) to module scope, added an injectable `extractAndParse(buf, tmpDir, { maxExtractBytes })`, exported the function, and added two tests in `tests/openprinttagBrowser.test.ts`: one trips the cap with a 100-byte injected limit against a ~500-byte payload and asserts `/exceeds extraction limits/`; the symmetric happy-path test asserts a normal tarball parses under a generous cap. Production still uses the 256 MB default — the const is just the fallback now.

src/lib/openprinttagBrowser.ts
hyiger: Resolved in 275d742. Updated the budget-math comment (both `openprinttagBrowser.ts`:584 and the matching client-abort comment in `src/app/openprinttag/page.tsx`) to spell out that `EXTRACT_TIMEOUT_MS` bounds the gunzip→counter→tar.x pipeline only — the YAML parse loop that follows has no deadline, runs once per cold load, yields every 256 files, and the 300s client abort leaves ~42s of margin for it. Includes the suggested guidance for the next reader (if parse ever takes long enough to matter, give it its own per-file/whole-loop budget rather than widening `EXTRACT_TIMEOUT_MS`).

PR #936, opened 2026-06-30
feat(analytics): stack Usage-by-day chart by filament

hyiger:
Fixes #934.

## Summary

Adds an opt-in **Detailed** mode to the analytics Usage-by-day chart that stacks each bar by filament, with each segment painted in the filament's actual hex color.

- **Server (`/api/analytics`)**: `usageByDay[i]` now carries `byFilament[]` (`{id, name, color, grams}`), sorted DESC by grams so the chart renders largest-at-the-bottom without re-sorting. Color resolves via the existing variant→parent inheritance pattern through `displayColor()`, so a variant inherits its parent's color and a coextruded filament (null primary) falls through to `secondaryColors[0]`. The day's top-level `grams` is the sum of the rounded segment grams to eliminate sum-then-round vs round-then-sum drift between the stack and the Y-axis math.
- **Client (`/analytics`)**: a small **Detailed** toggle next to the section heading, persisted in `localStorage` under `filamentdb-analytics-usage-detailed`. **Default off — no behavior change for existing users**. Detailed-on renders each bar as a vertical stack of `<div>` segments with `style.backgroundColor` driven by each filament's hex; each segment carries a native `title=` + `aria-label=` tooltip (`{name}: {grams} g`). A small top-10 legend with "+N more" appears below the chart in Detailed mode only.
- **i18n**: 4 new keys (EN + DE in parity) under `analytics.usageByDay.*` — `detailed`, `detailed.tooltip`, `tooltipFormat`, `moreCount`.

## Test plan

- [x] `npx eslint src/app/api/analytics/route.ts src/app/analytics/page.tsx` clean
- [x] `jq empty` valid on both locale files; key counts equal (1622 each)
- [x] Full `npx vitest run` → 2507 / 2507 pass across 140 files
- [x] 4 new route cases in `tests/analytics-route.test.ts` covering single-filament day, multi-filament day (sort DESC + sum = grams), variant color inheritance from parent, and coextruded null-primary fallback via `displayColor()`
- [x] `tests/i18n-key-coverage.test.ts` still green (EN+DE parity invariant)

## Notes

- **Opt-in via the toggle** — default off means existing users see no change to the chart on first load after upgrading.
- **No new chart library** — the stack is built from the same hand-rolled `<div>` bars as the existing solid bar; the empty-state guard (`every(d => d.grams === 0)`) still operates on the day's total `grams`, which is unchanged in shape.


src/app/analytics/page.tsx
```diff
@@ -31,6 +48,30 @@ export default function AnalyticsPage() {
...
+  // survives a reload; default is off so existing behaviour is unchanged
+  // for users who haven't opted in.
+  const [detailed, setDetailed] = useState<boolean>(() => {
+    if (typeof window === "undefined") return false;
+    try {
+      return window.localStorage.getItem(DETAILED_STORAGE_KEY) === "1";
+    } catch {
+      return false;
+    }
+  });
```
hyiger: **[P2 — correctness] Hydration mismatch on the `detailed` toggle.**

The lazy `useState` initialiser reads `localStorage` synchronously, but `"use client"` components are still server-rendered by the App Router. SSR emits HTML with `detailed=false`; a returning user's first client render computes `detailed=true`, flipping `aria-pressed` and the className branch (`bg-blue-600 text-white border-blue-600` vs gray border, ~line 270) on the toggle button at hydration — React hydration warning + one-frame flash.

The codebase already documents the fix at `src/app/inventory/page.tsx:186` — initialise with the default, load the persisted value in a `useEffect`, guard post-mount writes with a `prefsLoaded` ref so the effect doesn't immediately re-persist the default. Apply the same pattern here.

src/app/analytics/page.tsx
```diff
@@ -222,11 +311,53 @@ export default function AnalyticsPage() {
...
+                                  const segLabel = t(
+                                    "analytics.usageByDay.tooltipFormat",
+                                    { name: seg.name, grams: formatGrams(seg.grams) },
+                                  );
+                                  return (
+                                    <div
+                                      key={seg.id}
+                                      title={`${d.date} — ${segLabel}`}
+                                      aria-label={`${d.date} — ${segLabel}`}
+                                      tabIndex={0}
```
hyiger: **[P2 — a11y + perf] Every stack segment is `tabIndex={0}` — turns the chart into a tab-stop minefield.**

Every segment renders as `<div tabIndex={0} aria-label=… />` with no `role`, no handler, no focus ring. With `DAY_OPTIONS` including 365 and several filaments per day, this is hundreds-to-thousands of sequential tab stops between the Detailed toggle and the legend — keyboard users can't reasonably get past the chart, and the non-detailed branch at ~line 357 correctly has zero. NVDA/JAWS browse mode also won't reliably expose `aria-label` on a roleless focusable `<div>`.

Drop `tabIndex={0}` — the wrapper at ~line 322 already carries the per-day aria-label. If you want SR access to the per-segment breakdown, render a visually-hidden `<table>` summary alongside.

```suggestion
              <div
                key={seg.id}
                title={`${d.date} — ${segLabel}`}
                aria-label={`${d.date} — ${segLabel}`}
                style={{
                  height: `${segPct}%`,
                  backgroundColor: seg.color,
                }}
              />
```

src/app/analytics/page.tsx
```diff
@@ -222,11 +311,53 @@ export default function AnalyticsPage() {
...
+                                      style={{
+                                        height: `${segPct}%`,
+                                        backgroundColor: seg.color,
```
hyiger, line 388: **[P2 — UX] No contrast strategy on stack segments.**

Segments paint `backgroundColor: seg.color` directly over the card, which has no background fill — pure-white filaments vanish in light mode, pure-black in dark, and the `displayColor()` `#808080` fallback collides with the page's mid-gray chrome. The legend chip at ~line 387 already solves this with `border border-gray-300 dark:border-gray-700`; mirror that 1px border on each segment so it survives every theme/hex combination.

hyiger (commented):
Adversarial multi-dimension review — correctness / a11y / perf / tests. After verification: **11 findings** (3 redundant tabIndex finds collapsed). Top three to fix before merge: hydration mismatch on the Detailed toggle (P2), tabIndex tab-stop minefield (P2), and the PrintHistory populate-path test gap (P2).

_Generated with [Claude Code](https://claude.com/claude-code)._

src/app/analytics/page.tsx
```diff
@@ -222,11 +311,53 @@ export default function AnalyticsPage() {
...
+                                  const segPct =
+                                    d.grams > 0 ? (seg.grams / d.grams) * 100 : 0;
+                                  const segLabel = t(
+                                    "analytics.usageByDay.tooltipFormat",
+                                    { name: seg.name, grams: formatGrams(seg.grams) },
+                                  );
+                                  return (
+                                    <div
+                                      key={seg.id}
+                                      title={`${d.date} — ${segLabel}`}
```
hyiger: **[P3 — i18n] Hardcoded en-dash + raw ISO date in segment tooltip bypass i18n.**

`title={`${d.date} — ${segLabel}`}` and the matching `aria-label` leak both the en-dash separator and the raw `YYYY-MM-DD` prefix past `analytics.usageByDay.tooltipFormat` (which only covers `{name}: {grams} g`). Add `analytics.usageByDay.segmentTooltipFormat` (e.g. `"{date} — {label}"`) and route `d.date` through `formatDate(new Date(d.date + 'T00:00Z'), locale)` so German users get a locale-appropriate format. The pre-existing `dayLabel` at ~line 314 has the same issue but predates this PR.

src/app/api/analytics/route.ts
```diff
@@ -59,15 +67,28 @@ export async function GET(request: NextRequest) {
...
     const parentCostMap = new Map<string, number | null>();
+    // GH #934: parent color map mirrors the cost lookup so a variant that
+    // leaves `color`/`secondaryColors` blank to inherit gets its parent's
+    // palette resolved here, the same way `cost` already did.
+    const parentColorMap = new Map<
+      string,
+      { color: string | null; secondaryColors: string[] }
+    >();
     if (parentIdSet.size > 0) {
       const parents = await Filament.find({
```
hyiger, line 110: **[P3 — correctness] Soft-deleted parent breaks variant color inheritance from PrintHistory.**

The parents batch query filters `_deletedAt: null`, but the PrintHistory `populate("usage.filamentId")` does not — so a soft-deleted variant referenced from an in-window job will surface here while its (also soft-deleted) parent is excluded from `parentColorMap`. `resolveColor` falls through to `displayColor(own)` and paints `"#808080"` even though the parent's real color is on disk. The same gap silently zeros `resolveCost` for these rows.

Drop `_deletedAt: null` from the parents query — parents are read-only here, and an in-trash parent's color/cost is still the right answer for a historical job that consumed it.

src/app/api/analytics/route.ts
```diff
@@ -129,20 +179,40 @@ export async function GET(request: NextRequest) {
...
           : null;
         const name = fdoc?.name ?? "(unknown)";
         const vendor = fdoc?.vendor ?? "(unknown)";
         // GH #223: was `fdoc?.cost ?? null` — read the variant's own cost
         // directly and contributed 0 to totalCost for every job against
         // an inheriting variant. resolveCost falls back to the parent.
         const cost = resolveCost(fdoc?.cost ?? null, fdoc?.parentId);
+        // GH #934: resolve color via variant→parent inheritance for the
+        // stacked chart segment.
+        const color = resolveColor(
```
hyiger, line 291: **[P3 — perf] `resolveColor` re-runs per usage row instead of per filament.**

`resolveColor()` is deterministic per `fid`, but is invoked once per usage row in both the PrintHistory loop (here) and the manual loop at ~line 249. For a filament with 200 usage rows in the window, the `Array.isArray` + length + `displayColor` + `parentColorMap.get` runs 200× for one deterministic answer.

Mirror the existing `resolveCost` pattern but cache via a `Map<fid, string>` populated on first encounter — collapses to O(filaments) instead of O(rows). Low priority — analytics isn't hot — but cheap to fix while the code is open.

src/app/api/analytics/route.ts
```diff
@@ -187,14 +263,41 @@ export async function GET(request: NextRequest) {
...
     }
 
+    // GH #934: emit each day with its per-filament breakdown for the
+    // stacked chart. Round each filament segment's grams individually,
+    // then derive the day-level `grams` as the SUM of those rounded
+    // values so the totals line up perfectly (no `0.5 + 0.5` rounding
+    // drift between the stack and the Y-axis math). Sort byFilament
+    // descending so the largest contributor renders at the BOTTOM of
+    // the stack — the client doesn't re-sort.
     const usageByDay = Array.from(byDay.entries())
```
hyiger: **[P3 — correctness] `byDay` accumulator written but never read.**

The emission destructures only `[date]` and recomputes `grams` from `byDayFilament.get(date)` — the `byDay.set(...)` writes at ~lines 165 and 244 are dead. Both maps are co-seeded with the same keys, so iteration via `byDay.entries()` works today, but the writes have asymmetric guards: line 244's `byDay.set` is unguarded while line 267's `byDayFilament` write is gated by `if (dayBucket)`. The `uDate < since` filter has no upper bound, so a future-dated manual entry would write to `byDay` only and surface as `grams: 0, byFilament: []`.

Either drop `byDay` entirely and iterate `byDayFilament.entries()`, or remove the two dead `byDay.set` calls and comment the map as "key-set placeholder".

src/app/api/analytics/route.ts
```diff
@@ -187,14 +263,41 @@ export async function GET(request: NextRequest) {
...
+        const byFil = Array.from(dayBucket.entries())
+          .map(([id, v]) => ({
+            id,
+            name: v.name,
+            color: v.color,
+            grams: Math.round(v.grams),
+          }))
+          .filter((e) => e.grams > 0)
+          .sort((a, b) => b.grams - a.grams);
+        const grams = byFil.reduce((sum, e) => sum + e.grams, 0);
```
hyiger: **[P3 — correctness] `totals.grams` (round-of-sum) can disagree with Σ `usageByDay[].grams` (sum-of-rounded).**

`usageByDay[].grams = byFil.reduce(sum, …)` over `Math.round(v.grams)` segments (sum-of-rounded). `totals.grams = Math.round(totalGrams)` at ~line 318 is round-of-sum off the raw running total. Many same-direction roundings produce a visible drift — chart totals 1004 g, headline reads 1005 g. The same mismatch holds for `byFilamentArr[].grams` (round-of-sum) vs. each filament's per-day rounded segments across `usageByDay`.

Derive `totals.grams = usageByDay.reduce((s, d) => s + d.grams, 0)` and compute `byFilamentArr` grams from the same per-day segments — the chart becomes the single source of truth and the comment's justification holds end-to-end.

tests/analytics-route.test.ts
```diff
@@ -112,3 +112,179 @@ describe("/api/analytics — manualEntries counter (GH #204)", () => {
...
+  let Filament: any;
+
+  beforeEach(async () => {
+    delete mongoose.models.Filament;
+    Filament = (await import("@/models/Filament")).default;
+    delete mongoose.models.PrintHistory;
+    await import("@/models/PrintHistory");
+  });
+
+  it("single-filament day → byFilament has one entry whose grams equals the day total", async () => {
```
hyiger, line 137: **[P2 — tests] PrintHistory populate path has zero color coverage.**

All four new GH #934 tests seed usage via `Filament.create` with `usageHistory[].source: "manual"`, exercising only the manual loop at `route.ts:229–276`. The PrintHistory loop at `route.ts:177–216` — whose color resolution depends on the populate string `.populate("usage.filamentId", "name vendor cost parentId color secondaryColors")` at `route.ts:45` — is the actually-novel surface this PR adds (pre-PR it was `"name vendor cost parentId"`).

A future contributor trimming the select back to match the GH #223 comment one line above would silently regress every PrintHistory-driven segment to `"#808080"` and the suite would still pass. Add a test that creates a `PrintHistory` with `usage[].filamentId` referencing (a) a variant that inherits color from its parent and (b) a coextruded filament with null primary, asserting `usageByDay[].byFilament[].color`, to pin the populate-select shape.

tests/analytics-route.test.ts
```diff
@@ -112,3 +112,179 @@ describe("/api/analytics — manualEntries counter (GH #204)", () => {
...
+    const parent = await Filament.create({
+      name: "Parent PLA",
+      vendor: "V",
+      type: "PLA",
+      color: "#ABCDEF",
+    });
+    await Filament.create({
+      name: "Variant PLA",
+      vendor: "V",
+      type: "PLA",
```
hyiger, line 256: **[P2 — tests] Rounding invariant never actually stressed.**

The "total = sum of segments" assertion uses integer grams (10/50/25), so `sum === day.grams` is trivially true regardless of strategy — a regression that computed the day total as `Math.round(totalGramsForDay)` would still pass. The route comment specifically justifies the sum-of-rounded-segments design by citing `0.5 + 0.5`-style drift, but no test here feeds non-integer grams.

Add a case with three fractional entries on the same day (e.g. 0.4 / 0.4 / 0.4) — sum-of-rounded yields 3, round-of-sum yields 1 — pinning the invariant the comment claims to protect.

tests/analytics-route.test.ts
```diff
@@ -112,3 +112,179 @@ describe("/api/analytics — manualEntries counter (GH #204)", () => {
...
+    const day = body.usageByDay.find(
+      (d: { grams: number }) => d.grams > 0,
+    );
+    expect(day).toBeDefined();
+    expect(day.byFilament).toHaveLength(1);
+    expect(day.byFilament[0].name).toBe("Variant PLA");
+    expect(day.byFilament[0].color).toBe("#ABCDEF");
+  });
+
+  it("coextruded filament (null primary) resolves to secondaryColors[0] via displayColor()", async () => {
```
hyiger, line 280: **[P2 — tests] Coextruded test only exercises the own-`secondaryColors` path; the inherited fallback is unverified.**

The coextruded test stores `secondaryColors` directly on the filament with the usage entry, so it hits the `ownHasSecondary → displayColor(own)` early return at `route.ts:115` and never enters the parent path. The other inheritance test gives the parent a primary `color`, so the `parentColorMap.get(...) → displayColor(p)` branch and the `Array.isArray(p.secondaryColors)` defensive read at `route.ts:88–90` have no test where `secondaryColors` is the load-bearing field on the **parent**.

A future contributor trimming `.select("_id cost color secondaryColors")` at `route.ts:82` to drop `secondaryColors` would silently break only the inherited-coextruded case. Add a third test: parent with `color: null, secondaryColors: ["#112233", …]` + variant with `color: null, secondaryColors: []` + the usage entry, asserting the inherited hex.

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #937, opened 2026-06-30
feat(openprinttag): SHA-aware cache refresh — skip tarball when upstream unchanged

### LINKED ISSUE #931: Limit downloads on the OpenPrintTag db
I notice that if I go refresh and I have a stale download of the db it takes forever to download.   However I know that it hasn't been updated since the last time it downloaded (My guess is your caching for x hours.)   I hit this issue in filament-bridge as well.   The time to cache is great, but when you download you stash the sha of the current commit.   Then you only need to download the large file when the sha changes.   Caching local is still good for x hours without a lot of pings to GH, but I find with the SHA it can be days before that file gets update at times.  which is what we want.   Just some feedback, keep up the great product!

hyiger:
## Summary

Fixes #931.

The OpenPrintTag browser's Refresh button used to call `clearCache()` and immediately re-fetch the ~3 MB tarball + re-parse ~11k YAML files, even when nothing upstream had changed. Now, before re-downloading the tarball, we probe GitHub's commits API for the latest SHA on `main` (~1 KB JSON) and compare against the SHA the cache was last parsed from. If unchanged, we slide the TTL forward and serve the cached database — no tarball round-trip, no event-loop hit.

- **Same SHA** → slide TTL, serve cached, no tarball fetch.
- **Different SHA** → fall through to the existing tarball download + extract path.
- **Cold cache** (no baseline) → straight to tarball, no probe.
- **Commits-API failure** → fall through to tarball (fail-open — the cheap probe must never wedge a refresh that was already willing to download the whole tarball).

The cached `OPTDatabase` now carries `sha` (extracted from the tarball directory name on parse, re-stamped with the full 40-char SHA on a successful probe match) and `shaCheckedAt` (ISO timestamp of the last probe). The OPT browser page surfaces both under the title as `Last refreshed from commit abc1234 · checked 3m ago`. Two new EN+DE i18n keys (`openprinttag.lastRefreshedFromCommit`, `openprinttag.checkedAgo`) land in lockstep per the existing parity invariant.

The route now goes through `fetchOpenPrintTagDatabase({ force: true })` instead of `clearCache() + fetch` so the probe has a baseline to compare against. `clearCache` itself is unchanged and stays exported for test setup.

The probe honours the existing `getProxyDispatcher()` so air-gapped / proxied deploys keep working, mirrors the User-Agent + Accept headers of the tarball request, and tags every diagnostic line with the existing `[openprinttag]` log prefix so a future bisect can see "commits SHA unchanged — sliding TTL" vs "commits SHA changed XXX → YYY — fetching tarball" in `main.log`.

## Test plan

- [x] `npx vitest run tests/openprinttagBrowser.test.ts` — 87 tests pass (16 new: 6 for `shasMatch`, 4 for `fetchUpstreamCommitSha`, 4 end-to-end SHA-aware refresh paths covering same-SHA / changed-SHA / probe-failure / cold-start, plus 2 existing tests that exercise the new path implicitly).
- [x] `npx vitest run tests/i18n-key-coverage.test.ts` — both new keys present in en + de.
- [x] Full `npx vitest run` — 2517 tests across 140 files, no regressions.
- [x] `npx tsc --noEmit` — clean.
- [x] `npm run lint` — clean.
- [x] `npm run build` — clean.


src/lib/openprinttagBrowser.ts
```diff
@@ -541,6 +594,79 @@ export async function fetchOpenPrintTagDatabase(): Promise<OPTDatabase> {
...
+ * party-controlled value being prefix-matched. (Git's own short-SHA collision
+ * risk is irrelevant: we're comparing the SAME repo's hashes, not searching
+ * for a commit by prefix.)
+ */
+export function shasMatch(a: string, b: string): boolean {
+  if (!a || !b) return false;
+  const minLen = Math.min(a.length, b.length);
+  // Refuse a degenerate "0"-length match — should never happen in practice
+  // (the regex extracts ≥7), but guard against an empty-string slip-through.
+  if (minLen < 4) return false;
```
hyiger: **[P1 — correctness] `shasMatch`'s 4-char floor is below every real source — collision rate ~1/65k random SHAs.**

The tarball regex at ~line 317 emits `[0-9a-f]{7,40}` and `fetchUpstreamCommitSha` returns GitHub's full 40-char SHA. The only way a 4–6 char value reaches `shasMatch` is a malformed/truncated upstream response — and `fetchUpstreamCommitSha` only rejects `body.sha.length === 0`. A `"abc1"` from a misbehaving proxy or compromised intermediary would prefix-match a cached `"abc1234…"`, slide the TTL forward, and serve stale data instead of fail-opening to the tarball path.

Raise the floor to **7** to match the regex minimum, and add `body.sha.length < 7` to the `fetchUpstreamCommitSha` guard so a degenerate response can't lock the cache.

```suggestion
  if (minLen < 7) return false;
  return a.slice(0, minLen).toLowerCase() === b.slice(0, minLen).toLowerCase();
```

src/lib/openprinttagBrowser.ts
```diff
@@ -541,6 +594,79 @@ export async function fetchOpenPrintTagDatabase(): Promise<OPTDatabase> {
...
+      signal: AbortSignal.timeout(timeoutMs),
+      ...(dispatcher ? { dispatcher } : {}),
+    } as RequestInit & { dispatcher?: Dispatcher });
+    if (!response.ok) {
+      console.warn(
+        `${LOG} commits API returned ${response.status} ${response.statusText}`,
+      );
+      return null;
+    }
+    const body = (await response.json()) as { sha?: unknown };
```
hyiger: **[P2 — security + correctness] Probe has no response-size cap before `.json()`.**

`fetchUpstreamCommitSha` calls `response.json()` with no size cap, breaking the #258/#260 hardening pattern the sibling `downloadTarballToBuffer` follows in the same file (streamed cap at `MAX_DOWNLOAD_BYTES = 128 MB`, comment: "a hostile/huge response can't OOM the embedded server"). The probe honours `getProxyDispatcher` (user-controlled `HTTP_PROXY` env), so a misconfigured/hostile proxy returning a multi-megabyte body buffers all of it before parsing — the 10s abort is the only backstop, which still admits hundreds of MB on a fast link.

Fix: read with `readBodyCapped(response, 64 * 1024)` from `src/lib/externalUrlGuard.ts` then `JSON.parse`. The expected GitHub commits payload is ~1 KB, so 64 KB is generous headroom and matches the fail-open contract every caller already handles.

src/lib/openprinttagBrowser.ts
```diff
@@ -519,12 +529,55 @@ async function sweepStaleTempDirs(): Promise<void> {
...
+
+  // Cache hit on the natural fast path (TTL not expired, not a forced refresh).
+  // The SHA-aware probe below intentionally does NOT run here — when the TTL
+  // is fresh the user just saw current data, no point hitting GitHub.
+  if (!force && cachedDatabase && Date.now() - cacheTimestamp < CACHE_TTL_MS) {
     return cachedDatabase;
   }
 
+  // #931: SHA-aware probe. Before we re-download the 3 MB tarball + parse
+  // ~11k files, ask GitHub's commits API for the latest SHA on `main` and
```
hyiger: **[P2 — security] SHA probe runs outside the single-flight guard.**

The probe at ~lines 545–578 runs *before* the single-flight guard at ~line 595, so N concurrent `fetchOpenPrintTagDatabase({force:true})` callers each fire their own commits-API request. With `POST /api/openprinttag` now routing every Refresh-button press through `force:true`, a user mashing Refresh across tabs can burn the unauthenticated **60 req/hr** GitHub quota faster than one probe per refresh; the existing `Retry-After`/403 path then 502s back to the UI for the next hour. The tarball fallback is correctly single-flighted, so this is probe-call amplification only — not the full N×tarball blow-up.

Fix: move the probe inside the single-flight block (wrap probe + decision + tarball fetch in `runFetchWithRetries`) so concurrent forced refreshes share one probe result. Also kills the slide-TTL race below.

src/lib/openprinttagBrowser.ts
```diff
@@ -519,12 +529,55 @@ async function sweepStaleTempDirs(): Promise<void> {
...
+      // last actual parse).
+      console.log(
+        `${LOG} commits SHA unchanged (${upstreamSha.slice(0, 7)}) — sliding TTL`,
+      );
+      cacheTimestamp = Date.now();
+      // Stamp the FULL upstream SHA so subsequent compares don't keep losing
+      // precision via the abbreviated tarball-dir SHA.
+      cachedDatabase = {
+        ...cachedDatabase,
+        sha: upstreamSha,
```
hyiger, line 603: **[P3 — correctness] Slide-TTL race can revert `sha` to a stale value while keeping the new content.**

Slide-TTL runs before the `inFlightFetch` guard, so two callers can overlap on `await fetchUpstreamCommitSha()`. Caller A's probe matches the cached SHA; caller B's later probe sees a moved SHA, takes the tarball path, writes a new `cachedDatabase` (new content, new `sha`). A then resumes and does `cachedDatabase = {...cachedDatabase, sha: upstreamSha, shaCheckedAt: now}` — the spread captures B's new content but reverts `sha` to A's now-stale value and re-stamps `shaCheckedAt`. The next probe compares against the wrong baseline and the UI shows wrong provenance until the next force-refresh.

Either extend the single-flight to cover the probe (preferred — kills the amplification finding above too) or only assign when `cachedDatabase` is still the reference the spread sourced from (CAS-style).

src/lib/openprinttagBrowser.ts
```diff
@@ -541,6 +594,79 @@ export async function fetchOpenPrintTagDatabase(): Promise<OPTDatabase> {
...
+/**
+ * #931: Cheap "did upstream change?" probe. Hits GitHub's commits API for the
+ * latest commit on `main` (~1 KB JSON) instead of re-downloading the 3 MB
+ * tarball. Returns the full SHA on success, `null` on any failure — every
+ * caller fail-opens to the tarball path on null, so this never wedges a
+ * refresh; the probe is pure latency savings.
+ *
+ * Honours the same proxy dispatcher as `downloadTarballToBuffer` so air-gapped
+ * / proxied deployments work the same as the tarball path.
+ */
```
hyiger, line 687: **[P3 — perf] Probe's 10s default erodes the documented retry budget.**

`fetchUpstreamCommitSha` defaults to a 10s `AbortSignal.timeout` for a ~1 KB JSON GET. The probe runs on the critical path of every force-refresh and every TTL-expired GET that has a cached SHA, and on failure falls through to the tarball path — so the full 10s is pure tax when GitHub's commits API is unhealthy. The budget comment at ~line 583 (`3×45 + 3.2 + 120 ≈ 258s, under the 300s client abort`) predates this probe; worst case is now ~268s, eroding the 42s margin to ~32s.

Drop the default to **3–5s** (a 1 KB API call that takes 10s means the connection is dead — fail fast and let the tarball path's existing retry backoff handle it) or update the budget comment to include the probe.

tests/openprinttagBrowser.test.ts
```diff
@@ -1231,3 +1233,298 @@ describe("extractAndParse maxExtractBytes cap", () => {
...
+    // Force=true on a clean cache. There's nothing to compare against, so
+    // the SHA probe must NOT run (it has no baseline) and we go directly to
+    // the tarball download.
+    const db = await fetchOpenPrintTagDatabase({ force: true });
+    expect(counters.commitsCalls()).toBe(0);
+    expect(counters.tarballCalls()).toBe(1);
+    expect(db.totalFFF).toBe(1);
+    expect(db.sha).toBe(SHA);
+  });
+});
```
hyiger, line 1675: **[P3 — tests] Same-SHA test doesn't pin that the cached payload object is reused.**

The test asserts `tarballCalls()` stays at 1 but doesn't pin reference identity on the heavy payload arrays. The slide path uses `{...cachedDatabase, sha, shaCheckedAt}` — a shallow spread that shares the `materials`/`brands` array references. A future refactor to `structuredClone()` or `JSON.parse(JSON.stringify(...))` would re-allocate ~11k materials on every probe and the test would still pass.

Add `expect(second.materials).toBe(first.materials)` (or whatever the heaviest array is on `OPTDatabase`) to pin the shallow-spread invariant.

hyiger (commented):
Adversarial multi-dimension review — correctness / security / perf / tests. After verification: **6 findings** (2 redundant size-cap finds collapsed). Top three to fix before merge: **shasMatch 4-char floor** (P1, stale-data risk on a malformed probe response), **probe `.json()` with no size cap** (P2, SSRF/OOM via a proxied response), and **probe outside the single-flight guard** (P2, burns GitHub's 60-req/hr unauthenticated quota under concurrent refresh).

_Generated with [Claude Code](https://claude.com/claude-code)._

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #961, opened 2026-07-03
fix(print-history): undo-aware usageHistory cap so DELETE refunds survive (#954)

hyiger:
## What & why

Issue **#954 finding #6** (dropped from PR #960): `POST /api/print-history` never capped a spool's `usageHistory`, unlike the manual-usage and dry-cycle routes (GH #304), so a looping client could grow a filament document toward the 16 MiB BSON limit.

A naive `slice(-1000)` is **not** a safe fix here. It evicts the **oldest** entry — exactly the one most likely to be a still-live `source:"job"` entry. `DELETE /api/print-history/{id}` refunds spool weight **only when it finds and removes** that entry (the GH #621 idempotency guard: *entry gone ⇒ already refunded ⇒ skip*). Evicting a live job's entry makes its later DELETE silently skip the refund → a **permanent inventory undercount**. This is Codex P2 #1 on the naive PR #960 version.

## Approach (product call: undo-aware trim, shared across both routes)

- **New pure helper `src/lib/capUsageHistory.ts`** (`capUsageHistory` + `MAX_SPOOL_HISTORY`): trims to 1000 but evicts non-undo entries (`manual`/`nfc`) oldest-first, and touches undo-relevant entries (a `jobId`, or the legacy `job`/`slicer` sources DELETE still matches on) only as a **last resort** when the array is *entirely* undo-relevant — a pathological 1000-live-jobs-on-one-spool case a real spool never reaches. Chronological order preserved; a job's just-pushed (newest, undo-relevant) entries always survive. 100% covered.
- **`POST /api/print-history`** trims each touched spool **once** after all usage rows are applied (trimming mid-loop could evict an entry an earlier row of the same job just pushed).
- **`/spools/[spoolId]/usage`** swaps its naive `slice(-1000)` for the same helper — closing the identical latent eviction hole through that path (a manual log could otherwise evict a live job entry too).

## On Codex P2 #2 (rollback)

The standalone-fallback rollback already removes **only this job's** entries via a `jobId` filter over a **fresh reload** — inherently concurrency-safe. The safe move is to **not** introduce a wholesale `usageHistory` snapshot/restore (that's what would clobber concurrent entries), and let trim-evicted entries stay evicted (they're already persisted-gone after save #1). So the rollback and DELETE's #621 idempotency are **left untouched**.

## Tests

- `tests/capUsageHistory.test.ts` — 14 pure-helper cases (undo-first eviction, source recognition, order preservation, last-resort branch, custom/default max, defensive non-array).
- `tests/print-history.test.ts` — end-to-end: an **old** live job entry survives a later job's trim, and deleting the old job **still refunds** its grams (would fail under a naive slice).
- `tests/spool-subroute.test.ts` — manual-usage route rolls off an old manual, never the live job entry.

Full suite green locally: **146 files / 2918 tests**. Lint + `tsc --noEmit` clean.

### DIFF
--- a/src/app/api/filaments/[id]/spools/[spoolId]/usage/route.ts
+++ b/src/app/api/filaments/[id]/spools/[spoolId]/usage/route.ts
@@ -3,12 +3,7 @@ import dbConnect from "@/lib/mongodb";
 import Filament from "@/models/Filament";
 import { errorResponse, errorResponseFromCaught, handleVersionError } from "@/lib/apiErrorHandler";
 import { assertSameOriginRequest } from "@/lib/requestGuard";
-
-/** GH #304: hard cap on a spool's embedded usageHistory array. Far
- * above any realistic per-spool history; exists to stop a client
- * looping POSTs from growing the filament document toward the 16MB
- * BSON limit. Oldest entries roll off once the cap is reached. */
-const MAX_SPOOL_HISTORY = 1000;
+import { capUsageHistory, MAX_SPOOL_HISTORY } from "@/lib/capUsageHistory";
 
 /**
  * POST /api/filaments/{id}/spools/{spoolId}/usage — manually log grams used.
@@ -86,10 +81,15 @@ export async function POST(
       // explicit at every call site.
       jobId: null,
     });
-    // GH #304: roll off the oldest entries once the cap is reached so
-    // the embedded array can't grow the filament document unbounded.
+    // GH #304 / #954 finding #6: roll off the oldest entries once the cap is
+    // reached so the embedded array can't grow the filament document unbounded.
+    // Undo-aware (capUsageHistory) rather than a plain `slice(-N)`: a manual log
+    // must not evict a still-live `source:"job"` entry, whose later
+    // DELETE /api/print-history refund keys off the entry still being present
+    // (GH #621). Manual/nfc entries are evicted first; job/slicer only as a last
+    // resort when the array is entirely undo-relevant.
     if (spool.usageHistory.length > MAX_SPOOL_HISTORY) {
-      spool.usageHistory = spool.usageHistory.slice(-MAX_SPOOL_HISTORY);
+      spool.usageHistory = capUsageHistory(spool.usageHistory, MAX_SPOOL_HISTORY);
     }
     // GH #905: usage logging only mutates this spool — validate modified paths
     // only so a legacy out-of-range field elsewhere can't block the log/debit.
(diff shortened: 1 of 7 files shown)


hyiger:
@codex review

hyiger:
Addressed the P2 (**Restore capped history rows on fallback rollback**) in d55746fd.

Root cause was exactly as flagged: the undo-aware trim mutated pre-existing manual/nfc rows, and in the standalone fallback it ran *before* the saves — so a committed filament save followed by a later failure left those evicted rows unrecoverable after rollback.

Fix — **defer the trim** rather than snapshot/restore (a wholesale usageHistory restore would clobber concurrent entries, the sibling concern the issue also called out):
- Split the trim out of `applyJobToFilaments` into `capTouchedSpools`.
- **Transaction path:** trim *before* the saves, inside the txn — a mid-write failure rolls the eviction back atomically.
- **Standalone fallback:** trim only *after* the job is durably recorded (best-effort re-save of just the shortened spools). A rolled-back request now never deletes rows it didn't mean to touch; being a couple entries over the cap until the next write is harmless.
- Rollback mechanism unchanged (jobId-filter over a fresh reload — no wholesale snapshot restore).

New regression test in `tests/print-history-concurrency.test.ts`: filament A pre-filled to the cap with manual logs + a forced save #2 failure asserts all `MAX_SPOOL_HISTORY` manual rows survive the rollback (confirmed it fails at `MAX-1` under the old eager trim). Full suite green (2919 tests), lint + `tsc` clean.

@codex review

hyiger:
Addressed the P2 (**Preserve the manual entry when capping full job histories**) in 19517bc3.

Good catch — the asymmetry was real: `capUsageHistory` evicts non-undo entries oldest-first, but the caller's freshly-appended row *is* non-undo (a manual/nfc log), so on a spool already full of undo-relevant `job`/`slicer` entries it was the only non-undo candidate and got dropped — while the manual route had already debited `totalWeight` and returned 201, silently losing the just-logged use.

Fix in the helper: **the last entry (the push-then-cap row the caller just recorded) is never an eviction candidate.** When protecting it forces evicting an old undo entry — only reachable on a spool ~1000 jobs deep — that's the right trade: losing what the user just did, with weight already debited, is worse than a rare refund gap on a decade-deep history. Realistic spools still roll off old manual/nfc rows first, so live job entries stay protected (the original #954 guarantee).

Tests: a pure case (spool full of jobs + new manual → new manual kept, oldest job evicted) and a manual-route case (201 persists the manual **and** its debit); both verified failing without the guard. Full suite green (2921).

The other thread (fallback-rollback) is anchored to the earlier commit and was resolved in d55746fd.

@codex review

hyiger:
Branch updated with `main` (merge `eaf8bfc8` pulled in #960; no conflicts, my files untouched). Re-gating on the current HEAD.

@codex review

PR #964, opened 2026-07-03
fix(filaments): preserve variant inheritance on import/sync round-trips (#951)

hyiger:
Fixes #951 — one bug class across three (really four) import/sync paths that
wrote parent-resolved **flattened** values as **local overrides** onto
variants, severing GH #106 / #628 live inheritance so a later parent edit
silently stopped propagating.

The exports resolve variants through `resolveFilament`, so a variant's preset /
CSV row / INI section carries the parent's density/cost/temps/… When those come
back on a round-trip, only fields that genuinely **differ** from the parent
should be pinned; parent-equal fields must keep inheriting.

## Fixes

**951.2 — PrusaSlicer per-id sync** (`POST /api/filaments/[id]`)
Routed the built `update` through the CSV importer's battle-tested
`splitInheritedImportSet` (GH #628/#649) before the write — for a variant it
drops each inheritable scalar/temp equal to the parent and `$unset`s a stale
diverging override. Reuses the already-fetched `calParent`; variant-only /
non-inheritable keys (color, colorName, name, settings, …) pass through.

**951.1 — CSV/XLSX importer create/resurrect** (`upsertImportRows`)
`splitInheritedImportSet` only ran on the UPDATE branch; CREATE + RESURRECT
wrote the flattened doc verbatim (export → import into a fresh DB / trashed-
variant restore pinned everything). New pure `pruneInheritedCreateDoc` nulls
scalars/temps and empties `secondaryColors` when they equal the parent, fetching
the parent by id on demand (a freshly-created parent isn't in `loadParentDocs`)
and guarding a null parent. Never prunes `vendor`/`type`/`name`/`color`.

**951.3 — PrusaSlicer bulk INI import** (`POST /api/filaments/prusaslicer`)
Extracted a shared `upsertIniFilament` (`src/lib/iniImportApply.ts`): a
three-phase atomic upsert (active → resurrect-trashed → create/race) that reads
the target then writes by `_id` (mirroring the Bambu bulk route's read-then-
write-by-id, with the soft-delete state re-checked in the filter so a concurrent
delete/purge/restore falls through instead of mis-writing). It flattens the
nested INI `temperatures` object to `temperatures.<sub>` dot-keys before calling
`splitInheritedImportSet` — which additionally stops an update from wiping
untouched `nozzleRange*`/`standby` siblings (the old whole-object `$set`
replaced the subdoc).

**Also fixed: the twin `POST /api/filaments/import`** (multipart UI upload)
The issue named the text-body `prusaslicer` route, but `/api/filaments/import`
is the same INI importer (shares `collapsePerNozzleImportSections(parseIniFilaments())`)
with the identical pinning bug. Both routes now call the one shared
`upsertIniFilament`, so the bug class is fully closed and can't drift.

## Open question — resolved (parity)

create/resurrect adopt the same *equal-to-parent === inherit* rule as the update
path, so a deliberate override that happens to match the parent today becomes
dynamically tracked. This matches the accepted #628/#649 trade-off. (Per the
recommendation in the task.)

## Tests

- `tests/importFilaments.test.ts` — pure `pruneInheritedCreateDoc` cases +
  create-into-fresh-DB / genuine-override / resurrect integration.
- `tests/api-route-correctness.test.ts` — variant sync doesn't pin parent-equal
  values (inheritance survives a later parent edit), genuine override written,
  stale override `$unset`.
- `tests/filaments-import-export-route.test.ts` — variant round-trip preserved
  on both INI routes, dot-key temps preserve range/standby, trashed-parent
  guard, stale-override `$unset`.
- `tests/iniImportApply.test.ts` — the shared upsert's create/E11000-race
  recovery branches.

Full suite green: **2982 tests pass**; coverage clears thresholds
(lines 99.51%, branches 97.44%, statements 99.04%, functions 97.34%);
lint + `tsc` + `npm run build` all pass.

### DIFF
--- a/src/app/api/filaments/[id]/route.ts
+++ b/src/app/api/filaments/[id]/route.ts
@@ -9,6 +9,7 @@ import { errorResponse, errorResponseFromCaught, handleDuplicateKeyError, isDupl
 import { assertSameOriginRequest } from "@/lib/requestGuard";
 import { mergeSlicerSettings } from "@/lib/slicerSettings";
 import { resolveSyncBackColor } from "@/lib/prusaSlicerBundle";
+import { splitInheritedImportSet } from "@/lib/importFilaments";
 import { escapeRegex } from "@/lib/matchFilament";
 import { assignSpoolToSlot } from "@/lib/spoolSlots";
 import {
@@ -917,6 +918,32 @@ export async function POST(
       }
     }
 
+    // GH #951: a variant's PrusaSlicer export flattens its inherited values
+    // through resolveFilament, so the fork echoes the parent's
+    // density/cost/temps/spoolWeight/diameter/… back on every sync. Blindly
+    // $set-ing them onto the variant pins each as a local override and severs
+    // GH #106 live inheritance — a later parent edit stops propagating. Before
+    // this, only `color` had echo suppression (resolveSyncBackColor). Reuse the
+    // CSV importer's battle-tested split (GH #628 / #649): drop each inheritable
+    // field whose incoming value equals the parent's (keep inheriting), and
+    // $unset a stale diverging local override so inheritance resumes. Variant-
+    // only + non-inheritable keys (color, colorName, name, settings, soluble,
+    // …) pass through untouched. `calParent` is the already-fetched parent doc;
+    // when it's null (standalone/parent, or a soft-deleted/missing parent —
+    // nothing to inherit) the update is written verbatim.
+    const mongoUpdate: Record<string, unknown> = { $set: update };
+    if (filament.parentId && calParent) {
+      const split = splitInheritedImportSet(
+        update,
+        filament.toObject(),
+        calParent.toObject(),
+      );
+      mongoUpdate.$set = split.set;
+      if (split.unset.length > 0) {
+        mongoUpdate.$unset = Object.fromEntries(split.unset.map((k) => [k, ""]));
+      }
+    }
+
     // GH #618: `runValidators` so the numeric range validators (#337)
     // actually fire on a PrusaSlicer sync — without it a config like
     // `filament_cost = -3` parses clean and persists a negative cost the
@@ -926,7 +953,7 @@ export async function POST(
     try {
       await Filament.findByIdAndUpdate(
         filament._id,
-        { $set: update },
+        mongoUpdate,
         { runValidators: true, context: "query" },
       );
     } catch (validationErr) {
--- a/src/app/api/filaments/import/route.ts
+++ b/src/app/api/filaments/import/route.ts
@@ -1,9 +1,9 @@
 import { NextRequest, NextResponse } from "next/server";
 import dbConnect from "@/lib/mongodb";
-import Filament from "@/models/Filament";
 import { parseIniFilaments } from "@/lib/parseIni";
 import { collapsePerNozzleImportSections } from "@/lib/prusaSlicerBundle";
-import { assertMultipartFormData, checkFileSize, isDuplicateKeyError } from "@/lib/apiErrorHandler";
+import { upsertIniFilament } from "@/lib/iniImportApply";
+import { assertMultipartFormData, checkFileSize } from "@/lib/apiErrorHandler";
 import { assertSameOriginRequest } from "@/lib/requestGuard";
 
 export async function POST(request: NextRequest) {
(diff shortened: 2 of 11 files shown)


hyiger:
@codex review

hyiger:
Addressed both findings in `add8f12d`:

- **F1 (settings on per-id sync)** — `splitInheritedImportSet` now has a `settings` branch that rebuilds the bag keeping only keys differing from the parent (parent-equal keys keep inheriting; a stale key self-heals). It activates only when the caller supplies the parent's settings — the per-id sync passes `calParent.toObject()`; the INI path intentionally doesn't project `settings` (its bag is a raw key-dump), so that path is unchanged.
- **F2 (`inherits` on INI import)** — added `inherits` to `IMPORT_INHERITABLE_SCALARS` and to both projections, so a variant round-trip no longer pins the parent's echoed `inherits`.

New tests cover both (parent-equal settings not pinned + parent edit still propagates; INI `inherits` stays inheriting). Full suite green (2987 tests), coverage clears thresholds.

@codex review

hyiger:
Addressed in `f59836d8` (each finding independently verified against the code before fixing):

**Finding A (settings on per-id sync) — respectfully, this is a false positive on the current HEAD.** It was fixed in the prior commit `add8f12d`: `splitInheritedImportSet` gained a `settings` branch that rebuilds the bag keeping only keys differing from the parent, and the per-id sync passes the parent via `calParent.toObject()` — so `update.settings` is filtered before `findByIdAndUpdate`, not passed through. The test `#951 (Codex F1)` in `tests/api-route-correctness.test.ts` proves end-to-end that a variant sync no longer pins parent-equal settings and that a later parent settings edit still propagates. No further change needed here.

**Finding B (`inherits` shadow in the raw INI settings bag) — real, and broader than `inherits`.** The INI importers stored the raw settings bag, which shadows the top-level fields `parseIni` also extracts. On a variant re-import the top-level scalars correctly stayed inheriting, but the settings-bag shadow survived and `filamentToSlicerKeys` (which seeds `keys` from settings and only overwrites when the resolved top-level value is truthy) re-emitted the stale value after a parent-clear — the same leak for cost/density/diameter/temps/maxVolumetricSpeed, not just `inherits`. Fix: strip the `parseIni`-extracted keys from the settings bag up front in `upsertIniFilament` (covers all write paths incl. the phase-3 `create`); the key set is exported from `parseIni.ts` as the single source of truth with a lockstep test. Round-trip re-export test added.

**Finding C (rename race) — real regression, fixed.** Added `name` to all three by-`_id` filters in `upsertIniFilament` so a concurrent rename in the read→write window misses and falls through (restoring the old single-op semantics). The Bambu bulk route stays safe differently — it never puts `name` in its `$set`. Rename-race tests added for phase-1 and resurrect + a no-rename regression.

Full suite green (2994 tests), coverage clears thresholds.

@codex review

hyiger:
Addressed in `83b8657c` (each finding independently re-verified against the code):

**Finding 3 (dotted range-temp keys) — real, fixed.** The doc-builder added `temperatures.nozzleRangeMin/Max/standby` as dotted keys alongside the nested `temperatures` object. `pruneInheritedCreateDoc` only nulls the nested object, so on **create** the dotted key overrode the null (re-pinning an inherited range temp), and on **resurrect** the dotted+nested pair collided in one `updateOne` → `MongoServerError code 40` → the row was caught and *skipped*, so the trashed variant never came back (worse than described — nice catch). Fix: dropped the three redundant dotted assignments; range temps flow via the temps-loop `$set` (update) and the nested object (create/resurrect, which the prune handles). Regression tests for both create and resurrect.

**Finding 4 (`filament_spool_weight` / `filament_shrinkage_*` shadow) — real, fixed.** These leaked the same way as cost/density (including the form-created-parent case a settings-only filter would miss). Fix: lifted `spoolWeight`/`shrinkageXY`/`shrinkageZ` to top-level `FilamentData` fields (they're already inheritable scalars), added them to the strip set + INI projection. The lift is guarded **carry-only-when-supplied** (undefined when the INI key is absent) so a partial INI can't `$set null` and clobber an existing value on a root — covered by a no-clobber test.

**Findings A (settings-on-sync) + B (`inherits` shadow) — respectfully, false-positive re-flags.** Both were fixed in `add8f12d`/`f59836d8` and remain fixed on this HEAD:
- A: `splitInheritedImportSet`'s `settings` branch filters `update.settings` per-key against `calParent.toObject()` before `findByIdAndUpdate` — the `#951 (Codex F1)` test proves a variant sync no longer pins parent-equal settings and that parent edits still propagate.
- B: `stripStructuredSettings` runs at the top of `upsertIniFilament`, deleting `INI_TOP_LEVEL_SETTING_KEYS` (incl. `inherits`) from `collapsed.settings` before every write path (update/resurrect/create/race); the R2-B test asserts none of those keys survive in the stored variant's settings bag.

Full suite green (3000+), coverage clears thresholds. Rebased onto latest `main`.

@codex review

hyiger:
@codex review

PR #965, opened 2026-07-03
fix(share): remap source ids on shared-catalog import (#956)

hyiger:
Fixes #956.

The client-side shared-catalog import (`src/app/share/[slug]/page.tsx`) POSTed **source-DB ObjectIds** verbatim, so destination validation 400'd and silently dropped data.

## Two findings
- **956.1** — imported variants kept their **source** `parentId` → `POST /api/filaments` rejected with "Parent filament not found" (400). Filaments were also POSTed in array order, so a variant could precede its parent.
- **956.2** — printers were POSTed as raw lean docs still carrying source `installedNozzles`/`installedBedTypes`/`amsSlots` → printers POST 400'd → `printerMap` stayed empty → every printer-scoped calibration silently degraded to `printer: null`.

## Fix
- **Serialized ref import**: nozzles + bed types first, then printers — so a printer's installed refs are remapped through the fresh id maps. `amsSlots` is stripped (source spool ids are meaningless on the destination; the publish route already strips spools).
- **Two-phase filament import**: roots first (building a source→local id map), then variants with `parentId` remapped through it.
- **Orphan-variant policy** (your call): a variant whose parent wasn't imported is **skipped + surfaced as a conflict** — importing it standalone would drop every inherited value (new `share.public.orphanVariant` key, en+de).
- **Printer refs policy** (your call): installed nozzles/bed types are **remapped** through the id maps (unresolved dropped); amsSlots stripped.
- A root name-collision (409) resolves the existing local filament by name, so variants still attach to a parent the recipient already owns (mirrors the ref rehydrate's 409-reuse).

The pure transforms live in a new unit-tested `src/lib/shareImport.ts` (`partitionByParent`, `buildFilamentImportBody`, `buildPrinterImportBody`) — the component itself is a `"use client"` page with no route test, so the id-remap/strip logic is covered there.

Full suite green (3023 tests), coverage clears thresholds, lint + tsc pass.

### DIFF
--- a/src/app/share/[slug]/page.tsx
+++ b/src/app/share/[slug]/page.tsx
@@ -7,6 +7,12 @@ import { useToast } from "@/components/Toast";
 import FilamentPicker from "@/components/FilamentPicker";
 import { useTranslation } from "@/i18n/TranslationProvider";
 import { formatDate } from "@/lib/dateFormat";
+import {
+  partitionByParent,
+  buildFilamentImportBody,
+  buildPrinterImportBody,
+  type ShareImportPrinter,
+} from "@/lib/shareImport";
 
 interface SharedFilament {
   _id: string;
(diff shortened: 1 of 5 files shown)


hyiger:
@codex review

PR #966, opened 2026-07-03
fix: data-integrity remaining batch — 4 fixes from #954

hyiger:
Fixes #954 — the four still-open sub-findings (the other five shipped in #959 / #960 / #961; the triage verified those against HEAD).

| # | Fix |
|---|-----|
| **954.1** | `getRemainingGrams` returned `null` on a null `spoolWeight`, so the home-list low-stock badge/filter never fired for legacy filaments — while by-location/dashboard/locations already fall back to a **0g tare**. Aligned the `spools[]` branch to the 0g-tare posture. |
| **954.5** | The 8px font floor could make the text block taller than the 128-dot print head, silently clipping lines. Now **throws** (fail-loud, like `renderQr`) when the text footprint exceeds the head, and `normalizeLabelFormat` **dedupes** its `lines` so a persisted format can't stack a field N times. |
| **954.8** | `spool-check` returned `ok:true` "no data" when every weighted spool was **retired**, suppressing PrusaSlicer's insufficient-filament warning. Now distinguishes "no weight data at all" (`ok:true`) from "all weighted spools retired" (`ok:false` + `warning`). |
| **954.9** | CSV/XLSX dropped `optTags`, so a coextruded/gradient/finish filament collapsed to solid on a create-path round-trip. Added a **`Tags`** column, parsed on CREATE/RESURRECT only (like `Parent`), with `optTags` added to `pruneInheritedCreateDoc`'s whole-array inheritance so a variant's tags aren't pinned. |

Product/UX calls were confirmed with the maintainer (0g-tare; throw; ok:false; add Tags column).

Two existing tests updated to the intended new behavior (`getRemainingGrams` 0g-tare; the #273 retired-parent-spool spool-check now warns instead of silently passing — a strict improvement).

Full suite green (3022 tests), coverage clears thresholds, lint + tsc pass.

### DIFF
--- a/src/app/api/filaments/[id]/spool-check/route.ts
+++ b/src/app/api/filaments/[id]/spool-check/route.ts
@@ -167,16 +167,17 @@ export async function GET(
       });
 
     if (spoolResults.length === 0) {
-      // GH #954: distinguish "no weight data at all" (genuinely can't check →
-      // ok:true) from "weight data exists but every such spool is RETIRED"
-      // (zero active stock). The retired exclusion above exists precisely to
-      // stop a retired spool from suppressing the slicer's insufficient-filament
-      // warning — so the all-retired case must warn (ok:false), not silently
-      // pass with the misleading "no data" message.
+      // GH #954: warn (ok:false) ONLY in the true zero-active-stock case —
+      // there are NO active spools AND the only weight data is on retired ones.
+      // If an ACTIVE spool exists but is simply unweighed, active stock exists
+      // (just unmeasured), so keep the original "no data → ok:true" behavior and
+      // don't emit a false warning (Codex). The retired exclusion still stops a
+      // retired spool from silently satisfying the check.
+      const hasActiveSpool = rawSpools.some((s) => !s.retired);
       const hasRetiredWeightData = rawSpools.some(
         (s) => s.totalWeight != null && s.retired,
       );
-      if (hasRetiredWeightData) {
+      if (!hasActiveSpool && hasRetiredWeightData) {
         return NextResponse.json({
           ok: false,
           filament: filament.name,
--- a/src/app/api/filaments/[id]/spool-check/route.ts
+++ b/src/app/api/filaments/[id]/spool-check/route.ts
@@ -131,16 +131,6 @@ export async function GET(
       });
     }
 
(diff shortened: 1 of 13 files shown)


hyiger:
@codex review

hyiger:
Addressed all three findings in the latest push:
- **optTags empty-cell clear** — a present-but-empty `Tags` column now sets `optTags = []` (rowToImport maps an empty mapped cell to `null`, distinct from an absent column's `undefined`), so a resurrect clears a tombstone's tags instead of leaving them.
- **empty-token guard** — empty tokens are dropped before `Number()`, so `28,16,` / `28,,16` no longer add a phantom tag 0 (glass-fiber).
- **spool-check** — now warns (ok:false) only when there are NO active spools and weight data is all-retired; an active-but-unweighed spool keeps the ok:true "no data" behavior (no false warning).

New tests for each. Full suite green (3025), coverage clears thresholds.

@codex review

hyiger:
@codex review

hyiger:
Both spool-check findings fixed in `30e10477`.

### ✅ Move retired-only detection before the tare guard
Confirmed real: when a filament had no active spools, retired weighed stock, and a null/inherited-missing `spoolWeight`, the `rawSpools.length === 0 || spoolWeight == null` guard returned `ok:true` before the retired-stock check ran, suppressing the warning for null-tare legacy data. The all-retired detection doesn't need the tare (it's a pure retired-vs-active question), so it now runs **before** the tare guard. Test: a filament with only retired weighed stock and no `spoolWeight` now returns `ok:false` with the retired warning.

### ✅ Preserve no-data behavior for unweighed active spools
The `hasActiveSpool` guard already covers this — an active spool with no `totalWeight` counts as active stock (just unmeasured), so `!hasActiveSpool && hasRetiredWeightData` is false and the request falls through to the "no data → `ok:true`" guard rather than warning. I kept that guard on the hoisted check and pinned it with a test: an unweighed active spool + a weighed retired spool returns `ok:true` (no false warning), not `ok:false`.

Net structure now: build spools → all-retired check (pre-tare, warns) → tare guard (no data → ok:true, includes unweighed-active) → per-spool check. Full tsc / lint / coverage gate green.

@codex review

PR #967, opened 2026-07-04
fix(nfc): tag codec correctness — 5 fixes from #952

hyiger:
Addresses #952 — **5 of 6** sub-findings (952.4 deferred, see below). Each is a byte-level encode/decode correctness fix; unit tests updated/added.

| # | Fix |
|---|-----|
| **952.1** | Bambu dual-color: block-16 second color read **reverse-ABGR** (was RGBA), per the RFID-Tag-Guide. ⚠️ Reference-derived — **on-hardware confirmation on a real dual-color Bambu spool is still owed** (you approved shipping it flagged). |
| **952.2** | OPT: **omit** an over-length `spoolUid` (>16) instead of truncating — a truncated id reads back as a *different* id and breaks scan-back matching. |
| **952.3** | OpenTag3D: `writeUintBE` **clamps** to the field's byte capacity (was a silent wrap — maxVolumetricSpeed 300 → 44); `setNum` surfaces a notice. |
| **952.5** | OpenTag3D decode **rejoins** material base + modifier into the typed field (`PC-ABS` decoded as bare `PC`), fixing the anchored vendor+type match + finish. Hyphen-default separator (your call). |
| **952.6** | OPT: **omit** `country_of_origin` when unset — the old `?? "US"` stamped a fabricated `US` onto every tag/re-write. |

### 952.4 deferred (needs your call — the triage under-specified it)
The bed/nozzle first-layer round-trip fix is **not** in this PR. You approved "decode: everyday=MIN, first-layer=MAX", but on inspection the encoder **synthesizes `MIN_BED = everyday − 10`** when there's no first-layer temp — so a decode-only change would make every *no-first-layer* filament read its bed temp 10° low. The correct fix requires an **encode** change (write `MIN==MAX` when there's no first-layer), which then **mis-reads pre-fix hardware tags by 10°**. That's a wire-format tradeoff on physical tags I didn't want to make unilaterally. I'll implement it on your go — just confirm the pre-fix-tag drift is acceptable.

Full suite green (3017 tests), coverage clears thresholds, lint + tsc pass.

### DIFF
--- a/src/lib/bambuTag.ts
+++ b/src/lib/bambuTag.ts
@@ -142,13 +142,17 @@ export function parseBambuBlocks(blocks: (Buffer | undefined)[]): BambuTagData {
   const b14 = block(14);
   const filamentLength = b14.readUInt16LE(4);
 
-  // Block 16: Format ID uint16 LE (0-1) + Color Count uint16 LE (2-3) + Second Color RGBA (4-7)
+  // Block 16: Format ID uint16 LE (0-1) + Color Count uint16 LE (2-3) + Second
+  // Color at 4-7. GH #952: the block-5 PRIMARY color is stored RGBA, but the
+  // Bambu-Research-Group RFID-Tag-Guide documents this block-16 SECONDARY color
+  // in REVERSE ABGR order — so read bytes 7..4 as [r,g,b,a]. (Reference-derived;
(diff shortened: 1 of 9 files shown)


hyiger:
@codex review

PR #968, opened 2026-07-04
fix(slicer): round-trip data-loss & id-first addressing — from #950

hyiger:
Addresses #950 — slicer round-trip data-loss & addressing bugs. **8 of 10 findings fixed; 950.4 + 950.8b deferred** (see below), so this does not auto-close the issue.

### 950.1 — `filament_soluble` / `filament_abrasive` silently dropped
The Filament schema has no `soluble`/`abrasive` columns, so the per-id sync route pulled these keys out of the settings bag and wrote them to a dead `update.soluble`/`update.abrasive` that Mongoose stripped — persisting them **nowhere**, and the exporter's `set("filament_soluble", filament.soluble)` was a permanent no-op. Removed from `STRUCTURED_KEYS` and dropped the dead export reads (PrusaSlicer + OrcaSlicer) so both keys ride the settings passthrough bag and a slicer round-trip preserves them.

### 950.2 — per-nozzle calibration bake clobbered a user-pinned `compatible_printers_condition`
The `#872` per-nozzle bake set `keys.compatible_printers_condition = nozzle_diameter[0]==<Ø>` **unconditionally**, overwriting a round-tripped user pin or the `nil` inheritance marker. Gated it the same way the non-calibration derivation already is — only set when absent or an empty-string "no restriction". `filamentdb_nozzle` stays unconditional (it's a routing hint for that nozzle, not a user setting).

### 950.3 — chamber temperature dropped on Bambu Studio import
`orcaSlicerBundle` exports `chamber_temperature` + `activate_chamber_temp_control`, but the Bambu importer never parsed them back → every export→calibrate→re-import cycle lost the chamber temp. Added `chamberTemp` to `CalibrationHints` + `CALIBRATION_KEYS`, honoring `activate_chamber_temp_control="0"` (heating off → don't import), and applied it onto `calibrations[].chamberTemp`.

### 950.5 — stale `filament_settings_id` re-exported after a rename
The INI bulk importer dumped every `key=value` line into the settings bag, so `filament_settings_id` survived verbatim. Since the export re-derives it from the filament's **current** name, a stale copy made a renamed filament export its old preset name. Added it to `SETTINGS_STRIP_KEYS` (alongside the routing hints) so it's stripped from the stored bag and always re-derived.

### 950.6 — slicer-facing routes resolved by name before _id (`_id`-shadow)
`resolveFilamentForExport` and the per-id OrcaSlicer sync tried a **name** lookup first, so a filament legitimately NAMED with 24 hex chars would shadow another filament's real `_id`. Made a 24-hex param authoritative (try `_id` first, fall back to name only on a miss) — matching the `#867` per-id sync semantics. **An adversarial-verification pass found two sibling slicer-facing GET routes with the same name-first ordering** — `GET /api/filaments/{id}/calibration` and `GET /api/filaments/{id}/spool-check` — which was *worse* than a plain shadow: the sync/export routes now resolve id-first while these read by-name, so a slicer round-trip on a 24-hex-named-decoy would write one filament and read another's calibration / spool availability. Both flipped to id-first for consistency.

### 950.7 — per-nozzle preset with a stale `filamentdb_id` 404'd → orphan
A `#872` preset is named `<base> <Ø type [HF]>`. When its `filamentdb_id` was stale/absent (it's DB-instance-specific) AND the full suffixed name missed, the sync 404'd and the fork spawned a `<base> <hint>` orphan that swallowed every later edit. Added a base-name retry (suffixed name minus the `filamentdb_nozzle` hint) before the 404.

### 950.8a — INI import clobbered base fields with parseIni defaults (non-hinted branch)
`parseIni` fills absent fields with defaults (`null` / `#808080` / `1.75` / `Unknown`); `$set`-ing those over a name-matched existing filament wiped its real cost/density/color/diameter/vendor/type/max-vol on a partial or hand-crafted section. Extended the "leave-when-omitted" idiom (already used by the hinted branch) to the non-hinted pass-through branch — drop each field the section didn't actually supply.

### 950.9 — INI import id/name mismatch resolved ambiguously
Captured the section's round-tripped `filamentdb_id` (`CollapsedFilamentData.filamentdbId`, never persisted) and added an id-first **refuse-ambiguous** pre-check to `upsertIniFilament`: when the id resolves to an ACTIVE filament whose stored name differs from the section name, the case is ambiguous (renamed preset vs copied/stale id), so it throws → per-row skip rather than silently mutating the wrong row or spawning an orphan. Mirrors the per-id sync route's `name_id_mismatch` conservatism.

## Deferred (not in this PR)
- **950.4** — deferred as an enhancement (per triage decision).
- **950.8b** — deferred (per triage decision).

## Tests
New coverage for every shipped finding across `prusaSlicerBundle`, `orcaSlicerBundle`, `iniImportApply`, `bambuStudioImport`, `bambuStudioApply`, `single-filament-slicer-export`, `calibration-route`, and the per-id sync + orcaslicer + spool-check routes (`api-route-correctness`). Existing tests that asserted the removed dead-field behavior were updated to the new settings-bag round-trip contract, plus a **regression guard** proving the exporter no longer reads a dead top-level `soluble`/`abrasive` field (settings bag is authoritative — a revert now fails a test).

## Verification
An adversarial multi-agent verification pass ran one skeptic per finding plus a completeness/scope critic and a test-quality critic. 6/8 findings passed clean; it surfaced the 950.6 scope gap (calibration + spool-check routes, now fixed) and the missing exporter regression guard (now added). Full `tsc`, `lint`, and coverage gate green.

### DIFF
--- a/src/app/api/filaments/[id]/calibration/route.ts
+++ b/src/app/api/filaments/[id]/calibration/route.ts
@@ -45,17 +45,24 @@ export async function GET(
       );
     }
 
-    // Find filament by name or ObjectId. `params.id` is ALREADY URL-decoded —
-    // re-decoding throws URIError on a name with a literal `%` (#671).
+    // GH #950 / #867: a 24-hex param is an ObjectId and is AUTHORITATIVE — try
+    // it FIRST, name lookup only when that _id misses (a preset legitimately
+    // NAMED with 24 hex chars). Name-first let such a name shadow another
+    // filament's real _id, so a slicer addressing this endpoint by id read the
+    // WRONG row's calibration — inconsistent with the id-first sync/export
+    // routes. `params.id` is ALREADY URL-decoded — do NOT re-decode (a literal
+    // `%` like "ABS 100%" throws URIError and 500s the request, #671).
     const decodedName = id;
-    let filament = await Filament.findOne({ name: decodedName, _deletedAt: null })
-      .populate("calibrations.nozzle")
-      .populate("calibrations.printer")
-      .populate("calibrations.bedType")
-      .lean();
+    let filament = /^[a-f0-9]{24}$/i.test(id)
+      ? await Filament.findOne({ _id: id, _deletedAt: null })
+          .populate("calibrations.nozzle")
+          .populate("calibrations.printer")
+          .populate("calibrations.bedType")
+          .lean()
+      : null;
 
-    if (!filament && /^[a-f0-9]{24}$/i.test(id)) {
-      filament = await Filament.findOne({ _id: id, _deletedAt: null })
+    if (!filament) {
+      filament = await Filament.findOne({ name: decodedName, _deletedAt: null })
         .populate("calibrations.nozzle")
         .populate("calibrations.printer")
         .populate("calibrations.bedType")
--- a/src/app/api/filaments/[id]/orcaslicer/route.ts
+++ b/src/app/api/filaments/[id]/orcaslicer/route.ts
@@ -240,7 +240,12 @@ export async function POST(
       return errorResponse(merge.error, 400);
     }
     const settingsAdded = merge.added;
-    if (settingsAdded.length > 0) {
+    // GH #950 (Codex P2 on PR #968 r5): also write when the merge PURGED a stale
+    // structured key from the existing bag (`removed`) — otherwise a sync that
+    // only changes structured fields discards the cleaned bag and the stale
+    // shadow (e.g. a legacy filament_settings_id) survives to shadow the
+    // re-derived export value.
+    if (settingsAdded.length > 0 || merge.removed.length > 0) {
       update.settings = merge.settings;
     }
 
--- a/src/app/api/filaments/[id]/orcaslicer/route.ts
+++ b/src/app/api/filaments/[id]/orcaslicer/route.ts
@@ -131,13 +131,17 @@ export async function POST(
     await dbConnect();
     const { id } = await params;
 
-    // Find filament by name or ObjectId. The App Router `params.id` is ALREADY
-    // URL-decoded — re-decoding throws URIError on a name with a literal `%`
-    // ("ABS 100%") and 500s the sync-back (#671; cf. resolveFilamentForExport).
+    // GH #950 / #867: a 24-hex param is an ObjectId and is AUTHORITATIVE — try it
+    // FIRST, name lookup only when that _id misses (a preset named with 24 hex
+    // chars). Name-first let such a name shadow another filament's real _id. The
+    // App Router `params.id` is ALREADY URL-decoded — do NOT re-decode (a literal
+    // `%` like "ABS 100%" would throw URIError and 500 the sync, #671).
     const decodedName = id;
-    let filament = await Filament.findOne({ name: decodedName, _deletedAt: null });
-    if (!filament && /^[a-f0-9]{24}$/i.test(id)) {
-      filament = await Filament.findOne({ _id: id, _deletedAt: null });
+    let filament = /^[a-f0-9]{24}$/i.test(id)
+      ? await Filament.findOne({ _id: id, _deletedAt: null })
+      : null;
+    if (!filament) {
+      filament = await Filament.findOne({ name: decodedName, _deletedAt: null });
     }
 
     if (!filament) {
--- a/src/app/api/filaments/[id]/route.ts
+++ b/src/app/api/filaments/[id]/route.ts
@@ -469,6 +469,22 @@ export async function POST(
         filament = await Filament.findOne({ name: decodedName, _deletedAt: null });
         if (filament) matchedBy = "name";
       }
+      // GH #950: a #872 per-nozzle preset is named "<base> <Ø type [HF]>". When
+      // its filamentdb_id is stale/absent (DB-instance-specific) AND the full
+      // suffixed name misses, retry the BASE name (decodedName minus the hint)
+      // so the sync updates the base filament instead of 404 → the fork spawning
+      // a "<base> <hint>" orphan that swallows every later edit.
+      if (!filament) {
+        const hint =
+          typeof config.filamentdb_nozzle === "string" ? config.filamentdb_nozzle.trim() : "";
+        if (hint && decodedName.endsWith(` ${hint}`)) {
+          const baseName = decodedName.slice(0, -(hint.length + 1)).trim();
+          if (baseName) {
+            filament = await Filament.findOne({ name: baseName, _deletedAt: null });
+            if (filament) matchedBy = "name";
+          }
+        }
+      }
     }
 
     if (!filament) {
@@ -594,9 +610,10 @@ export async function POST(
     if (config.filament_shrinkage_compensation_xy) { const v = parseFloat(config.filament_shrinkage_compensation_xy); if (!isNaN(v)) update.shrinkageXY = v; }
     if (config.filament_shrinkage_compensation_z) { const v = parseFloat(config.filament_shrinkage_compensation_z); if (!isNaN(v)) update.shrinkageZ = v; }
 
-    // Flags
-    if (config.filament_soluble) update.soluble = config.filament_soluble === "1";
-    if (config.filament_abrasive) update.abrasive = config.filament_abrasive === "1";
+    // GH #950: filament_soluble / filament_abrasive are NOT written as structured
+    // fields (the schema has no such columns — a Mongoose strict write dropped
+    // them). They now ride the settings bag (removed from STRUCTURED_KEYS below),
+    // which the slicer exports' settings seed and the OPT encoder read.
 
     // #859: write ONLY the temperature keys PrusaSlicer actually sent, as
     // dotted paths — never a $set of the whole `temperatures` object. #645 added
@@ -815,7 +832,12 @@ export async function POST(
       "filament_max_volumetric_speed", "temperature", "first_layer_temperature",
       "bed_temperature", "first_layer_bed_temperature",
       "filament_shrinkage_compensation_xy", "filament_shrinkage_compensation_z",
-      "filament_soluble", "filament_abrasive", "filament_settings_id",
+      // GH #950: filament_soluble / filament_abrasive are NOT structured — the
+      // Filament schema has no such fields, so a sync that pulled them out here
+      // (and wrote a stripped-by-Mongoose update.soluble) persisted them
+      // NOWHERE. Let them ride the settings bag, which is where the exports'
+      // settings seed and the OPT encoder already read them from.
+      "filament_settings_id",
       // #867: a routing hint, not filament data — consumed for id-first matching
       // above and re-emitted from the row's _id on export, so never stored in
       // the settings bag (avoids a stale duplicate of the canonical _id).
--- a/src/app/api/filaments/[id]/spool-check/route.ts
+++ b/src/app/api/filaments/[id]/spool-check/route.ts
@@ -42,12 +42,19 @@ export async function GET(
       return errorResponse("weight must be a non-negative number", 400);
     }
 
-    // Find filament by name or ObjectId. `params.id` is ALREADY URL-decoded —
-    // re-decoding throws URIError on a name with a literal `%` (#671).
+    // GH #950 / #867: a 24-hex param is an ObjectId and is AUTHORITATIVE — try
+    // it FIRST, name lookup only when that _id misses (a preset legitimately
+    // NAMED with 24 hex chars). Name-first let such a name shadow another
+    // filament's real _id, so a slicer addressing this endpoint by id checked
+    // the WRONG row's spool availability — inconsistent with the id-first
+    // sync/export routes. `params.id` is ALREADY URL-decoded — do NOT re-decode
+    // (a literal `%` like "ABS 100%" throws URIError and 500s the request, #671).
     const decodedName = id;
-    let filament = await Filament.findOne({ name: decodedName, _deletedAt: null }).lean();
-    if (!filament && /^[a-f0-9]{24}$/i.test(id)) {
-      filament = await Filament.findOne({ _id: id, _deletedAt: null }).lean();
+    let filament = /^[a-f0-9]{24}$/i.test(id)
+      ? await Filament.findOne({ _id: id, _deletedAt: null }).lean()
+      : null;
+    if (!filament) {
+      filament = await Filament.findOne({ name: decodedName, _deletedAt: null }).lean();
     }
 
     if (!filament) {
--- a/src/lib/bambuStudioApply.ts
+++ b/src/lib/bambuStudioApply.ts
@@ -180,15 +180,28 @@ export async function prepareBambuUpdate(
   // filament_soluble / slicer options. `settingsResult.settings` is always the
   // full {existing, ...incoming} bag. Skip when the merge errored (the route 400s
   // on it anyway) so we never build on a partial bag.
-  if (
-    parsed.calibrationHints.chamberTemp != null &&
-    !calibrationOutcome.applied &&
-    !settingsResult.error
-  ) {
-    const settings = { ...settingsResult.settings };
-    settings.chamber_temperature = String(parsed.calibrationHints.chamberTemp);
-    settings.activate_chamber_temp_control = "1";
-    update.settings = settings;
+  if (parsed.calibrationHints.chamberTemp != null && !settingsResult.error) {
+    if (!calibrationOutcome.applied) {
+      // Unresolved: no structural home → preserve the raw chamber keys in the bag.
+      const settings = { ...settingsResult.settings };
+      settings.chamber_temperature = String(parsed.calibrationHints.chamberTemp);
+      settings.activate_chamber_temp_control = "1";
+      update.settings = settings;
+    } else if (
+      "chamber_temperature" in settingsResult.settings ||
+      "activate_chamber_temp_control" in settingsResult.settings
+    ) {
+      // Codex P2 on PR #968 r4: the chamber value went to calibrations[].chamberTemp
+      // (authoritative, per-nozzle). Strip any STALE raw chamber keys the merged bag
+      // carried over from a PRIOR unresolved/disabled import — otherwise they'd
+      // re-export as a filament-global chamber value that double-counts the
+      // calibration. The parser already excludes the INCOMING chamber keys from the
+      // bag, so this only clears a carried-over value from `existing`.
+      const settings = { ...settingsResult.settings };
+      delete settings.chamber_temperature;
+      delete settings.activate_chamber_temp_control;
+      update.settings = settings;
+    }
   }
 
   // GH #892: reject an inverted nozzle range, mirroring the OrcaSlicer sync
--- a/src/lib/bambuStudioApply.ts
+++ b/src/lib/bambuStudioApply.ts
@@ -404,12 +404,16 @@ export async function resolveAndApplyCalibration(
     return { applied: false, unresolved: hints.hasAnyHint };
   }
 
-  // GH #950 (Codex r5/r6): a "chamber-disable-only" sync — an explicit chamber
+  // GH #950 (Codex r5/r6/r7): a "chamber-disable-only" sync — an explicit chamber
   // disable with NO real calibration hint and no new chamber temp. Its ONLY job is
-  // to clear a pre-existing calibrations[].chamberTemp; it must NOT fabricate a new
-  // per-nozzle calibration row out of the profile's ordinary top-level temps
-  // (those already land on the filament via buildStructuredUpdate, and a fabricated
-  // row would later shadow user-edited top-level temps in /calibration).
+  // to CLEAR a pre-existing calibrations[].chamberTemp. It must therefore carry
+  // JUST the chamber clear: it must not fabricate a new per-nozzle calibration, nor
+  // update an existing row's OTHER fields. Every non-chamber value the profile
+  // carries (ordinary temps, and maxVolumetricSpeed — which is excluded from
+  // hasAnyHint precisely because it has a top-level home) already lands on the
+  // filament via buildStructuredUpdate, so gating ALL of them behind
+  // !chamberClearOnly is the root fix that closes this whole leak class at once
+  // (each round Codex found another individual field slipping through).
   const chamberClearOnly =
     hints.chamberDisabled === true && !hints.hasAnyHint && hints.chamberTemp == null;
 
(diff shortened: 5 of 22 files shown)


hyiger:
@codex review

hyiger:
Fixed in d6141812.

**P2 (chamber_temperature dropped on unresolved import)** — confirmed the regression: with the chamber keys in `CALIBRATION_KEYS` they're excluded from the settings passthrough, and `resolveAndApplyCalibration` writes `calibrations[].chamberTemp` only when it resolves a printer/nozzle context — so a standalone profile with no matchable printer dropped the value entirely (worse than the pre-fix "misfiled in settings" state #950 rated acceptable).

Fix, matching the `maxVolumetricSpeed` precedent (no data loss ⇒ no spurious warning):
- excluded `chamberTemp` from `hasAnyHint` so a chamber-only profile no longer trips `calibrationUnresolved`, and
- in `prepareBambuUpdate`, when no calibration row was written (`chamberTemp != null && !applied`) it falls back to preserving the raw chamber keys in the settings bag.

A *resolved* profile still keeps chamber cleanly in `calibrations[].chamberTemp` and out of the filament-global bag (the #950 intent). New tests: resolved→calibration (out of settings) and unresolved→settings-bag fallback (no warning), plus updated `hasAnyHint` expectations. Full tsc / lint / coverage gate green.

@codex review

hyiger:
Round 2 addressed in e9146521 (all confirmed real).

**P1 — settings-bag clobber** (`bambuStudioApply.ts:178`): correct — the fallback based on `update.settings` (undefined when the profile adds no passthrough keys) made the `$set` replace the whole bag. Now bases on the merged `settingsResult.settings` (full `{existing, ...incoming}`), skipping when the merge errored. Regression test: unresolved chamber-only onto an existing filament preserves `filament_notes`/`filament_soluble`.

**P2 — resolution vs. warning coupling** (`bambuStudioApply.ts:380`): correct — excluding `chamberTemp` from `hasAnyHint` also skipped resolution. Decoupled: `wantsResolution = hasAnyHint || chamberTemp != null`, while the unresolved warning stays gated on `hasAnyHint`. Test: a chamber-only profile with a resolvable printer now lands in `calibrations[].chamberTemp` (not the bag), no warning.

**P2 — chamber preserved on unresolved** (`bambuStudioImport.ts:152`): resolved by the corrected fallback above.

I also ran an adversarial pass that caught a **third** defect the two P2s were adjacent to: a **disabled** chamber (`activate_chamber_temp_control="0"`) clears `chamberTemp`, so neither the calibration row nor the applier fallback carried it while `CALIBRATION_KEYS` still stripped the raw keys — dropping them entirely. The parser now keeps the raw chamber keys in the settings bag when the chamber is disabled/ineffective (their only home), so a disabled-chamber profile round-trips. Test added.

Full tsc / lint / coverage gate green.

@codex review

hyiger:
Round 3: one new finding fixed (`e04322df`); the other two are re-flags of findings already fixed in `e9146521` — evidence below.

### ✅ P2 `prusaSlicerBundle.ts:227` — pinned `compatible_printers_condition` lost on multi-nozzle re-import — **FIXED**
Correct and thanks — this is the import half of 950.2. The 950.2 export now carries a user pin (or the `nil` marker) through each per-nozzle section, but `collapsePerNozzleImportSections` stripped `compatible_printers_condition` unconditionally via `PER_NOZZLE_BAKED_SETTING_KEYS`, and since the importer `$set`s the whole settings bag the pin was destroyed on export→import. Fix: in the hinted collapse, strip **only** the auto-derived `nozzle_diameter[...]` shape (or an empty "no restriction"); preserve a user pin (any other non-empty string) and `nil` (null). The single-nozzle/non-hinted branch never stripped this key, so it already round-tripped there. Tests: keep-pin, strip-derived, keep-nil, and a full multi-nozzle export→parse→collapse round-trip.

### ↩️ P2 `bambuStudioImport.ts:152` — "chamber dropped when unresolved" — **already fixed in `e9146521` (re-flag)**
This describes the pre-`e9146521` behavior. The applier now preserves the raw chamber keys in the settings bag when calibration can't resolve — `src/lib/bambuStudioApply.ts:183-192` (`chamberTemp != null && !applied && !settingsResult.error` → writes `chamber_temperature`/`activate_chamber_temp_control` into `update.settings`, based on the merged bag). Pinned by `tests/bambustudio-route.test.ts` → *"an UNRESOLVED chamber_temperature falls back to the settings bag (not dropped)"*, which asserts `stored.settings.chamber_temperature === "50"` after a standalone (no-printer) import. Green. So a standalone import does **not** drop it.

### ↩️ P2 `bambuStudioApply.ts:403` — "`row.chamberTemp` unreachable for chamber-only" — **already fixed in `e9146521` (re-flag)**
The early-return guard was decoupled: `const wantsResolution = hints.hasAnyHint || hints.chamberTemp != null;` (`src/lib/bambuStudioApply.ts:378`), so a chamber-only profile no longer returns before matching — it proceeds to `matchPrinterNozzle` and, when a printer resolves, reaches `row.chamberTemp` (line 403). Pinned by `tests/bambustudio-route.test.ts` → *"a chamber-ONLY profile with a resolvable printer lands in `calibrations[].chamberTemp`"*, which asserts `cal.chamberTemp === 48` (and it is **not** in the settings bag). Green. So the assignment is reachable and chamber-only profiles resolve structurally.

Full tsc / lint / coverage gate green.

@codex review

hyiger:
Round 4: two new findings fixed (`c21d9f5d`); the rest are re-flags of already-fixed findings.

### ✅ P2 `bambuStudioApply.ts:185` — stale settings-bag chamber not cleared on a resolved import — **FIXED (new)**
Good catch. When an existing filament carried a raw `settings.chamber_temperature` from a prior unresolved/disabled import, a later resolvable import wrote `calibrations[].chamberTemp` but left the stale bag key in place → a double filament-global value on re-export. The `prepareBambuUpdate` chamber block now has an **applied** branch: when the value went to `calibrations[].chamberTemp`, it strips `chamber_temperature`/`activate_chamber_temp_control` from the merged bag (preserving unrelated settings). Test: *"a resolved chamber import CLEARS a stale settings-bag chamber from a prior import"* (asserts `cal.chamberTemp===45`, bag chamber keys gone, `filament_notes` kept).

### ✅ P2 `prusaSlicerBundle.ts:240` — over-broad derived-condition match — **FIXED (new)**
Correct — the `includes("nozzle_diameter[")` substring test would strip a compound user pin like `printer_model==MK4 and nozzle_diameter[0]==0.4`. Tightened to match ONLY the exact generated shape: `^nozzle_diameter[\d+]==<num>( or nozzle_diameter[\d+]==<num>)*$`. A user pin that merely references a diameter is now kept. Tests: keep-compound-pin, strip-multi-term-derived.

### ↩️ P2 `prusaSlicerBundle.ts:246` — pinned condition lost on collapse — **fixed in `e04322df` (re-flag)**
The hinted collapse no longer deletes a user pin / nil (only the auto-derived shape, now precisely matched per the fix above). Pinned by the collapse unit tests + a full multi-nozzle export→parse→collapse round-trip in `tests/prusaSlicerBundle.test.ts`.

### ↩️ P2 `bambuStudioImport.ts:152` — chamber dropped when unresolved — **fixed in `e9146521` (re-flag, 4th)**
The applier preserves the raw chamber keys in the settings bag when calibration can't resolve — `bambuStudioApply.ts:183-192`. Green regression test: *"an UNRESOLVED chamber_temperature falls back to the settings bag (not dropped)"* (`stored.settings.chamber_temperature === "50"` after a no-printer import). A standalone import does not drop it.

### ↩️ P2 `bambuStudioApply.ts:403` — `row.chamberTemp` unreachable for chamber-only — **fixed in `e9146521` (re-flag, 4th)**
`wantsResolution = hints.hasAnyHint || hints.chamberTemp != null` (`bambuStudioApply.ts:378`) — a chamber-only profile no longer returns early; it resolves and reaches line 403. Green regression test: *"a chamber-ONLY profile with a resolvable printer lands in `calibrations[].chamberTemp`"* (`cal.chamberTemp === 48`, not in the bag).

### 🔎 Proactive — latent per-id settings-bag leak closed (adversarial sweep)
While flushing the remaining edges, an adversarial pass found a **latent P3** in the same 950.5 area: `mergeSlicerSettings` seeded the bag from `{...existing}` and only skipped structured keys from the *incoming* config — so a stale `filament_settings_id`/`filamentdb_id` already sitting in a legacy bag would survive the per-id merge and shadow the re-derived export value (the INI bulk path already purged it via a full-replace `$set`). Fixed by stripping structured keys from the seeded existing bag too, bringing the per-id PrusaSlicer/OrcaSlicer sync to parity with the INI path. Unit + route tests added. (Everything else the sweep traced — the full chamber state-space and the `compatible_printers_condition` round-trip — came back sound across three independent tracers.)

Full tsc / lint / coverage gate green.

@codex review

hyiger:
Round 5: two new findings fixed (`997044de`, incl. a latent bug they exposed); the other three are re-flags of findings already fixed (evidence below).

### ✅ P2 `slicerSettings.ts:58` — settings purge discarded without a new passthrough key — **FIXED (new)**
Correct. `mergeSlicerSettings` now reports purged structured keys in a new `removed[]`, and the OrcaSlicer per-id route writes `update.settings` when `added.length > 0 || removed.length > 0`, so a structured-only sync still persists the cleaned bag. Unit test (`removed` populated) + route test (an orca sync changing only `type` purges a stale `density` shadow while keeping real passthrough).

### ✅ P2 `bambuStudioImport.ts:350` — disabled chamber doesn't clear an existing calibration — **FIXED (new)**
Correct. Added `chamberDisabled` to the parse (distinct from "chamber absent"); the applier now (a) attempts resolution when a chamber disable is present, (b) sets `row.chamberTemp = null` on the matched calibration to clear it, and (c) skips creating an empty row when a disable-only profile matches nothing. So `/calibration` no longer re-enables chamber on the next round-trip. **This also surfaced a latent pre-existing bug**: the bulk/per-id routes pass a HYDRATED Mongoose doc, so `existing.calibrations[i]` are subdocuments whose data lives in `_doc` — `{ ...subdoc }` dropped it, silently corrupting any re-sync that UPDATES an existing calibration row (only reachable with a pre-seeded calibration, which no route test had). Fixed by `.toObject()`-normalizing existing rows before the merge. Tests: disable clears the row (keeps sibling `pressureAdvance`), disable-only-no-match creates no empty row, `chamberDisabled` parse.

### ↩️ P2 `bambuStudioImport.ts:152` — chamber dropped when unresolved — **fixed in `e9146521` (re-flag, 5th)**
The applier preserves the raw chamber keys in the settings bag on unresolved (`bambuStudioApply.ts` chamber block). Green regression test `tests/bambustudio-route.test.ts` → *"an UNRESOLVED chamber_temperature falls back to the settings bag (not dropped)"* asserts `stored.settings.chamber_temperature === "50"`. A standalone import does not drop it.

### ↩️ P2 `bambuStudioApply.ts:416` — `row.chamberTemp` unreachable for chamber-only — **fixed in `e9146521` (re-flag, 5th)**
`wantsResolution = hints.hasAnyHint || hints.chamberTemp != null || hints.chamberDisabled === true`, so a chamber-only profile resolves and reaches the assignment. Green regression test → *"a chamber-ONLY profile with a resolvable printer lands in `calibrations[].chamberTemp`"* asserts `cal.chamberTemp === 48`.

### ↩️ P2 `prusaSlicerBundle.ts:254` — pinned condition lost on collapse — **fixed in `e04322df`/`c21d9f5d` (re-flag)**
The hinted collapse keeps a user pin / `nil` and strips only the exact auto-derived `nozzle_diameter[...]` shape (anchored regex). Green tests: keep-pin, keep-compound-pin, strip-derived, keep-nil, full round-trip.

Full tsc / lint / coverage gate green.

@codex review

hyiger:
Round 6: one new finding fixed (`fabbc001`, a refinement of the r5 disable-clear); the other four are re-flags of findings already fixed.

### ✅ P2 `bambuStudioApply.ts:459` — chamber-disable sync fabricates a calibration from ordinary temps — **FIXED (new)**
Correct, and thanks — this refines the r5 disable-clear. A typical disabled-chamber profile carries nozzle/bed temps, which were copied into `row`, so a disable-only sync that matched no existing row pushed a new per-nozzle calibration (temps + `chamberTemp:null`) that would later shadow user-edited top-level temps. Fix: for a chamber-disable-only sync (explicit disable, no real hint, no new chamber temp) the ordinary parsed temps are NOT copied into `row`, so it (a) only clears an existing row's chamber (temps/EM/PA on that row untouched) and (b) never fabricates a row when nothing matches — the temps still land on the filament top level via `buildStructuredUpdate`. Test: disable-only-with-temps + no existing row → zero calibration rows, `temperatures.nozzle` on the filament, disable marker in the settings bag.

### ↩️ P2 `bambuStudioImport.ts:355` — disabled chamber doesn't clear an existing calibration — **fixed in `997044de`, refined here (re-flag)**
The applier records `chamberDisabled`, resolves on a disable, and sets `row.chamberTemp = null` to clear the matched row. Green test: *"a DISABLED chamber import CLEARS a pre-existing calibrations[].chamberTemp"* (`cal.chamberTemp` null, sibling `pressureAdvance` preserved).

### ↩️ P2 `bambuStudioImport.ts:152` — chamber dropped when unresolved — **fixed in `e9146521` (re-flag, 6th)**
Preserved in the settings bag on unresolved; green test *"an UNRESOLVED chamber_temperature falls back to the settings bag"* (`stored.settings.chamber_temperature === "50"`).

### ↩️ P2 `bambuStudioApply.ts:420` — `row.chamberTemp` unreachable for chamber-only — **fixed in `e9146521` (re-flag, 6th)**
`wantsResolution` includes `chamberTemp != null` and `chamberDisabled`; green test *"a chamber-ONLY profile with a resolvable printer lands in calibrations[].chamberTemp"* (`cal.chamberTemp === 48`).

### ↩️ P2 `prusaSlicerBundle.ts:254` — pinned condition lost on collapse — **fixed in `e04322df`/`c21d9f5d` (re-flag)**
The hinted collapse keeps a user pin / `nil`, strips only the exact auto-derived shape; green tests: keep-pin, keep-compound-pin, strip-derived, keep-nil, round-trip.

Full tsc / lint / coverage gate green.

@codex review

hyiger:
Round 7: one new finding fixed at the ROOT (`0f9c0985`); the other four are re-flags of findings already fixed.

### ✅ P2 `bambuStudioApply.ts:436` — chamber-disable sync updates max-vol on the matched row — **FIXED at the root (new)**
Correct — `maxVolumetricSpeed` is excluded from `hasAnyHint` (it has a top-level home), so a disable-only profile carrying it stays `chamberClearOnly`, yet `row.maxVolumetricSpeed` was populated and would overwrite an existing row's max-vol. This is the third field to slip through the disable-only guard (temps r6, now max-vol r7), so rather than patch another field I fixed it structurally: when `chamberClearOnly`, the row now carries **only** the chamber clear (`chamberTemp: null`) — every non-chamber value (temps, max-vol, EM/PA/retraction/fans) is gated behind `!chamberClearOnly`. Each of those already has its own top-level home via `buildStructuredUpdate`, so nothing is lost, and no field can leak into the calibration row anymore. Test: a disable-only sync carrying max-vol leaves the existing calibration's `maxVolumetricSpeed` untouched (12, not 20) while clearing chamber and landing max-vol=20 on the filament top level.

### ↩️ Re-flags — all fixed, with green regression tests
- **`bambuStudioImport.ts:355`** (disable doesn't clear existing calibration) — fixed in `997044de`; test *"a DISABLED chamber import CLEARS a pre-existing calibrations[].chamberTemp"*.
- **`bambuStudioImport.ts:152`** (chamber dropped when unresolved, 7th) — fixed in `e9146521`; test *"an UNRESOLVED chamber_temperature falls back to the settings bag"* (`=== "50"`).
- **`bambuStudioApply.ts:429`** (`row.chamberTemp` unreachable for chamber-only, 7th) — fixed in `e9146521`; test *"a chamber-ONLY profile with a resolvable printer lands in calibrations[].chamberTemp"* (`=== 48`).
- **`prusaSlicerBundle.ts:254`** (pinned condition lost on collapse) — fixed in `e04322df`/`c21d9f5d`; tests keep-pin / keep-compound-pin / strip-derived / keep-nil / round-trip.

Full tsc / lint / coverage gate green.

@codex review

hyiger:
Round 8: one new finding fixed (`b4239d1f`, a regression in my own sweep purge); the other four are re-flags of findings already fixed.

### ✅ P2 `slicerSettings.ts:69` — per-nozzle sync erased shared calibration defaults from the bag — **FIXED (new)**
Correct — this was a regression in the sweep purge I added. `mergeSlicerSettings` stripped the WHOLE `structuredKeys` set from `existing`, but the Prusa per-id calibration sync lists context-only keys (`extrusion_multiplier`, retraction, fan speeds) in `structuredKeys` — and those have no top-level home, so they can legitimately live in the bag as shared filament-wide defaults. Stripping them erased those defaults on every per-nozzle sync.

Fix: the purge is now narrow. A new `NEVER_BAGGED_KEYS` set = `{filament_settings_id, filamentdb_id, filamentdb_nozzle}` (re-derived-on-export + routing hints — the only keys that genuinely must never persist in the bag and cause a real export bug when stale). `mergeSlicerSettings` purges ONLY those from `existing`, independent of `structuredKeys`. The other structured keys either have a top-level field that overrides a stale bag shadow on export (harmless) or are legit shared defaults (must survive). This keeps the real 950.5 filament_settings_id fix while dropping the over-reach. Tests: a per-nozzle sync preserves a shared `extrusion_multiplier` bag default; the helper purges `filament_settings_id` but not a non-never-baggable structured key.

### ↩️ Re-flags — all fixed, with green regression tests
- **`bambuStudioImport.ts:355`** (disable doesn't clear existing calibration) — fixed `997044de`; test clears `chamberTemp`, keeps sibling PA.
- **`bambuStudioImport.ts:152`** (chamber dropped when unresolved, 8th) — fixed `e9146521`; test `settings.chamber_temperature === "50"`.
- **`bambuStudioApply.ts:425`** (`row.chamberTemp` unreachable for chamber-only, 8th) — fixed `e9146521`; test `cal.chamberTemp === 48`.
- **`prusaSlicerBundle.ts:254`** (pinned condition lost on collapse) — fixed `e04322df`/`c21d9f5d`; keep-pin / keep-compound-pin / strip-derived / keep-nil / round-trip tests.

Full tsc / lint / coverage gate green.

@codex review

hyiger:
Round 9: two new findings fixed (`f586f3b8`), both in the sweep hardening; the other five are re-flags of findings already fixed.

### ✅ P2 `slicerSettings.ts:89` — never-baggable keys not skipped from INCOMING — **FIXED (new)**
Correct consistency gap. The purge removed `NEVER_BAGGED_KEYS` only from `existing`; the incoming loop skipped only `structuredKeys`, so a caller whose structured set omits `filament_settings_id` (the OrcaSlicer per-id sync) would re-add an incoming copy to the bag and re-shadow the re-derived name. Fix: the incoming loop now skips `structuredKeys` OR `NEVER_BAGGED_KEYS`, so never-baggable keys stay out of the bag from BOTH sources regardless of the caller's structured set. Tests: unit (incoming `filament_settings_id` not added even with empty structuredKeys) + route (orca sync with an incoming `filament_settings_id` doesn't persist it).

### ✅ P2 `bambuStudioApply.ts:188` — chamber fallback bypassed the settings cap — **FIXED (new)**
Correct. The unresolved chamber-fallback keys were appended after `mergeSlicerSettings` had already enforced `MAX_SETTINGS_KEYS`, so an at-cap bag + chamber fallback could write 401/402 keys. Fix: the fallback now routes the chamber keys THROUGH `mergeSlicerSettings` (capped), so an over-cap result surfaces as `settingsResult.error` → the route 400s instead of silently over-filling. Test: an at-cap (400-key) filament + unresolved chamber import → 400, saved row unchanged (no chamber key sneaked in).

### ↩️ P2 `slicerSettings.ts:86` — per-nozzle sync erases shared defaults — **fixed in `b4239d1f` (re-flag)**
The purge is now narrow — it removes only `NEVER_BAGGED_KEYS`, NOT the whole `structuredKeys` set, so a shared `extrusion_multiplier`/retraction/fan default in the bag survives a per-nozzle sync. Green test: *"a per-nozzle sync PRESERVES a shared calibration default in the settings bag"* (`extrusion_multiplier === "0.98"` after the sync).

### ↩️ Re-flags — all fixed, with green regression tests
- `bambuStudioImport.ts:355` (disable doesn't clear existing calibration) — fixed `997044de`.
- `bambuStudioImport.ts:152` (chamber dropped when unresolved, 9th) — fixed `e9146521`.
- `bambuStudioApply.ts:425` (`row.chamberTemp` unreachable for chamber-only, 9th) — fixed `e9146521`.
- `prusaSlicerBundle.ts:254` (pinned condition lost on collapse) — fixed `e04322df`/`c21d9f5d`.

Full tsc / lint / coverage gate green.

@codex review

hyiger:
Round 10: one new finding fixed (`a8e1d216`); the other six are re-flags of findings already fixed.

### ✅ P2 `bambuStudioApply.ts:183` — bambu path didn't persist a never-baggable purge — **FIXED (new)**
Correct — the bambu equivalent of the OrcaSlicer `removed` fix (r5). `prepareBambuUpdate` wrote `update.settings` only when `added.length > 0`, so a bambu sync that only updates structured fields on a row with a stale `filament_settings_id`/`filamentdb_id` discarded the cleaned bag and the stale key kept shadowing the re-derived name. Fixed: it now writes when `added.length > 0 || removed.length > 0`, matching the orca route. This closes the last settings-write gate — all three sync consumers (PrusaSlicer per-id writes unconditionally, OrcaSlicer + Bambu honor `removed`) now persist the purge. Test: a bambu structured-only sync on a row with a stale `filament_settings_id` purges it (keeps real passthrough).

### ↩️ Re-flags — all fixed, with green regression tests
- `slicerSettings.ts:89` (skip never-bagged from incoming) — fixed `f586f3b8` (r9); unit + orca-route tests.
- `slicerSettings.ts:86` (per-nozzle erases shared defaults) — fixed `b4239d1f` (r8, narrow purge); test preserves a shared `extrusion_multiplier`.
- `bambuStudioImport.ts:355` (disable doesn't clear existing calibration) — fixed `997044de`.
- `bambuStudioImport.ts:152` (chamber dropped when unresolved, 10th) — fixed `e9146521`.
- `bambuStudioApply.ts:434` (`row.chamberTemp` unreachable for chamber-only, 10th) — fixed `e9146521`.
- `prusaSlicerBundle.ts:254` (pinned condition lost on collapse) — fixed `e04322df`/`c21d9f5d`.

Full tsc / lint / coverage gate green.

@codex review

hyiger:
Round 11: one new finding fixed (`2cc9ada4`); the other seven are re-flags of findings already fixed.

### ✅ P2 `bambuStudioApply.ts:442` — chamber-ENABLED-only sync pinned ordinary temps into the calibration — **FIXED (new)**
Correct — the enabled-chamber twin of the r6/r7 disable cases. A chamber-only profile has `hasAnyHint === false` but, when chamber is enabled, `chamberTemp != null`, so my `!chamberClearOnly` gate still copied the profile's ordinary nozzle/bed temps (and max-vol) into `calibrations[]`, fabricating a per-nozzle override from top-level-homed settings.

Fixed at the root: the non-chamber fields are now gated on `hasAnyHint` (a real per-nozzle hint), NOT on "chamber present". The chamber value alone goes on the row; ordinary temps/max-vol are copied only for a genuine calibration write. This unifies the enabled + disabled chamber-only cases (I removed the `chamberClearOnly` construct entirely) — chamber still lands in `calibrations[].chamberTemp` (its structured home), but no ordinary top-level-homed setting is ever pinned into a per-nozzle row. Test: a chamber-enabled-only sync → `cal.chamberTemp === 48`, `cal.nozzleTemp`/`cal.bedTemp` null, temps on the filament top level.

### ↩️ Re-flags — all fixed, with green regression tests
- `bambuStudioApply.ts:188` (bambu persist purge) — fixed `a8e1d216` (r10).
- `slicerSettings.ts:89` (skip never-bagged from incoming) — fixed `f586f3b8` (r9).
- `slicerSettings.ts:86` (per-nozzle erases shared defaults) — fixed `b4239d1f` (r8).
- `bambuStudioImport.ts:355` (disable doesn't clear existing calibration) — fixed `997044de`.
- `bambuStudioImport.ts:152` (chamber dropped when unresolved, 11th) — fixed `e9146521`.
- `bambuStudioApply.ts:439` (`row.chamberTemp` unreachable for chamber-only, 11th) — fixed `e9146521`.
- `prusaSlicerBundle.ts:254` (pinned condition lost on collapse) — fixed `e04322df`/`c21d9f5d`.

Full tsc / lint / coverage gate green.

@codex review

PR #970, opened 2026-07-04
fix(nfc): OpenPrintTag encode keeps the everyday temp in MAX (closes #952)

hyiger:
Fixes #952 — the last remaining sub-finding (952.4; items 1/2/3/5 shipped in #967, and a re-triage confirmed those still hold).

## 952.4 — first-layer temperature round-trip inflation
The OpenPrintTag encoder anchored the everyday temp to one range bound and derived the other from the first-layer temp, then `Math.min/max`-ordered the pair (GH #634, to keep `MIN ≤ MAX`). When the first-layer temp was the **hotter** one — the *normal* slicer convention for the bed — the ordering **swapped** everyday↔first-layer, and since the decoder reads the primary temp from `MAX`, the app's own NFC write→scan path decoded the everyday bed **inflated** by the first-layer delta (`bed 85 / first-layer 90` → decoded `bed 90`), with the first-layer value lost. Same drift on the nozzle when `firstLayer > nozzle + 20`.

**Fix (encode-only, per the maintainer's call):** always write the **everyday** temp into `MAX_BED` / `MAX_PRINT`; derive `MIN` downward only (`min(everyday, first-layer-derived)`), so `MIN ≤ MAX` still holds. The decoder is unchanged — it reads the everyday temp from `MAX` and now gets the right value on newly-written tags.

Trade-offs (accepted, documented in the code):
- A first-layer temp *above* the everyday temp can't extend a pinned `MAX`, so its excess isn't stored — but first-layer was **already** dropped on decode (OpenPrintTag has no first-layer field), so this loses nothing new; it just makes the *everyday* (primary) temp correct.
- OpenPrintTag has **no version field** in the codec, so tags written before this fix can't be distinguished and keep the old (swapped) reading. This corrects newly-written tags only. (Confirmed: no version gate exists to support dual-scheme reads.)
- The no-first-layer case is unchanged: the encoder still synthesizes `MIN = everyday-10` (bed) / `everyday-20` (nozzle), and `MAX` already held the everyday temp there — a decode-only "read from MIN" fix would have regressed exactly this case, which is why the fix is on the encode side.

## Tests
New `openprinttag-decode` round-trip block covering bed first-layer-hotter / cooler / none and nozzle first-layer>+20 / none (everyday round-trips, no-first-layer preserved). Updated the mechanism-note tests in `openprinttag-decode`, `openprinttag` (#634), and `nfc-roundtrip` that pinned the old inflated read. Full tsc / lint / coverage gate green.

### DIFF
--- a/src/lib/openprinttag.ts
+++ b/src/lib/openprinttag.ts
@@ -716,14 +716,22 @@ function buildMainMap(input: OpenPrintTagInput): number[] {
   // before it reaches the encoder so a bad input can't crash the whole
   // NFC-write path.
   if (input.nozzleTemp != null) {
-    // Use nozzle temp as max, derive min as temp - 20.
-    // GH #634: a first-layer temp ≥ nozzle+20 made the derived min exceed
-    // the max — order the pair with Math.min/Math.max exactly like the
-    // bed temps below, and derive preheat from the effective min.
+    // GH #952.4: the EVERYDAY nozzle temp must ALWAYS land in MAX_PRINT — the
+    // decoder reads it back from MAX as the primary nozzleTemp (openprinttag-
+    // decode.ts). GH #634 derived MIN from the first-layer temp and Math.max-ordered
+    // the pair, which SWAPPED them when firstLayer > nozzle+20: the everyday temp
+    // fell into MIN and the decoder read the inflated first-layer value as everyday.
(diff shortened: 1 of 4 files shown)


hyiger:
@codex review

PR #974, opened 2026-07-04
fix(import): clear parent-equal variant pins on round-trip (scalar/temp/array) (#971)

hyiger:
Fixes #971.

## Problem

`splitInheritedImportSet` (shared by the CSV/XLSX importer, the per-id PrusaSlicer sync route, and the bulk INI importer) cleared a variant's parent-equal override **only when it diverged** from the incoming value. A bundle/CSV export — and the PrusaSlicer fork's per-id sync — flatten a variant through `resolveFilament`, so a deliberate *pin* and a true *inherit* serialize to the identical value. When the variant's stored value already **equalled** the parent, the divergence gate (`variantVal !== incoming`) was false, so no `$unset` fired and the pin survived — silently blocking future parent edits from propagating (GH #106). #969 fixed this for the settings bag; this extends the same presence-based clear to the three scalar/array branches.

## Fix

Presence-based `$unset` (clear any local override of a field the section reports equal to the parent), in `splitInheritedImportSet`:
- **scalar** (`IMPORT_INHERITABLE_SCALARS`): drop the `variantVal !== incoming` gate — the `IMPORT_REQUIRED_FIELDS` (vendor/type) guard above it is unchanged, so those are still never `$unset`.
- **temperatures.***: `$unset` on `variantVal != null`.
- **secondaryColors**: `$unset` any non-empty variant-local array when the incoming array equals the parent's.

One shared helper → all three call sites pick it up.

## Why this is correct on ALL three paths (incl. per-id sync)

The per-id PrusaSlicer sync route (`POST /api/filaments/[id]`) receives config the fork produces by flattening the variant through `resolveFilament` (see the route docblock), so a parent-equal incoming value is **indistinguishable from an inherit** there exactly as on the CSV/INI bundle paths. Presence-based clearing is therefore the correct *shared* default, not a bundle-only concern. Added a code comment stating this invariant at the call site.

## Tests
- Unit (`importFilaments.test.ts`): parent-equal `$unset` for scalar + temperature; flipped the secondaryColors parent-equal test to expect `$unset` (both arms of the `variantArr.length > 0` gate covered).
- Route round-trip: bulk INI (`filaments-import-export-route.test.ts`) and per-id sync (`api-route-correctness.test.ts`) each seed a parent-equal pin, sync/import, and assert it clears + a later parent edit propagates.
- Full coverage gate green (98.99 / 97.44 / 97.49 / 99.51).

## Out of scope (tracked)
The per-id sync route does not yet clear parent-equal **settings**-bag pins (`settingsSelfHealUnset` is INI-only) — a separate pre-existing gap tracked in #972.

### DIFF
--- a/src/app/api/filaments/[id]/route.ts
+++ b/src/app/api/filaments/[id]/route.ts
@@ -948,11 +948,17 @@ export async function POST(
     // this, only `color` had echo suppression (resolveSyncBackColor). Reuse the
     // CSV importer's battle-tested split (GH #628 / #649): drop each inheritable
     // field whose incoming value equals the parent's (keep inheriting), and
-    // $unset a stale diverging local override so inheritance resumes. Variant-
+    // $unset a stale local override so inheritance resumes. Variant-
     // only + non-inheritable keys (color, colorName, name, settings, soluble,
     // …) pass through untouched. `calParent` is the already-fetched parent doc;
     // when it's null (standalone/parent, or a soft-deleted/missing parent —
     // nothing to inherit) the update is written verbatim.
+    // GH #971: because the fork ALWAYS sends resolveFilament-flattened values
+    // (per the docblock above), a parent-EQUAL incoming value is indistinguishable
+    // from a true inherit on this path exactly as on the CSV/INI bundle paths — so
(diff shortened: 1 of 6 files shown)


hyiger:
@codex review

PR #981, opened 2026-07-05
feat(nfc): remember a default NTAG type so batch writes skip the size prompt (#973)

hyiger:
Follow-up to #973 / v1.64.1. On readers that reject `GET_VERSION` (e.g. the ACR1552U), the NTAG size picker fired on **every** write — painful for batch writing. This makes the choice **configurable** so you set it once.

## What's new
- **"Remember this — don't ask again" checkbox** in the size picker → saves your pick as the default; later writes skip the prompt.
- **Settings → Devices → NFC Tools → "Default NTAG type"** selector (Ask each time / NTAG213 / NTAG215 / NTAG216) to view/change/clear it.
- Persisted via the electron-store / localStorage bridge (mirrors the currency / label-format settings), so it survives restarts.

## Implementation
- New `src/hooks/useNtagDefaultSize.ts` (module-store + `useSyncExternalStore`, `"ask" | NTAG213/215/216`).
- **Electron IPC wiring for `ntagDefaultSize`** in `get-config` / `save-config` (+ `preload` + `electron.d.ts` types, no cast) — without this the setting would silently not persist on the desktop app.
- The write flow uses the saved default silently when set; only `"ask"` prompts. The picker resolves `{ size, remember }`.

## Safety
Unchanged: the `FF B0` brick-guard probe still validates the physical chip size before any write, so a wrong default (too large) is **refused, not bricked**.

## Verification
Root `tsc`, `electron:typecheck`, full lint, i18n parity, and full coverage (3145 tests) pass. Renderer/hook + electron IPC aren't CI-exercised (hardware/desktop-owed).

## Pre-review hardening
An adversarial pass caught (and this includes the fix for) a **P1**: the initial commit added the hook but not the electron `get-config`/`save-config` branches, so the default would have silently reset every restart on the desktop app — now wired end-to-end, with the `as Record<string,string>` cast dropped so tsc guards the key.


hyiger:
@codex review

Issue #446, opened 2026-05-30
i18n: dates use toLocaleDateString() with no locale arg — German users see English dates


hyiger:
## Summary
Every date display in the app calls `toLocaleDateString()` / `toLocaleString()` with no locale argument. Output follows the browser locale, not the app's selected i18n locale. A user on a German app locale with an English browser sees English dates everywhere despite picking DE.

## Evidence
- `src/app/trash/page.tsx:191-193`
- `src/app/dashboard/page.tsx:87, 205, 228`
- `src/app/inventory/page.tsx:592`
- `src/components/SyncStatusIndicator.tsx:273`

## Suggested fix
Read `locale` from `useTranslation()` and pass it explicitly. Centralise in a small helper:

```ts
export function formatDate(iso: string, locale: string) {
  return new Date(iso).toLocaleDateString(locale === "de" ? "de-DE" : "en-US");
}
```

Use the helper at every callsite.

Filed from full-codebase review on 2026-05-29.

PR #984, opened 2026-07-05
feat(settings): user-selectable date format (#983)

hyiger:
Closes #983 — a Settings control to choose how dates render across the app. A UK user currently gets US-ordered `M/D/YY` because the app ships only `en`/`de` locales and `Intl("en", {dateStyle:"short"})` is US-ordered; this gives every user explicit control.

## What you get (Settings → UI → Date format)
- **System (device region)** — follows the OS/PC regional setting (the issue's "follow PC system setting"). Default.
- **ISO** `YYYY-MM-DD`, **US** `MM/DD/YYYY`, **European** `DD/MM/YYYY` — locale-independent fixed formats.
- **Custom** — a numeric-token pattern (`YYYY` / `YY` / `MM` / `M` / `DD` / `D` + any literal separators), e.g. `DD-MM-YY`. Live preview + inline validation; an invalid/token-free pattern falls back to system and never renders garbage.

## How it works
- **`src/lib/dateFormatPref.ts`** (pure, DB-free, **100% line/branch coverage**): preference shape, validation, preset→pattern resolution, and a deterministic single-pass token formatter. Field values come from one fixed-locale `Intl…formatToParts` so numeric output is stable ASCII digits, and `timeZone` threads through so UTC calendar-days (spool dates, analytics day-keys) never shift ±1 day.
- **`src/lib/dateFormat.ts`**: `formatDate`/`formatDateTime` gain an optional `pattern`. A null/absent pattern preserves the existing locale-aware `Intl` path — **every existing caller is unchanged**. `formatDateTime` patterns the date part; time stays locale-short (the preference is date-only).
- **`src/hooks/useDateFormat.ts`**: persists to electron-store `dateFormat` / localStorage `filamentdb-date-format` via a module store + `useSyncExternalStore` (mirrors `useLabelFormat`). A **`ready` gate** keeps the first SSR + hydration render byte-identical to today's output; the preference (incl. device-region `system` via `navigator.language`) only applies post-mount — **no React 19 hydration mismatch**.
- All **13 date renders across 9 files** migrated to the hook, preserving every `{ timeZone: "UTC" }` and the analytics/usage-history memo dependencies.
- i18n: 12 new `settings.dateFormat.*` keys in **en + de** (parity gate); `settings.group.ui.desc` updated.

## ⚠️ Interaction with #446 (needs a look)
`system` mode = **device/OS region**, and it's the default. #446 deliberately made dates follow the **app language** (so a German-app/English-OS user saw German dates). For users whose OS locale differs from their app language, this changes the *default* rendering back to OS region — which is exactly what #983 asks for. Most users (OS locale == app language) see no change, and anyone can pick an explicit preset. Flagging in case you'd rather keep app-language as the default and make device-region opt-in.

## Verification
- Full suite green — **2979 tests**; new pure module at **100% coverage**; `tsc`, `eslint` clean.
- In-browser (dev): preset switching, custom pattern + live preview, invalid-pattern validation + safe fallback, persistence round-trip across reload, and **no console/hydration warnings**.


hyiger:
@codex review

hyiger:
@codex review

hyiger:
Addressed both findings in e883d4ea:

- **P1 (Electron config bridge)** — wired `dateFormat` through `electron/main.ts` (`get-config`/`save-config`) and `electron/preload.ts`, mirroring `labelFormat`. Desktop preference now persists across restart. `electron:typecheck` passes.
- **P2 (custom pattern hydration)** — `DateFormatSection` now derives the input from `config.pattern` instead of a one-shot `useState` initializer, so a directly-loaded saved custom format hydrates correctly (verified: reload restores `DD-MM-YYYY`, no false error).

@codex review

hyiger:
Fixed in 0b63e92d: the invalid/empty custom-pattern fallback now uses the device region (keyed on `pattern === null`, covering both `system` mode and the null-pattern custom fallback) instead of the app locale — so an invalid pattern can no longer reintroduce app-locale date ordering. tsc + eslint + date/i18n tests green.

@codex review

hyiger:
@codex review

Issue #678, opened 2026-06-12
Bambu/Orca import collapses multi-element settings arrays to their first element (compatible_printers round-trip data loss)


hyiger:
**Severity:** P3 · **Area:** `lib-resolve-slicer` · **Category:** data-integrity
**Location:** `src/lib/bambuStudioImport.ts:158-169 (unwrap), 313-371`

### Problem
`unwrap()` returns only `String(value[0])` for any array value. Single-element arrays (the OrcaSlicer/Bambu convention for most filament keys) round-trip fine, but genuinely multi-valued keys lose every element after the first. The concrete case is `compatible_printers`, which in real Bambu/Orca filament presets is a multi-entry array (a preset compatible with several printer/nozzle combos, e.g. `["Bambu Lab P1S 0.4 nozzle", "Bambu Lab X1C 0.4 nozzle", ...]`). `compatible_printers` is in neither STRUCTURED_KEYS nor CALIBRATION_KEYS, so it falls into the settings passthrough loop (line 365-371) and is stored as only its first element via `unwrap`. The same first-element-only collapse also feeds `calibrationHints.compatiblePrinters` (line 317). On a later export through `filamentToOrcaSlicerKeys` (which re-emits the settings bag verbatim), the re-exported preset is now compatible with only one printer instead of the original set. The round-trip guarantee in the file header ('export → import → export produces the same JSON') is violated for any multi-valued passthrough key. Practical impact is limited because Filament DB intentionally exports an empty `compatible_printers` on its own bundle paths, but the per-id Bambu import → settings → Orca/Bambu re-export path silently drops the extra entries.

### Suggested fix
For passthrough settings keys, preserve the original array shape instead of collapsing via `unwrap`. In the settings loop (line 365-371) store the raw array (stringified element-wise) when `Array.isArray(value) && value.length > 1`, mirroring how `filamentToOrcaSlicerKeys` already handles `Array.isArray(value) ? value.map(String) : [String(value)]` on the export side. Alternatively add `compatible_printers` to an explicit multi-value allowlist that bypasses `unwrap`.

---
<sub>Filed from a full automated source audit. Each finding was independently re-verified against the code before filing; this one's verification: CONFIRMED REAL and CURRENT. Re-read src/lib/bambuStudioImport.ts and src/lib/orcaSlicerBundle.ts.

Mechanism verified:
[...]

PR #985, opened 2026-07-05
feat(import): bulk OrcaSlicer filament-library import (relates to #341)

dcode:
## Summary

Replaces the deliberately-unguarded `POST /api/filaments/orcaslicer` 501 stub
with a real bulk importer for OrcaSlicer filament-library folders, plus a UI
entry point on `/import-export`. The import is idempotent, allowing updates to
previously imported profiles by reimporting them.

The import process only instantiates a filament profile where `"instantiation" == "true"`
and it flattens all ancestors where this condition is false. Selecting a top-level OrcaSlicer
directory will present both the system presets and the user profiles and will properly
import inherited concrete profiles as variants of the parent.

> [!NOTE]
> @hyiger, if this workflow doesn't align with your preferences, I'm happy to make adjustments, as needed.

{1A1208F2-9C38-4507-BCF6-02F7A8EE9004}
{2ECC3C8F-F1A7-4D0B-ADEE-47FABEEEC554}

---

- **Planner** (`src/lib/orcaSlicerImport.ts`, DB-free): indexes raw presets by
  name, classifies `instantiation:"false"` abstract templates vs concrete
  profiles, resolves/flattens `inherits` chains (child key wins, identity
  keys never inherited), and plans the import onto Filament DB's one-level
  variant model — each selection's root concrete ancestor becomes a parent
  record (templates baked in, auto-imported even if unselected); the
  selection itself becomes a variant storing only its raw-key diffs vs the
  flattened parent. The bed-plate (`bedTypeTemps[]`) and calibration
  (`calibrations[]`) key groups diff atomically since they inherit as whole
  arrays. 3+-level concrete chains collapse to the concrete root.
- **Shared upsert** (`src/lib/bambuUpsert.ts`): extracted the bambustudio
  route's three-phase active → resurrect-trashed →
  create-with-E11000-race-recovery upsert into a shared lib so the new route
  can reuse the exact same write path instead of duplicating it. `parentId`
  applies on the phase-3 create ONLY — updates never re-parent an existing
  row.
- **Route** (`POST /api/filaments/orcaslicer`): body `{selected, profiles}`
  (the chosen presets plus their full ancestor closure). CSRF-guarded,
  Content-Length + buffered-byte 413s, 10k-profile cap. Collision handling:
  same-parent variant → idempotent diff update; existing root → full-payload
  in-place update, never re-parented; variant of a *different* parent →
  per-profile error rather than silently severing its inheritance.
  Per-profile failures never sink the batch.
- **UI**: `OrcaLibraryImportDialog` on `/import-export` — folder pick via
  `webkitdirectory`, client-side parsing, a searchable picker with
  chain-aware vendor/material filters, ancestor-closure collection, and a
  created/updated/variants/error summary on submit.
- **Docs**: `public/openapi.json`, `docs/api.md`, `docs/de/api.md`, and
  `CLAUDE.md` all updated; the CSRF-sweep doc's stale "one stub left
  unguarded" note is removed now that the route is guarded.

## Test plan

- [x] `npm run lint`
- [x] `tsc --noEmit`
- [x] `npm run electron:typecheck`
- [x] Full suite green (`tests/orcaSlicerImport.test.ts` — pure planner,
      `tests/orcaslicer-import-route.test.ts` — route w/ mongodb-memory-server,
      `tests/bambuUpsert.test.ts` — race/failure branches the routes can't
      reach directly, plus a `destructive-route-guard.test.ts` addition)
- [x] `npm run test:coverage -- --maxWorkers=4` passing all thresholds
- [x] Manually exercised end-to-end in the dev server (light + dark): folder
      scan → picker → import (created parent + variant) → idempotent
      re-import → export round-trip with effective values → cross-origin
      request correctly rejected (403)

## Backlog (non-blocking, noted for follow-up)

- Variant re-import can't un-pin a settings-bag key that later becomes equal
  to the parent's (`mergeSlicerSettings` is additive) — matches existing
  INI-import semantics; revisit if it bites users.
- Multi-element `compatible_printers` arrays collapse to their first element
  in the settings bag (pre-existing #678); the library importer inherits
  that limitation.
- Picker polish: consider rendering only the filtered slice if a
  full-library folder (~4k profiles) makes the list sluggish.

## Open questions

- None blocking. Two behaviors are deliberate design decisions rather than
  bugs — flagging in case maintainers want different semantics:
  - Collision posture: an existing variant of a *different* parent is
    skipped with a per-profile error rather than being reassigned.
  - Resurrecting a trashed filament keeps its original `parentId` rather
    than adopting the import's target parent.


hyiger:
@codex review

hyiger (commented):
### 🔍 Claude Code review (multi-agent, Codex-style)

Automated adversarial review across 6 dimensions (correctness, security, data-integrity, API/OpenAPI, tests, conventions); each finding was independently verified against the code before posting. **7 raw findings → 5 after dedupe + refutation** (1 refuted, 2 merged).

**Summary:** 1 × P2 — a variant-import against a *trashed* same-name row resurrects an unlinked, diff-only orphan — plus 4 × P3 (a docs/runtime `errors[]` format mismatch and three test-coverage gaps clustered on the resurrect/variant path). **No P1s:** the CSRF guard is present (`route.ts:161`) and cross-site-rejection is tested; the upsert strips `spools`/`usageHistory`/`dryCycles` (subdoc protection holds); bad-body → 400 and mass-assignment handling check out. Nice work overall — the flattening/instantiation logic and the three-phase upsert are solid; the P2 is a real edge where the variant path and the trashed-row resurrect collide.

Details inline. Useful? React with 👍 / 👎.

src/app/api/filaments/orcaslicer/route.ts
```diff
@@ -105,23 +113,207 @@ export async function GET(request: NextRequest) {
...
+          // the create path, which links correctly).
+          const existing = (await Filament.findOne({
+            name: entry.name,
+            _deletedAt: null,
+          })
+            .select("parentId")
+            .lean()) as { parentId?: unknown } | null;
+          if (!existing) {
+            // No active row: create as a variant carrying only its diffs.
+            rawPayload = entry.diffRaw!;
```
hyiger, line 318: **<sub><sub>P2 Badge</sub></sub>  Variant import onto a trashed same-name row resurrects an orphaned, diff-only filament**

When a plan entry is a **variant** and there is no *active* same-name row (`findOne({ name, _deletedAt: null })`, ~L244), the route takes the create-variant branch: `rawPayload = entry.diffRaw` (diff-only keys) + `parentId: createParentId`. But `upsertParsedBambuFilament` first runs its **phase-2 resurrect** of any *trashed* (non-purged) same-name row — and phase 2 **ignores `opts.parentId`** (it's only applied on the phase-3 `Filament.create`, `src/lib/bambuUpsert.ts:205-209`). So the trashed row is resurrected with **(a)** no `parentId` link — it isn't actually a variant and inherits nothing from the intended parent — and **(b)** only the variant's diff keys as `$set`, so every field meant to be inherited (density, temps, …) is missing/stale. It's silently reported as `updated`. The route comment (~L285) acknowledges a resurrect keeps the trashed row's old parent state, but only compensates the `variants` *count*, not the payload/parent choice. (Found independently by the correctness and data-integrity reviewers; verified against the code.)

**Suggested fix:** On the variant branch, when a trashed same-name row will be resurrected, send the **full flattened** payload and set `parentId` on the resurrect path too (or refuse with a clear error) — don't hand a diff-only payload into a resurrect that can't apply the parent link.

Useful? React with 👍 / 👎.

docs/api.md
```diff
@@ -467,6 +468,41 @@ Query parameters:
...
+Returns:
+```json
+{
+  "created": 2,
+  "updated": 0,
+  "variants": 1,
+  "filaments": ["Generic PLA @System", "Polymaker PolyLite PLA @System"],
+  "calibrationApplied": 0,
+  "calibrationUnresolved": 0,
+  "errors": ["Orphan PLA: inherits \"missing_base\" not found in the submitted set — include the base profile"]
```
hyiger: **<sub><sub>P3 Badge</sub></sub>  Documented errors[] example doesn't match runtime output (quoted vs unquoted name)**

The `errors[]` example in `docs/api.md`, `docs/de/api.md` (~L480) and the OpenAPI description renders a plan failure as `Orphan PLA: inherits "missing_base" not found…` (name **unquoted**). But `planOrcaImport`/`indexOrcaProfiles` prefix the profile name in **double quotes**, so the real string is `"Orphan PLA": inherits "missing_base" not found…` (pinned by `tests/orcaslicer-import-route.test.ts:270`). Per-entry loop errors in the route (~L236/265/278/298) use the *unquoted* `name:` form, so the two error sources also disagree with each other. A client parsing `errors[]` by the documented shape mis-extracts the profile name.

**Suggested fix:** Make the docs/OpenAPI example match the runtime (quoted name), and ideally unify the plan-level and per-entry error prefixes to one format.

Useful? React with 👍 / 👎.

src/app/api/filaments/orcaslicer/route.ts
```diff
@@ -105,23 +113,207 @@ export async function GET(request: NextRequest) {
...
+  if (sizeError) return sizeError;
+
+  try {
+    await dbConnect();
+
+    const body = await request.text();
+    // Byte length, not String.length (UTF-16 code units) — a non-ASCII UTF-8
+    // body can exceed 10 MB of bytes while staying under the char count when
+    // Content-Length was missing/wrong (Codex P2 on PR #685).
+    if (Buffer.byteLength(body, "utf8") > MAX_UPLOAD_SIZE) {
```
hyiger, line 195: **<sub><sub>P3 Badge</sub></sub>  413 byte-length branch and indexOrcaProfiles error-surfacing have no route test**

Two advertised behaviors aren't pinned by `tests/orcaslicer-import-route.test.ts`: **(1)** the post-read `Buffer.byteLength(body) > MAX_UPLOAD_SIZE` → 413 branch (L174-176) — the only 413 test sends `content-length: 11MB`, which `checkContentLength` (L164) rejects *before* the body is read, so the missing/under-declared-Content-Length path this branch exists for is never exercised; **(2)** `indexOrcaProfiles` errors (duplicate/malformed profile in the POSTed array) are merged into the response at ~L216, but no route test submits such input, so the index-errors→response surfacing is only unit-tested on the pure helper, never through the route.

**Suggested fix:** Add a route test that POSTs an oversized body with no/incorrect Content-Length (hits the byte-length 413) and one with a duplicate/malformed profile (asserts the merged `errors` surface in the response).

Useful? React with 👍 / 👎.

src/lib/bambuUpsert.ts
```diff
@@ -0,0 +1,306 @@
...
+    _deletedAt: { $ne: null },
+    _purged: { $ne: true },
+  });
+  if (existingTrashed) {
+    const payload = await prepareBambuUpdate(
+      parsed,
+      await augmentExistingWithParent(existingTrashed),
+    );
+    const payloadFailure = bambuPayloadFailure(payload);
+    if (payloadFailure) return payloadFailure;
```
hyiger, line 178: **<sub><sub>P3 Badge</sub></sub>  Phase-2 resurrect and E11000 race payload-failure returns are untested (gated file)**

`bambuUpsert.ts` is under the `src/lib/**` 99%/96% coverage gate. Two 400-producing branches are never exercised: the `if (payloadFailure) return payloadFailure;` guard on the **phase-2 resurrect** path (~L152) and the analogous `racePayloadFailure` return on the **E11000 race-recovery** path (~L231). `tests/bambuUpsert.test.ts` hits phase-2/race only with valid payloads. The `?? null` fallbacks + parentId ternary in `augmentExistingWithParent` (~L273-279) are also uncovered.

**Suggested fix:** Add a resurrect and a race test whose recomputed payload trips `bambuPayloadFailure` (e.g. an inverted nozzle range / over-cap settings bag) to pin both returns and the augment fallbacks.

Useful? React with 👍 / 👎.

tests/orcaslicer-import-route.test.ts
```diff
@@ -0,0 +1,386 @@
...
+    expect(res.status).toBe(200);
+    const body = await res.json();
+    expect(body.created).toBe(0);
+    expect(body.updated).toBe(1);
+
+    const rows = await Filament.find({ name: GENERIC.name });
+    expect(rows).toHaveLength(1);
+    expect(rows[0]._deletedAt).toBeNull();
+    expect(rows[0].vendor).toBe("Generic");
+  });
```
hyiger, line 668: **<sub><sub>P3 Badge</sub></sub>  No test resurrects a trashed row that collides with an intended VARIANT**

The resurrect test (L194-213) only covers a trashed row colliding with a **root** import (Generic), where the diff-vs-full / parent-link concern can't arise. There is no test where a trashed row's name matches a profile the plan classifies as a **variant** — which is exactly the branch that produces the orphan in the P2 finding above. Because the phase-2 path is still hit by the root-resurrect test, instrumentation coverage can stay green while the semantically-broken variant-resurrect interaction is entirely unverified.

**Suggested fix:** Add a test that trashes a row named like a planned variant, imports the library, and asserts the resurrected row is either correctly linked+fully-flattened or refused (not a sparse, unlinked orphan).

Useful? React with 👍 / 👎.

src/app/api/filaments/orcaslicer/route.ts
dcode: Fixed in 261c92d. The advisory pre-check now also looks up a trashed non-purged row when no active row matches (active-first, mirroring the upsert's own phase order), and applies the same collision tree: a trashed variant of the same parent gets the diff payload (phase 2 keeps its parentId, so the link survives resurrection); a trashed root gets the FULL flattened payload so the resurrect can't strand stale/missing inherited fields; a trashed variant of a different parent is skipped with a per-profile error. Pinned by the new "trashed-row collisions (PR #985)" suite in `tests/orcaslicer-import-route.test.ts`.

src/lib/orcaSlicerImport.ts
```diff
@@ -0,0 +1,515 @@
...
+ *     variant at the gray default, not at the parent's color.
+ *   - `filament_id` / `setting_id` — the vendor-catalog identity of THIS
+ *     preset; inheriting the parent's via the settings bag would mislabel
+ *     the variant.
+ */
+const DIFF_ALWAYS_KEEP = new Set<string>([
+  "name",
+  "filament_type",
+  "filament_vendor",
+  "filament_colour",
```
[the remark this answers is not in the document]
dcode: Fixed in 261c92d. filament_diameter is now in DIFF_ALWAYS_KEEP (with a comment naming the schema's default: 1.75 trap), so a variant doc always carries its diameter instead of getting the wrong pinned default. On re-import of an existing variant the GH [#403](https://github.com/dcode/filament-db/issues/403) machinery still $unsets a parent-equal value, so live inheritance isn't lost where it's possible. Tests: a diffOrcaRaw unit case (2.85 chain keeps the key) and a route test asserting the created variant's diameter is 2.85, not 1.75.

src/app/api/filaments/orcaslicer/route.ts
```diff
@@ -105,23 +113,207 @@ export async function GET(request: NextRequest) {
...
+        if (result.created) created++;
+        else updated++;
+        // A resurrect keeps the trashed row's old parent state, so only a
+        // fresh create or a confirmed same-parent update counts as a variant.
+        if (intendsVariant && (result.created || createParentId === null)) {
+          variants++;
+        }
+        names.push(result.doc.name);
+        if (entry.kind === "root") {
+          parentIds.set(entry.name, String(result.doc._id));
```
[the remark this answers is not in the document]
dcode: Fixed in 261c92d. Root entries now run the same advisory pre-check before any write: if the same-name row (active, or trashed non-purged — the trashed path had the identical hole via phase-2 resurrect) carries a parentId, the entry is skipped with "Name": already exists as a variant of another filament — skipped and its planned children fail through the existing parent failed to import path, so a variant's _id can never be registered in parentIds. Route tests cover both the active-variant and trashed-variant collision, asserting no mutation and no children created.

docs/api.md
dcode: Fixed in 261c92d. Runtime prefixes are unified to the planner's double-quoted form — all four errors.push sites in the route now emit "Name": … — and the errors[] example in both docs (plus the OpenAPI errors description) shows the quoted form. Both docs' collision bullets were also updated for the new trashed/root-vs-variant behavior.

src/app/api/filaments/orcaslicer/route.ts
dcode: Fixed in 261c92d. Added: (a) a body with a lying Content-Length: 100 header that is 6M chars but 12 MB of UTF-8 bytes → 413 via the post-read Buffer.byteLength branch only (also pins the #685 byte-not-char semantics); (b) a payload with a duplicate-name profile, a non-object entry, and a nameless object → all three indexOrcaProfiles messages surface in errors[] while the valid selection still imports.

src/lib/bambuUpsert.ts
dcode: Fixed in 261c92d. New tests in `tests/bambuUpsert.test.ts` hit both pre-write bambuPayloadFailure returns via the GH #892 inverted-range guard, chosen because it's phase-specific: an incoming min of 300 is only inverted when merged with an existing doc's stored max of 250 — a trashed row for the phase-2 gate, the E11000 racing winner for the race-recovery gate (the create gate, computed against null, passes). Both assert {ok:false, status:400} and that no write happened. A third test feeds phase 1 a minimal plain-object doc (missing optional fields) to cover the `?? null` fallbacks in `augmentExistingWithParent`, which are unreachable through a hydrated Mongoose doc.

tests/orcaslicer-import-route.test.ts
dcode: Fixed in 261c92d. This is the regression pin for finding 1: the new suite trashes a row named like a planned variant and asserts all three branches — same-parent → diff resurrect with the link intact and cost still inheriting; root-shaped → full-payload resurrect with inherited-only fields pinned and `parentId` null; different-parent → per-profile error with the row left trashed and untouched.

hyiger:
@codex review

src/app/api/filaments/orcaslicer/route.ts
```diff
@@ -105,23 +113,244 @@ export async function GET(request: NextRequest) {
...
+            // No same-name row at all: create as a variant carrying only
+            // its diffs.
+            rawPayload = entry.diffRaw!;
+            createParentId = parentDocId;
+            intendsVariant = true;
+          } else if (existing.parentId && String(existing.parentId) === parentDocId) {
+            // Variant of the same parent — active (diff update) or trashed
+            // (diff resurrect; phase 2 keeps its parentId, so the link
+            // survives): idempotent re-import either way.
+            rawPayload = entry.diffRaw!;
```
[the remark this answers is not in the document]
dcode: Fixed in ccd788b, with one deliberate deviation from the suggested fix: the parent-equality pruning only covers inheritable scalars (and now temperatures), so feeding the full flattened profile would pin parent-equal settings-bag / bed-plate / calibration values onto the variant instead of clearing anything. The update payload is therefore the diff plus the parent-equal prunable keys (PRUNABLE_RAW_KEYS), which is exactly what lets the machinery in #403  $unset the stale override; fresh creates keep the bare diff. Regression-tested at both the unit and route level.

src/lib/orcaSlicerImport.ts
```diff
@@ -0,0 +1,576 @@
...
+  );
+
+  const out: Record<string, unknown> = {};
+  for (const key of childKeys) {
+    const keep =
+      DIFF_ALWAYS_KEEP.has(key) ||
+      (key in BED_PLATE_KEYS
+        ? anyPlateDiffers
+        : CALIBRATION_KEYS.has(key)
+          ? anyCalibrationDiffers
```
hyiger, line 519: **P2 · `filament_max_volumetric_speed` triggers the whole calibration keep-group, fabricating a spurious `calibrations[]` row and severing inheritance.**

`filament_max_volumetric_speed` is a member of `CALIBRATION_KEYS`, so when a variant's *only* genuine override is max-vol (a common per-material Orca tune), `anyCalibrationDiffers` is true and this ternary force-keeps every calibration key on the flattened child — including parent-equal `filament_flow_ratio` / `pressure_advance` / `fan_max_speed` inherited from the template. `parseBambuStudioProfile` then sees those non-null and flips `hasAnyHint` true, defeating the deliberate max-vol exclusion (#387/#950): with no resolvable printer the route does `calibrationUnresolved++` (the misleading toast that exclusion removed), and with a resolvable printer the applier pins a parent-equal `calibrations[]` row that — since `calibrations[]` inherits as a whole array with no apply-side pruning — stops the variant inheriting the parent's calibration.

Suggest excluding `filament_max_volumetric_speed` from the calibration-diff trigger (it already feeds only the top-level field via `PRUNABLE_RAW_KEYS`). Currently untested — the atomic-group tests use `pressure_advance` / `filament_flow_ratio`, never max-vol as the sole differing key.

src/lib/orcaSlicerImport.ts
```diff
@@ -0,0 +1,576 @@
...
+ * bed-plate group (`bedTypeTemps[]` — whole-array inherit) and the
+ * calibration group (`calibrations[]`) have NO parent-equality pruning on
+ * the apply side, so feeding them parent-equal values would PIN them on
+ * the variant and sever inheritance rather than restore it.
+ * (`filament_max_volumetric_speed` is safe despite being a calibration
+ * key: it's excluded from `hasAnyHint`, so riding it alone never
+ * fabricates a calibrations[] row — it only feeds the top-level field's
+ * pruning.)
+ */
+export const PRUNABLE_RAW_KEYS: readonly string[] = [
```
hyiger, line 231: **P3 · Bed-temp keys are absent from `PRUNABLE_RAW_KEYS`, so a stale diverging `temperatures.bed` override never re-inherits on re-import.**

The new per-key temperature pruning in `buildStructuredUpdate` covers `temperatures.bed` (fed by `hot_plate_temp`), but `hot_plate_temp` is only a `BED_PLATE_KEY` — it is not in `PRUNABLE_RAW_KEYS` (which lists the #403 scalars + the four nozzle keys). On a same-parent variant re-import whose `hot_plate_temp` has reconverged to the parent's value, `diffOrcaRaw` drops it (no plate key differs → `anyPlateDiffers` false) and `variantUpdateRaw` won't re-add it, so `parseBambuStudioProfile` yields `temperatures.bed = undefined`, the apply-side divergence-clear never sees the key, and a previously-diverged local bed temp stays pinned — the exact stuck-override bug the pruning fixes, fixed only for nozzle temps.

The gap is genuinely awkward (adding `hot_plate_temp` to `PRUNABLE_RAW_KEYS` would pin the un-pruned `bedTypeTemps[]` Hot Plate entry instead), so at minimum a code comment should acknowledge that the top-level bed divergence-clear is currently unreachable.

src/lib/orcaSlicerImport.ts
```diff
@@ -0,0 +1,576 @@
...
+
+/**
+ * The update payload for an EXISTING variant of the same planned parent:
+ * the diff plus any parent-equal PRUNABLE_RAW_KEYS from the flattened
+ * child (keys the diff kept — i.e. genuinely differing values — win).
+ * Fresh creates keep the bare diff: the create path has no existing doc,
+ * so the pruning can't run there and a parent-equal key would be pinned
+ * at birth instead of inheriting.
+ */
+export function variantUpdateRaw(entry: OrcaImportPlanEntry): Record<string, unknown> {
```
hyiger, line 253: **P3 · `PRUNABLE_RAW_KEYS` hand-mirrors the apply-side prune set across two files with no parity guard.**

This list is a hand-written raw-key mirror of `buildStructuredUpdate`'s `setIfNotInherited` scalars + the per-key temperature merge in `bambuStudioApply.ts`. If a future maintainer adds an inheritable scalar to `setIfNotInherited` but forgets to mirror it here, `variantUpdateRaw` won't re-attach the now-parent-equal key, the `$unset` that resumes inheritance never fires, and the variant stays pinned to its stale override — silently, with all tests green. This module already imports `BED_PLATE_KEYS`/`CALIBRATION_KEYS` from `bambuStudioImport.ts`; deriving this set from the apply-side source (or at least a parity test against it) would prevent the drift.

src/lib/orcaSlicerImport.ts
```diff
@@ -0,0 +1,576 @@
...
+ *   - CALIBRATION keys → `calibrations[]`: same rule; additionally keep
+ *     the printer-context keys (printer_settings_id etc.) so the route can
+ *     resolve the variant's own calibration row.
+ */
+export function diffOrcaRaw(
+  flatChild: Record<string, unknown>,
+  flatParent: Record<string, unknown>,
+): Record<string, unknown> {
+  const childKeys = Object.keys(flatChild);
+  const plateKeys = childKeys.filter((k) => k in BED_PLATE_KEYS);
```
hyiger: **P3 (hardening) · `k in BED_PLATE_KEYS` walks `Object.prototype`, misclassifying prototype-named preset keys.**

`BED_PLATE_KEYS` is a plain object literal, so `in` returns true for a preset key literally named `constructor` / `toString` / `hasOwnProperty` / `valueOf` (here and at line 477), mis-routing it into the bed-plate atomic group. `rawValuesEqual` then compares against the inherited prototype value (`JSON.stringify` → `undefined`), flips `anyPlateDiffers` true, and force-keeps all real bed-plate keys even when identical to the parent. `CALIBRATION_KEYS` is a `Set`, so its `.has` is already safe — switch the bed-plate checks to `Object.prototype.hasOwnProperty.call(BED_PLATE_KEYS, k)` (or make it a `Set`). Requires a hand-rolled/adversarial preset, so it's a robustness gap rather than a real-user defect.

src/app/api/filaments/orcaslicer/route.ts
```diff
@@ -105,23 +117,256 @@ export async function GET(request: NextRequest) {
...
+            // survives): idempotent re-import either way. The payload is
+            // the diff PLUS the parent-equal PRUNABLE_RAW_KEYS from the
+            // flattened child — riding those along is what lets the GH
+            // #403 pruning in buildStructuredUpdate clear a stale local
+            // override the profile no longer carries, instead of leaving
+            // the variant stuck on it forever (Codex P2 on PR #985). The
+            // groups with no apply-side pruning (settings bag, bed-plate,
+            // calibrations) stay diff-only so parent-equal values aren't
+            // pinned.
+            rawPayload = variantUpdateRaw(entry);
```
hyiger, line 341: **P3 (narrow TOCTOU) · Same-parent-variant update sends a sparse diff with `createParentId` still `null`, so a mid-flight purge makes the fall-through create a parentless root.**

In this branch `rawPayload = variantUpdateRaw(entry)` (a sparse diff) with `intendsVariant = true`, while `createParentId` stays `null`. If the existing variant row is soft-deleted *and* purged between the advisory `findCollision()` and the upsert's phase-1/2 atomics, both return null and phase 3 *creates* a new row from the sparse diff with `parentId` unset — a standalone root missing every parent-inherited field the diff omitted (the required-field guard passes because type/vendor/name/diameter are all `DIFF_ALWAYS_KEEP`). The counter at line 340 then still runs `variants++`, mislabeling the corrupt root. The window is narrow, but setting `createParentId = parentDocId` in this branch would make a fall-through create re-link instead of orphan.

src/app/api/filaments/orcaslicer/route.ts
```diff
@@ -105,23 +117,256 @@ export async function GET(request: NextRequest) {
...
+
+        // Advisory pre-check for the collision trees below — the atomic
+        // phases in upsertParsedBambuFilament are what actually protect the
+        // write (updates never $set parentId, so a re-parent can't slip
+        // through a TOCTOU window; worst case a racing delete/trash demotes
+        // a collision case to the create or resurrect path). Resolution
+        // order mirrors the upsert's own: an ACTIVE row first (phase 1),
+        // else a TRASHED non-purged row (phase 2 resurrects it, so it
+        // collides exactly like an active row — PR #985 P2).
+        const findCollision = async () =>
```
hyiger, line 273: **P3 (efficiency) · `findCollision` duplicates the name lookups `upsertParsedBambuFilament` immediately re-runs.**

`findCollision` issues `findOne({ name, _deletedAt: null })` and, on a miss, `findOne({ name, trashed })` purely to read the existing row's `parentId` for variant-vs-standalone routing — then `upsertParsedBambuFilament` re-issues those byte-for-byte identical queries in phases 1 and 2 (`bambuUpsert.ts:109/141`). That's 4 round-trips where 2 suffice on a fresh create, per entry, scaling linearly with the selection (cap 10k). Having the upsert return the resolved existing doc's `parentId` (or a small collision descriptor) would let the route drop `findCollision` entirely — halving the lookups and removing the TOCTOU divergence the comment currently has to reason around.

src/components/OrcaLibraryImportDialog.tsx
```diff
@@ -0,0 +1,457 @@
...
+    document.addEventListener("keydown", handler);
+    return () => document.removeEventListener("keydown", handler);
+  }, [onClose]);
+
+  const handleFolder = async (e: React.ChangeEvent<HTMLInputElement>) => {
+    const allFiles = Array.from(e.target.files ?? []).filter((f) =>
+      f.name.toLowerCase().endsWith(".json"),
+    );
+    if (folderInputRef.current) folderInputRef.current.value = "";
+    if (allFiles.length === 0) {
```
hyiger, line 133: **P3 · Empty-folder re-scan leaves stale `selected`/`results`, enabling a wasted Import POST.**

This early return does `setByName(new Map())` + `setScanInfo(...)` but skips the `setSelected(new Set())` / `setResults(null)` that the success path runs (lines 173–174). So after scanning folder A (with profiles checked) and then re-scanning an empty / non-Orca folder, `selected` still holds folder-A names, the Import button stays enabled (`disabled={submitting || selected.size === 0}`), and `handleSubmit`'s `!byName` guard passes on the truthy empty Map — POSTing `{ selected: [...], profiles: [] }`, which the route rejects with a single misleading 400 toast while folder A's stale summary still shows. Mirror the success path: clear `selected` and `results` in this branch too.

hyiger (commented):
## Automated review — bulk OrcaSlicer filament-library import

Multi-agent review: 10 finder dimensions → adversarial per-finding verification → synthesis. Overall this is a careful, well-commented PR. The three-phase upsert extraction into `bambuUpsert.ts` is clean, the planner's chain flatten/collapse logic is sound, the CSRF guard + 10 MB byte cap + 10k-profile cap are all in place, the `orcaImport.*` i18n keys are at EN/DE parity, and the docs/OpenAPI match the implementation. **No P1s.**

The residual risks cluster in two places: `diffOrcaRaw`'s atomic calibration/bed-plate keep-groups, and the parent-equality *prune parity* between the planner (`orcaSlicerImport.ts`) and the apply layer (`bambuStudioApply.ts`).

**Findings:** 1 × P2, 6 × P3 — details inline.

_Two candidates were investigated and **refuted**: a same-library concurrent-import orphan (topological plan ordering + the upsert's atomic phases prevent it) and prototype-pollution via attacker JSON keys (no reachable gadget)._

src/lib/orcaSlicerImport.ts
dcode: Fixed — added a `CALIBRATION_ATOMIC_KEYS` set (`CALIBRATION_KEYS` minus `filament_max_volumetric_speed`) and used it in `diffOrcaRaw`'s trigger/keep logic instead of the full `CALIBRATION_KEYS`. Max-vol now falls through to ordinary per-key diffing, matching the parser's existing exclusion of it from `hasAnyHint` — a variant whose only real override is max-vol no longer force-keeps every other parent-equal calibration key. Added regression tests covering max-vol as the sole differing key, and max-vol differing alongside a genuine calibration key (still correctly force-keeps the group in that case).

src/lib/orcaSlicerImport.ts
dcode: Agreed this is a real gap, but I didn't find a clean fix — adding `hot_plate_temp` to `PRUNABLE_RAW_KEYS` would make `variantUpdateRaw` re-attach a lone parent-equal `hot_plate_temp` on re-import, which the parser turns into a `bedTypeTemps` entry that the applier unconditionally `$set`s, *pinning* the un-pruned "Hot Plate" bed-plate entry instead of restoring inheritance — trading one stuck-override bug for a different one. Took your suggested minimum: documented the gap directly on `PRUNABLE_RAW_KEYS` (a "KNOWN GAP" block explaining why `hot_plate_temp`/`hot_plate_temp_initial_layer` are deliberately excluded and what it would take to close properly — apply-side pruning for `bedTypeTemps[]`, which is out of scope here) so the top-level bed-temp divergence-clear stays a documented, intentional limitation rather than a silent one.

src/lib/orcaSlicerImport.ts
dcode: Fixed — rather than deriving `PRUNABLE_RAW_KEYS` from the apply-side source (the two sides operate on different key spaces — raw Orca keys here vs. DB field names in `buildStructuredUpdate`, so a clean derivation isn't straightforward), added a parity regression test that parses `bambuStudioApply.ts`'s actual `setIfNotInherited(...)` call sites via source inspection and asserts every optional inheritable scalar it prunes (excluding the schema-required `type`/`vendor` and the always-kept `diameter`) has a corresponding raw key in `PRUNABLE_RAW_KEYS`. A future scalar added to one side and forgotten on the other now fails CI instead of silently drifting.

src/lib/orcaSlicerImport.ts
dcode: Fixed — switched both bed-plate key checks (here and the other occurrence) from `k in BED_PLATE_KEYS` to `Object.prototype.hasOwnProperty.call(BED_PLATE_KEYS, k)`, closing the prototype-chain misclassification. Added a regression test with a preset carrying a `constructor`-named key, asserting it doesn't get misrouted into the bed-plate atomic group.

src/app/api/filaments/orcaslicer/route.ts
dcode: Fixed, and then hardened further. Initially set `createParentId` in the same-parent-variant-update branch as suggested, so a mid-flight purge falls through to a fallback create that still links to the parent.

On a closer pass this turned out to be one instance of a broader pattern: the advisory `findCollision()` check and the atomic write in `upsertParsedBambuFilament` can observe different states of the same row across all four collision branches, not just this one — a concurrent write between the two could apply a diff/payload baselined against the wrong parent, or let a planned root silently land on what turned out to be a variant of another filament (cascading a variant-of-a-variant to its children). Rather than patching this branch alone, added an opt-in `expectedParentId` option to `upsertParsedBambuFilament` — Phase 1's and Phase 2's atomic `findOneAndUpdate` filters and the E11000 race-recovery merge all now assert the caller's observed parent expectation, failing closed with a 409 on mismatch instead of silently writing. Wired it into all four collision branches (root; fresh variant create; same-parent update; existing-standalone collision). Four dedicated regression tests reproduce each race deterministically (via mocked call interleaving, not timing) and confirm the racing document's content is provably untouched when the guard fires.

src/app/api/filaments/orcaslicer/route.ts
dcode: Deferring this one. It's pure efficiency with no correctness impact, and a proper fix means threading a pre-resolved document through the shared `bambuUpsert.ts` (also used by the Bambu Studio route) without weakening its atomicity guarantees — I'd rather not destabilize that path for a latency win alone. Worth noting: the TOCTOU divergence concern this comment raises as a side effect of the duplicate lookups is now closed by the `expectedParentId` mechanism added for the sibling finding on this same route (see the reply on the "Same-parent-variant update" comment) — a stale `findCollision()` observation now fails the write atomically rather than silently applying against a state that's since diverged. Happy to pick up the pure query-count optimization as a follow-up if it's still wanted.

src/components/OrcaLibraryImportDialog.tsx
dcode: Fixed — the empty-scan early return now also calls `setSelected(new Set())` and `setResults(null)`, mirroring the success path.

hyiger:
@codex review

src/app/api/filaments/orcaslicer/route.ts
```diff
@@ -105,23 +117,331 @@ export async function GET(request: NextRequest) {
...
+        // WRITE itself with `!result.ok` before this line is ever reached,
+        // so `result.doc.parentId` is unconditionally falsy here for a
+        // root-kind entry. Pinned by the "fails the write closed..."
+        // regression test below.
+
+        if (result.created) created++;
+        else updated++;
+        // A resurrect keeps the trashed row's old parent state, so only a
+        // fresh create or a confirmed same-parent update counts as a variant.
+        if (intendsVariant && (result.created || createParentId === null || sameParentVariantUpdate)) {
```
hyiger: **P3 · `variants` summary counter under-reports a race-recovered variant merge**

When two concurrent imports create the *same brand-new variant* under the same parent, the loser's phase-3 E11000 recovery merges and returns `created=false`. This gate then fails all three disjuncts (`result.created` false, `createParentId` is the non-null parent id, `sameParentVariantUpdate` false), so the item is tallied `updated++` instead of `variants++` — inconsistent with the same-parent-update path. The data written is correct; only the response summary counts mislabel it.

Also: `createParentId === null` is effectively dead for variant entries (no variant branch leaves it null), so it never rescues this case.

_Distinct from the resolved same-parent-variant TOCTOU/diff findings — this is response-count only, on the race-recovery path._

src/components/OrcaLibraryImportDialog.tsx
```diff
@@ -0,0 +1,459 @@
...
+      }
+      skipped += Math.max(0, files.length - MAX_FILES);
+      const indexed = indexOrcaProfiles(raws);
+      skipped += indexed.errors.length;
+      setByName(indexed.byName);
+      setScanInfo({
+        found: [...indexed.byName.values()].filter((n) => n.concrete).length,
+        skipped,
+      });
+      setSelected(new Set());
```
hyiger, line 178: **P3 · Folder rescan doesn't reset the search / vendor / material filters**

This block resets `byName`/`scanInfo`/`selected`/`results` on a new scan, but not `search`/`vendorFilter`/`materialFilter`. Scenario: pick folder A → set the vendor filter to "Polymaker" → pick folder B (Prusa only). `scanInfo` reports "12 found", but the stale `vendorFilter` filters them all out, and `<select value="Polymaker">` (not in B's option list) renders **blank** — so the picker looks unfiltered yet shows "no matches" over 12 found profiles. Reset the three filter fields here too.

_Distinct from the resolved empty-folder finding — that reset `selected`/`results` (present here); the filter fields are a separate, still-unreset set._

src/lib/bambuUpsert.ts
```diff
@@ -0,0 +1,353 @@
...
+      resurrectUpdate.$set = {
+        ...(resurrectUpdate.$set as Record<string, unknown>),
+        _deletedAt: null,
+      };
+      const resurrectFilter: Record<string, unknown> = {
+        _id: existingTrashed._id,
+        _deletedAt: { $ne: null },
+        _purged: { $ne: true },
+      };
+      if (opts?.expectedParentId !== undefined) {
```
hyiger, line 194: **P3 (test gap) · The phase-2 resurrect `expectedParentId` guard has no protective test**

This constraint is the core concurrency hardening, but no test asserts its *blocking* behavior: every test that reaches phase 2 passes a matching parent, and every parent-mismatch test has nothing trashed (so phase 2 is skipped). Deleting these two lines leaves the whole suite green. Add a test that resurrects a trashed row whose parent no longer equals `expectedParentId` and asserts phase 2 is blocked (falls through to create / 409).

_Distinct from the resolved "no resurrect-collision test" finding — that test was added, but it uses a matching parent, so the mismatch-guard path is still unexercised._

hyiger (commented):
Fresh adversarial review of the current HEAD (`a4fce99a`), deduped against the 15 prior resolved findings. **3 new findings, all P3 (0 P1 / 0 P2)** — each sits next to a previously-resolved finding but is a distinct residual gap. Nothing blocking.

- **#2 (Dialog filter reset)** is the only user-visible one.
- **#1** is a cosmetic response-count mislabel (the data written is correct).
- **#3** is a test-coverage gap (the guard works; a regression just wouldn't be caught).

src/app/api/filaments/orcaslicer/route.ts
dcode: Fixed in 4ba6d24. `variants++` now fires whenever `intendsVariant` is true, regardless of `result.created`/`createParentId`/the removed `sameParentVariantUpdate` flag — by the time this line runs, a parent-expectation mismatch has already failed closed via `expectedParentId` (added for the sibling TOCTOU finding), so `intendsVariant` alone is sufficient to know the write is a legitimate variant write, including an E11000 race-recovery merge. New route test races a concurrent create of the same brand-new variant and asserts `variants: 1` / `updated: 2` rather than the old mislabeled split.

src/components/OrcaLibraryImportDialog.tsx
dcode: Fixed in 4ba6d24. The success-path scan now also resets `search`/`vendorFilter`/`materialFilter` alongside `selected`/`results` (and the empty-folder early return picked up the same three resets too), so switching folders can no longer leave a stale filter silently hiding — or blank-rendering — a fresh scan's rows.

src/lib/bambuUpsert.ts
dcode: Fixed in 4ba6d24. New test creates a trashed row parented under one ObjectId, then calls upsert with a different `expectedParentId` — no mocking, so the resurrect atomic's own `parentId: opts.expectedParentId` filter genuinely misses and falls through to phase 3 (a fresh create), leaving the original trashed row untouched (still trashed, still carrying its real parent). Confirms the guard actually blocks the mismatch rather than merely being passed by every existing test's matching-parent setup.

src/lib/bambuUpsert.ts
```diff
@@ -0,0 +1,353 @@
...
+      parsed,
+      await augmentExistingWithParent(racing),
+    );
+    const racePayloadFailure = bambuPayloadFailure(racePayload);
+    if (racePayloadFailure) return racePayloadFailure;
+    delete (racePayload.update as Record<string, unknown>).spools;
+    try {
+      const merged = await Filament.findOneAndUpdate(
+        { _id: racing._id, _deletedAt: null },
+        composeMongoUpdate(racePayload),
```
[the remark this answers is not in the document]
dcode: Working on this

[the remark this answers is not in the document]
dcode: Fixed in cfc4470. Baked `parentId: opts.expectedParentId` into the E11000 merge's own atomic `findOneAndUpdate` filter (matching the pattern phases 1 and 2 already use), so a second reparent landing between the pre-check and the merge write now makes the atomic write itself miss — not just the earlier advisory check. When it misses and `expectedParentId` was set, the route now returns a distinct 409 ("belongs to a different parent than expected") instead of falling through to the generic 500 "Failed to create filament." New regression test forces this exact double-race with a real second `parentId` mutation (not a mocked null return) landing between the pre-check and the write, then asserts the racing row's content is provably untouched. Went through an adversarial review pass (Antigravity/gemini-3.1-pro) before landing — approved with no findings; it ran the full suite + lint and traced the test to confirm it exercises the atomic filter for real.

hyiger:
@codex review

src/lib/bambuUpsert.ts
```diff
@@ -0,0 +1,370 @@
...
+  if (!parsed.filament.vendor) {
+    return {
+      ok: false,
+      status: 400,
+      error:
+        "Bambu Studio profile is missing filament_vendor — required to create a new filament",
+    };
+  }
+
+  const createPayload = await prepareBambuUpdate(parsed, null);
```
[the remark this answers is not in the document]
hyiger: @dcode if you like this PR merged please implement this fix.

dcode:
Yes I will. Sorry for the delay, I've been a bit tied up with work and
traveling. I will have time this week to wrap it up correctly (hopefully).
It's clearly more complicated than I initially thought.

PR #987, opened 2026-07-05
feat(settings): user-selectable number format (i18n grouping + decimals)

hyiger:
Adds a **Number format** control to Settings → UI, following the same setup as the date-format feature (#984). Requested: format numbers per locale — `1,245,414.45` / `1.245.414,45` / `1 245 414` — plus a custom format.

## What you get (Settings → UI → Number format)
- **System (device region)** — follows the OS/device locale via `Intl`. Default.
- **US / UK** — `1,234,567.89`
- **European** — `1.234.567,89`
- **Space-separated** — `1 234 567,89` (U+202F narrow no-break space)
- **Custom** — pick your own thousands + decimal separator (Swiss `1'234.56`, etc.), with live preview + validation.

Applies to **all displayed numbers** — weights, costs, counts, analytics totals, densities/diameters/lengths — **and to currency** (a European user sees `€1.234,56`).

## How it works (mirrors the date feature)
- **`src/lib/numberFormatPref.ts`** (pure, DB-free, **100% coverage**): preference shape, `isValidSeparators` (group≠decimal, single non-digit), preset table, and a deterministic branchless `formatWithSeparators` (manual grouping + parser-shift rounding). Space preset uses ` ` to match modern ICU `fr-FR`.
- **`src/hooks/useNumberFormat.ts`**: module store + `useSyncExternalStore`, electron-store `numberFormat` / localStorage `filamentdb-number-format`, a **`ready` gate** so the first SSR + hydration render equals today's output — **no React 19 hydration mismatch** (verified: zero console warnings). Exposes `formatGrams` (drop-in), `formatNumber`, `separators`.
- **Currency**: threaded into `useCurrency` — a non-system preset swaps **only** the `group`/`decimal` parts via `Intl.NumberFormat.formatToParts`, so each currency's own decimal count (JPY=0, BHD=3) and symbol placement are preserved. System mode is untouched.
- **Electron bridge** wired in all three files (`main.ts` get/save + setter, `preload.ts`, `electron.d.ts`) — the P1 that was missed on the date PR.
- Migrated **~35 `formatGrams` + `kg`/density/diameter/length `toFixed` + count `toLocaleString`** across 10 files. **Excluded** (machine-parseable / non-display): CSV/XLSX + slicer exports, `snapToStep`, `FilamentSwatch` CSS %, `<input>` values, `api/**`, integer temps/percentages.
- i18n: `settings.numberFormat.*` in **en + de** (parity gate green).

## ⚠️ Behaviour note
In **system** mode, weights now group per the device locale (e.g. `1,234 g` on a US device) where they were previously ungrouped — a visible default change, `ready`-gated so there's no hydration flash on client-fetched data. Flagging in case you'd prefer system mode to stay ungrouped.

## Verification
- Full suite green — **3182 tests**; new core at **100%** coverage; `tsc`, `electron:typecheck`, `eslint` clean.
- In-browser (live data): all 5 presets, custom `'`/`.` → `1'245'414.45`, invalid pair → error + safe fallback to system, persistence across reload, and currency reformatting (`$20,95` under European — `$` kept, separators swapped). No hydration warnings.


hyiger:
@codex review

hyiger:
@codex review

hyiger:
Fixed in 3cae80e6: `useNumberFormat` now exposes `systemLocale` (the device locale, post-hydration) and `useCurrency` uses it as the Intl base (`localeOverride ?? systemLocale ?? locale`). So in System mode currency groups per device, consistent with weights/counts; presets still swap group/decimal parts and pre-hydration keeps the app locale (hydration-safe). tsc + eslint + tests green.

@codex review

hyiger:
Both addressed in ab152981 — currency now branches on three explicit cases:
- **Preset/custom**: app-locale Intl base (symbol placement per #821) + swap group/decimal only — fixes the placement regression at :278.
- **System mode**: device-locale base for ISO *and* custom (non-ISO) codes; custom codes now use `Intl.NumberFormat(deviceLocale, 2dp)` instead of raw `toFixed(2)` — fixes :313.
- **Pre-hydration**: app locale / `toFixed(2)`, matching the server.

tsc + eslint + tests green.

@codex review

hyiger:
@codex review

hyiger:
Added a **None** preset (no thousands grouping, e.g. `12345689.56`) in 776d6530 — Custom can't express no-grouping (its separator must be a single char), so None fills that gap as a fixed `{group:"", decimal:"."}` preset. Core stays at 100% coverage.

@codex review

hyiger:
Fixed in a50997c9: the inventory group summary now formats `count` through `formatNumber({maxDecimals:0})` too, so a 1,000+-spool group no longer mixes grouped grams with an ungrouped count. Pluralization still keys on the raw count.

@codex review

hyiger:
@codex review

PR #993, opened 2026-07-07
fix(api): cap POST /api/spools/import body size before buffering (#991)

hyiger:
## Problem

`POST /api/spools/import` accepts JSON `{csv}`, multipart, and raw-text CSV, but **every branch buffered and parsed the full body before enforcing any byte-size limit**. The only bound was `parseCsv`'s **row-count** cap — which runs *after* the whole CSV string is already in memory, and a few rows with enormous cells slip past it entirely.

Because [`next.config.ts`](next.config.ts) raises `proxyClientMaxBodySize` to `52mb` for snapshot restores, this "small paste/spreadsheet" endpoint inherited a ~52 MB request budget. On the default **unauthenticated** local/LAN API, any reachable client could spend server memory/CPU before anything was rejected.

The `checkFileSize`/`checkContentLength` helpers (10 MB → 413) already exist in [`apiErrorHandler.ts`](src/lib/apiErrorHandler.ts) and **every sibling import route already uses them** (import-csv, bambustudio, prusaslicer, nfc/decode). This route just never called them.

## Fix

Adopt the siblings' 10 MB (`MAX_UPLOAD_SIZE`) posture, per input shape, in [`route.ts`](src/app/api/spools/import/route.ts):

- **raw / JSON** — `checkContentLength(request)` preflight **plus** a post-buffer `Buffer.byteLength(csvText, "utf8") > MAX_UPLOAD_SIZE` re-check (catches a missing/lying `Content-Length`; byte length, not `String.length`, per the same UTF-8 reasoning as prusaslicer's Codex P2 on #685).
- **multipart** — `checkFileSize(file)` before `file.text()`. Deliberately **not** `checkContentLength` here: the multipart `Content-Length` includes MIME-envelope overhead that would over-count a legitimate ~10 MB file. `file.size` is the correct bound (matches bambustudio).

### Two load-bearing placement details
1. The `checkContentLength` preflight sits **outside** the read `try/catch`, so a genuine 413 isn't downgraded into the 400 "Failed to read request body" path.
2. The multipart guard runs *after* `formData()` has buffered the envelope (undici buffers parts to build the `File`) — so it prevents the expensive `file.text()` decode + CSV parse but not the envelope buffering itself. This is the **identical** limitation the sibling routes already accept — consistent, not a new gap.

Kept the existing `assertSameOriginRequest` guard first, and the 10 MB ceiling matches the other import routes (triage decision).

## Tests

6 new cases in [`tests/spools-import-route.test.ts`](tests/spools-import-route.test.ts), each asserting **413 before any CSV parsing / DB work** (`body.results` / `body.imported` are absent):
- raw text/csv with oversized **declared** `Content-Length` (preflight — tiny body + fabricated header, fast)
- raw body **actually >10 MB** with an **understated** `Content-Length` (post-buffer byteLength re-check)
- JSON `{csv}` with oversized declared `Content-Length` (preflight)
- JSON `{csv}` whose csv value is >10 MB with an understated `Content-Length` (byteLength re-check)
- multipart `File` >10 MB (`checkFileSize`)
- regression: a normal small body with an honest `Content-Length` still imports (200) — the preflight isn't over-eager

A single ~11 MB body is built once and shared across the large-payload cases to keep the suite fast. Full route suite: **49 passed**. `tsc --noEmit` and `eslint` on the changed files both clean.

Fixes #991

### DIFF
--- a/src/app/api/spools/import/route.ts
+++ b/src/app/api/spools/import/route.ts
@@ -15,6 +15,17 @@ import { unsanitizeCsvCell } from "@/lib/csvWriter";
 import { isValidIsoDateString, validateSpoolInstanceId, MAX_SPOOL_TEXT_LENGTH } from "@/lib/validateSpoolBody";
 
 /**
+ * Slack added to the 10 MB cap for the multipart Content-Length preflight, to
+ * cover the MIME envelope (boundary lines + per-part headers + CRLFs) so a
+ * legitimate ~10 MB file isn't rejected for its framing bytes. The route reads
+ * only the `file` field, so a real upload's overhead is a few hundred bytes;
+ * 64 KB is generous headroom while still far below the multi-MB abuse this
+ * guards against (GH #991). `checkFileSize` enforces the exact 10 MB on the
+ * file part itself.
+ */
+const MULTIPART_OVERHEAD_ALLOWANCE = 64 * 1024;
+
+/**
  * POST /api/spools/import — bulk-create OR upsert spools from CSV.
  *
  * Accepts either:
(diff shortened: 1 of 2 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
Addressed all three P2 findings in a1645e00:

1. **Spoofed multipart parameter bypass** (line 60) — the branch and the size-guard gating now derive from the parsed **media-type essence** (`type/subtype`, parameters stripped), not a substring match. `application/json; x="multipart/form-data"` now correctly branches as JSON and both guards apply. Legitimate `charset=`/`boundary=` params are still ignored.
2. **Cap the full JSON payload, not just csv** & 3. **Check raw JSON bytes before parsing** (line 117) — the non-multipart branches now read the raw body and byte-check the **whole buffer before any `JSON.parse`**, so a JSON body with a huge sibling field (or a missing/lying/chunked `Content-Length`) is rejected on total size, not just the decoded `csv` string.

Added 2 regression tests pinning both (small-csv/huge-padding JSON → 413; spoofed content-type param → 413). Full route suite: 51 passed; `tsc` + `eslint` clean.

@codex review

hyiger:
@codex review

hyiger:
Addressed the round-2 P2 in 9e51c5d1: the Content-Length preflight now runs for **every** shape, including multipart, with a 64 KB envelope-overhead allowance (`MULTIPART_OVERHEAD_ALLOWANCE`). So a large total multipart body — a big file **or** a small file plus tens of MB of extra fields — is rejected **before** `formData()` buffers it, while a legitimate ~10 MB file isn't tripped by its boundary/header framing. `checkFileSize` still enforces the exact 10 MB on the file part (defense for a chunked request with no Content-Length).

Added a test asserting an oversized-Content-Length multipart body → 413 before the form is parsed. Route suite: 52 passing; `tsc` + `eslint` clean.

@codex review

PR #1009, opened 2026-07-12
fix(data-integrity): P1 batch — purged-zombie import + snapshot rollback (#1004 F1/F2)

hyiger:
First PR from the #1004 data-integrity audit — the two **P1** permanent-data-loss paths, both reachable from normal user flows, plus a startup repair for rows the first bug already created.

## F1 — CSV/XLSX import resurrected `_purged` tombstones (`src/lib/importFilaments.ts`)
Permanent delete sets `_purged: true` **and** `_deletedAt`. The importer bucketed any `_deletedAt != null` doc as a resurrect candidate and revived it **without clearing `_purged`** → an active "zombie" that:
- poisons hybrid sync (the engine short-circuits on `_purged` before LWW — stamps the peer dead, never syncs the row's edits), and
- vanishes unrecoverably if re-trashed (the trash listing filters `_purged`).

Fix: `_purged` added to the projection; bucketing excludes purged docs (a purged name falls through to `Filament.create`, legal under the partial-unique-on-`_deletedAt:null` name index); the resurrect write is guarded (`{ _purged: { $ne: true } }`) and, on `matchedCount === 0` (a permanent delete landing mid-import), mints a fresh doc instead of falsely reporting `updated`.

## F2 — snapshot restore silently lost legacy docs (`src/app/api/snapshot/route.ts`)
- **Rollback** now uses `lean: true` + per-collection count checks. The backup was read verbatim from this DB via `.lean()`; re-validating it (old `ordered:false` + default `throwOnValidationError:false`) **silently dropped** any legacy doc failing the *current* schema while claiming a full rollback — exactly the population the #905 `validateModifiedOnly` fixes exist for.
- **Pre-validation** before the destructive wipe: a bad snapshot → **400 with the DB untouched**, instead of wipe→fail→lossy-rollback (which restoring one's *own* legacy-carrying snapshot triggered every time).

## Migration (`src/lib/mongodb.ts` `purgedZombies`)
Re-tombstones any active `_purged` rows the pre-fix importer already created (restores `_deletedAt` — gone-forever is their intended state). Retry-tracked + idempotent; matches nothing on healthy installs. **Release note owed:** on an affected install, rows a user currently sees will move into the (purged, invisible) state on first connect.

## Tests
purged→create, Parent-column interplay, mid-import purge **race guard**, mixed same-name ordering; snapshot pre-validation 400 (no wipe), the P1 **rollback-preserves-legacy-doc** pin; migration repair + transient-retry. New `src/lib` code at **100% lines**. Full suite green (**3229**).

Part of #1004 (findings 1, 2). The other 5 findings ship in follow-up PRs.


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #1010, opened 2026-07-12
perf: stop shipping spool blobs that handlers discard (#1005 F1/F2/F4)

hyiger:
From the #1005 performance audit — three read paths fetched or piped every spool subdocument (base64 `photoDataUrl` up to 5 MB each + `usageHistory` ledgers) and then discarded them. Same class as the fixed #517 dashboard issue and `matchFilament`'s `MATCH_PROJECTION` (#955.12).

## F1 — `/api/analytics` (P1)
`.select(...spools)` → **`.select(...spools.usageHistory)`** (the only spool subfield the manual-entries loop reads). Fires on `/analytics` mount and every `days` toggle; at 5,000 spools with ~500 photos that's ~1 GB per request into the Node heap.

## F2 — slicer bundle exports (P2)
`.select("-spools")` on all **5** find sites (prusaslicer main + missing-parents, orcaslicer main + missing-parents, export). Bundle generation never reads spools (`grep spools src/lib/*SlicerBundle.ts` = 0), yet the PrusaSlicer fork hits these on **every slicer startup**.

## F4 — aggregation pipelines (P3)
- **by-location**: early `{ $unset: ["spools.photoDataUrl", "spools.usageHistory"] }` after the first `$match` (the #777 whole-`$spools` reference otherwise forces every subfield through `$lookup`/`$unwind`/`$group`). **Keeps `spools.dryCycles`** — `dryCycleCount`/`lastDryAt` need it.
- **locations `?stats=true`**: `$unset` all three heavy subfields (no dry stats computed there).

## Safety
Responses are **byte-identical** — none of these fields ever appeared in the outputs. Tests: analytics projection spy + photo/dry behavioral case; by-location photo-laden dry-stats regression; prusaslicer variant-inheritance via the missing-parents fetch (with a photo-laden parent spool). Full suite green (**3224**).

Part of #1005 (findings 1, 2, 4). Finding 3 (client-side Intl formatter caching) ships separately.


hyiger:
@codex review

hyiger:
@codex review

PR #1011, opened 2026-07-12
fix(sync): faithful whole-doc LWW copies + share view no longer bumps updatedAt

hyiger:
Part of #1004 (findings 3, 4, 5). Third PR in the post-1.66.1 audit remediation series.

## F3 — hybrid sync LWW must copy whole documents faithfully (`replaceOne`, not `$set`)

The last-write-wins path in `electron/sync-service.ts` used `updateOne({ $set: ... })` for whole-document copies. `$set` can only add/overwrite keys — it can never *remove* one the source dropped. So when a variant override is un-pinned (the #951/#969/#971 `$unset` flows leave a field absent so [GH #106](https://github.com/hyiger/filament-db/issues/106) inheritance resumes), the stale value **lingered on the peer**.

Worse: `$set` also copied the source's newer `updatedAt` onto the peer, leaving the two sides **content-divergent at an equal timestamp** — a state LWW can never resolve (frozen forever).

The four whole-doc copy sites (both LWW branches + both resurrect branches) now use `replaceOne`, which writes the stripped source verbatim (`stripForTransfer` already removes `_id`/`__v`, so the target `_id` is preserved) and drops fields the source no longer carries. The targeted flag writes (`_purged` / `_deletedAt` propagation) intentionally stay `$set` — they mutate a single key.

## F5 — a public share GET must not advance `updatedAt`

`GET /api/share/{slug}` increments `viewCount` with `findOneAndUpdate` on a `timestamps:true` schema, which auto-bumped `updatedAt` on **every view**. The hybrid sync engine resurrects a locally-unpublished catalog whenever the remote copy's `updatedAt` out-runs the local `_deletedAt` — so a mere GET on a still-live peer could **revive a revoked share**, defeating the DELETE handler's "unpublish sticks across peers" promise. Fixed by passing `timestamps: false` on that update. `viewCount` divergence across peers is cosmetic; the timestamp bump is not.

## F4 (warn-only) — hybrid wipe/restore limitation

In hybrid mode a local database wipe or snapshot restore appears not to "take" because the next sync re-copies the cloud (Atlas) docs back. Per the product decision (warn first, defer the propagating fix), added an amber warning banner:

- **Settings → Danger** (before the wipe)
- **Settings → Backup** (before a restore)

Shown only when `connectionMode === "hybrid"` (read post-mount via `electronAPI.getConfig()`; hidden entirely in web mode). New i18n keys `settings.danger.hybridWarning` / `settings.backup.hybridWarning` (en + de in lockstep). The actual propagate-the-wipe/restore fix is deferred as a follow-up.

## Tests
- `tests/sync-service-expansion.test.ts` — both LWW `replaceOne` branches: an un-pinned field is dropped on the peer (not retained), and the sides converge in one cycle (equal-timestamp no-op, no freeze).
- `tests/share-route.test.ts` — a GET increments `viewCount` **without** advancing `updatedAt` (pinned via a known past instant so a regression is unambiguous).

Full suite green (3223 tests). Root `tsc`, `electron:typecheck`, and ESLint all clean.


hyiger:
@codex review

PR #1012, opened 2026-07-12
fix(api): print-history refund survives startedAt edits + guard concurrent parentId cycles

hyiger:
Part of #1004 (findings 6, 7). Fourth PR in the post-1.66.1 data-integrity remediation series.

## F6 — editing `startedAt` no longer orphans the DELETE spool refund

`PUT /api/print-history/{id}` accepts `startedAt`. The DELETE refund's **tier-3 legacy matcher** pairs a usage row to its spool's `usageHistory` entry by `(grams, date)` where `date === job.startedAt`. A legacy (pre-`jobId`) entry keeps its **original** date, so editing `startedAt` made that matcher miss on a later delete — the spool weight silently never refunded even though the job was tombstoned *"Deleted and refunded"*. (Entries carrying a `jobId` — tiers 1–2 — were already immune, since they match on `jobId`, not date.)

**Fix:** on a real `startedAt` change, PUT backfills this job's id onto the matching legacy entries — found via the **old** `startedAt` + `grams`, one entry per usage row, restricted to print-driven `"job"`/`"slicer"` entries with no `jobId` (the exact tier-3 candidate set). A subsequent DELETE then resolves them through the date-independent `jobId` tiers. Reloading the filament per row + saving after each stamp means an already-stamped entry (now carrying a `jobId`) is skipped by the next row, so two rows against the same spool consume distinct entries. A backfill `save()` racing a concurrent write surfaces as 409 (retryable), matching the DELETE handler.

## F7 — concurrent re-parent PUTs can no longer create an A⇄B cycle

`PUT /api/filaments/{id}` validates `parentId` (parent exists, parent isn't itself a variant, this doc has no variants) then writes with **no re-assertion** — a check-then-act race. Two opposing re-parents (`A→B` and `B→A`) can each pass validation against pre-write state and both persist, forming a mutual A⇄B cycle that every single-level read path (`resolveFilament`) assumes cannot exist, leaving both rows undeletable "parents with variants".

**Fix:** after the write, on an actual re-parent to a non-null parent, re-assert the invariant against **post-write** state (the new parent is still a root; this doc gained no children). On violation, revert just this doc's `parentId` and return **409** so the caller retries. Un-parenting (parentId → null) can't create a cycle and is skipped.

## Tests
- `tests/api-validation-and-semantics.test.ts` — F6 backfill + end-to-end DELETE refund (pre-fix the refund stranded at 975 g; now 1000 g), plus a guard that a non-`startedAt` edit leaves `usageHistory` untouched.
- `tests/filament-put-validation-status.test.ts` — F7 happy path (no false 409), a **deterministic** post-write revert+409 (a `countDocuments` spy on the route's own model object simulates the race), and a concurrent double-re-parent invariant (never a mutual cycle).

Full suite green (3225 tests). Root `tsc` + ESLint clean. No `src/lib`/`src/models` changes, so the coverage gate doesn't apply.


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #1013, opened 2026-07-12
fix(ui): 5 frontend audit fixes (t() $-corruption, theme hydration, label i18n, chip a11y, compare AbortError)

hyiger:
Fixes #1007. Fifth PR in the post-1.66.1 audit remediation series — all 5 frontend findings.

## F1 (P2) — `t()` interpolation corrupts values containing `$` patterns
A **string** second argument to `String.prototype.replace` treats `$$`, `$&`, `` $` `` and `$'` as special replacement patterns. So a param value carrying them — a filament named `Cheap $$ PLA`, a raw server-error string — rendered **corrupted**, including in the destructive **delete/retire confirm dialogs** (the user confirms against garbled wording). The substitution is extracted to a pure `interpolate()` helper (`src/i18n/interpolate.ts`) using a **function** replacement, which inserts values verbatim.

## F2 (P3) — ThemeProvider hydration mismatch
`preference`/`systemDark` were seeded in the `useState` initializers, which run during hydration. A stored non-`"system"` theme made the client's first render (ThemeSection highlights `preference`) disagree with the server's `"system"` render → a React 19 hydration mismatch + full client re-render on every `/settings/ui` visit. Now mirrors the [#639](https://github.com/hyiger/filament-db/issues/639) pattern (seed to SSR-safe constants, read `localStorage` + `matchMedia` in a mount effect). A `hydrated` gate keeps the `<html>` class-application effect from overwriting the anti-FOUC inline script's class during the hydration frame — **no theme flash**.

## F3 (P3) — label-format preset names untranslated in German
`LABEL_PRESETS` labels were hardcoded English and rendered directly. The i18n parity test can't catch them (it only scans literal string-argument translation calls). `LABEL_PRESETS` now carries a `labelKey`; the editor renders `t(labelKey)`; four keys added to en+de.

## F4 (P3) — FilamentPicker chips misused `role="tablist"`/`"tab"`
The type-filter chips carried `role="tab"` + `aria-selected` with no tabpanel, no roving tabindex, no Arrow/Home/End, and **deselect-on-reclick** (tabs never do that). SR users heard "tab, selected", arrow keys did nothing. Now plain buttons with `aria-pressed` inside a `role="group"` (correct toggle-chip semantics), keeping the group `aria-label`.

## F5 (P3) — compare page didn't filter `AbortError`
`.catch(() => setLoading(false))` flipped loading off even for `AbortError`, so a rapid selection change (which aborts the prior request *after* the new one set `loading=true`) lost the "Loading…" cue and the stale table read as current. Now filters `AbortError` like every sibling fetch (analytics / inventory / home).

## Tests
- `tests/i18nInterpolate.test.ts` — the F1 `$`-pattern cases (`$$`, `$&`, `` $` ``, `$'`, `$1`), verbatim insertion.
- `tests/labelFormat.test.ts` — pins that every preset `labelKey` resolves in en+de (F3).
- F2/F4/F5 are UI-behavior fixes with no unit test (the test env is `node` — no jsdom/testing-library).

Full suite green (3233 tests). Root `tsc` + ESLint clean. The one coverage-gated change (`labelFormat.ts`) is a data-field rename (no executable-line change).


hyiger:
@codex review

hyiger:
@codex review

PR #1015, opened 2026-07-12
fix(electron/nfc): all 4 audit findings — SLIX2 mid-write, get-status contract, conn-recover, stop-mongod (#1006)

hyiger:
Closes #1006 (all 4 findings).

**F1 (P2)** — `writeSlix2Impl` caught a `writeBlock` failure at ANY block, broke the loop, and still resolved `{ success: true }`. But `wrapNdefForTag` lays live CC/NDEF data across blocks 0–78; only the last block (79, the config/lock area) is legitimately write-protected. So a tag lifted (or RF NAK) at block 20 reported success — the user pocketed a tag truncated mid-CBOR. Now only the last block (`i >= numBlocks - 1`) is the expected protected-tail case; any earlier failure rethrows.

**F4 (P3)** — the `nfc-get-status` null-service fallback returned no `lastError`, though `src/types/electron.d.ts` types it required and `NfcService.getStatus()` always includes it. Added `lastError: null` to the fallback and to preload's local `NfcStatus` interface (whose omission hid the drift from typecheck).

**F2 (P3)** — a connection-mode change caught a `startProductionServer` failure, only logged it, then unconditionally reached `loadURL` + returned `{ success: true }` — so a failed respawn (EADDRINUSE port race) left NO embedded server yet reported success and reloaded into a dead page. Now retries once on the resolved uri (resolveMongoUri already stood up this mode's mongod/sync, so the new uri is the coherent recovery target); if recovery fails too, it stops mDNS, skips the loadURL, and returns `{ success: false }`.

**F3 (P3)** — `resolveMongoUri`'s Atlas branch destroyed a leftover syncService (#672) but never stopped an embedded mongod from a prior offline/hybrid session, so pure Atlas idled a ~200–300 MB process for the session. Added `await stopLocalMongo()` to the Atlas success path (no-op when none running).

`electron:typecheck` + root `tsc` + ESLint clean. `electron/` is typecheck-gated, not part of the vitest suite; the NFC + connection-mode lifecycle is validated manually on the desktop build. The SLIX2 error-propagation change and the connection-recover path should be reconfirmed on hardware (a real block-79 protection still succeeds; earlier NAKs now fail).

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1192,6 +1192,12 @@ ipcMain.handle("nfc-get-status", (event) => {
     readerName: null,
     tagPresent: false,
     tagUid: null,
+    // GH #1006 F4: NfcService.getStatus() always includes `lastError`, and
+    // src/types/electron.d.ts types it as required (`lastError: {...} | null`,
+    // GH #450). This null-service fallback (SCardSvr stopped / pcsclite threw —
+    // the exact degraded-host path #450 surfaces errors for) must match the
+    // contract, or the renderer receives `undefined` where it's typed `null`.
+    lastError: null,
   };
 });
 
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -766,6 +766,12 @@ async function resolveMongoUri(): Promise<string | null> {
       await client.close();
 
       store.set("mongodbUri", atlasUri);
+      // GH #1006 F3: pure Atlas doesn't use the embedded mongod — stop one a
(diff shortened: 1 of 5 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #1016, opened 2026-07-12
fix(slicer): convert filament_shrink at the Orca/Bambu boundary — 0-based ⇄ 100-based (#1008 F1)

hyiger:
Part of #1008 (finding 1 of 6; F2–F6 remain).

The DB stores 0-based shrinkage (0% = none), the same convention as PrusaSlicer's `filament_shrinkage_compensation_xy`. But Orca/Bambu's `filament_shrink` is a 100-based "remaining size" (94% = the part measures 94 mm per 100; default 100% = no shrink) — and export/import treated the same number raw both ways. So (a) `shrinkageXY=0.4` exported `filament_shrink="0.4%"` (Orca reads "shrinks to 0.4% of size"), and (b) a stock Bambu `filament_shrink="98%"` imported as `shrinkageXY=98` (shown as 98% shrinkage, then re-exported into PrusaSlicer's 0-based key as nonsense).

Convert at the boundary: export emits `100 - shrinkageXY` (skipping 0 = Orca's implicit 100% default); import stores `100 - value` (100/absent → 0). `shrinkageZ` rides the PrusaSlicer-named 0-based key, so it stays raw both ways. Round-trip pinned: `shrinkageXY 0.4 → "99.6%" → 0.4`.

No data migration: a value stored under the old (raw) import can't be distinguished from a legitimate 0-based user entry, so auto-correcting would risk clobbering real data. The fix prevents future corruption; any rare pre-existing bad value is fixed by re-import or a manual edit.

**Remaining #1008 work (not in this PR):** F2 (Orca sync inheritance split — needs the route's nested-`temperatures` object restructured to dotted keys so `splitInheritedImportSet` can split per-subfield), F3 (INI import leave-when-omitted for temps + `inherits`), F4 (Bambu apply temps/bedTypeTemps inheritance), F5 (hot_plate/bed overload), F6 (OT3D scan → create uses range-max as everyday temp).

### DIFF
--- a/src/lib/bambuStudioImport.ts
+++ b/src/lib/bambuStudioImport.ts
@@ -297,9 +297,13 @@ export function parseBambuStudioProfile(raw: unknown): BambuParseResult {
     settings: {},
   };
 
-  // Shrinkage — "0.5%" → 0.5
+  // Shrinkage. GH #1008 F1: `filament_shrink` is Orca/Bambu's 100-based
+  // "remaining size" (94% → 6% shrink; 100% or absent → no shrink), so convert
+  // to the DB's 0-based shrinkage: `100 - value`. A stock profile's default
+  // 100% (or "98%") thus stores 0 (or 2), not a bogus 100/98. `shrinkageZ` uses
+  // the PrusaSlicer-named 0-based key, so it stays raw.
   const shrink = num(json.filament_shrink);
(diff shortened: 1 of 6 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #1017, opened 2026-07-13
fix(import,nfc): INI import stops nulling temps+inherits; create-from-tag uses recommended temp (#1008 F3+F6)

hyiger:
Part of #1008 (findings 3 and 6; F2/F4/F5 follow in a sibling PR).

## F3 (P2) — bulk INI import nulls all four temps (and `inherits`) on a name-matched filament
`parseIni` always materialises `temperatures` as `{nozzle: null, …}` and `inherits: null` when the keys are absent, and the non-hinted collapse branch in `prusaSlicerBundle.ts` only applied the #950 leave-when-omitted guard to the scalar fields — so the all-null temps rode through, got flattened to dot-keys, and `$set` every stored temp to null. Real PrusaSlicer user presets are `inherits = <system preset>` + only diverged keys, so importing such a bundle over an existing filament wiped its temps while reporting "updated".

The collapse now **clones** `temperatures` (the spread shared it by reference — deleting in place would corrupt the caller's parsed source) and deletes each subfield whose source INI key is absent (`temperature`→nozzle, `first_layer_temperature`→nozzleFirstLayer, `bed_temperature`→bed, `first_layer_bed_temperature`→bedFirstLayer), dropping the whole key when empty; `inherits` is dropped when absent. An **explicit** `temperature = nil` / `inherits = nil` still writes null through (parseIni stores nil as present-with-null, so the `in` check distinguishes omitted from nil). Same `inherits` guard on the hinted branch.

## F6 (P3) — OpenTag3D scan → create stores the range MAX as the everyday nozzle temp
For OpenTag3D, `decoded.nozzleTemp = maxPrintTemp ?? recPrintTemp` (the Extended max when present) while the Core recommended value survives only in `aux.opentag3d_recommended_print_temp_c`. Create-from-tag mapped `nozzle` AND `nozzleRangeMax` from `decoded.nozzleTemp` — so a filament written with nozzle=215/rangeMax=230 scanned back and created with `temperatures.nozzle = 230` (every downstream consumer then uses the ceiling). `decodedTagToFilamentPayload` now prefers the preserved recommended values (`opentag3d_recommended_print_temp_c` / `_bed_temp_c`, coerced via a finite-number guard) for the everyday nozzle/bed temps, keeping `nozzleRangeMax = decoded.nozzleTemp`. OpenPrintTag/Bambu tags carry no such aux keys — unchanged.

Note: when a recommended **bed** value exists, the Extended max bed temp has no schema home (`bedRangeMin/Max` don't exist) and is not stored — matching the issue's prescribed fix; documented in a code comment.

## Tests
- 9 collapse unit cases in `prusaSlicerBundle.test.ts` (incl. the by-reference-clone pin and explicit-nil semantics), 3 end-to-end DB-backed cases in `iniImportApply.test.ts` through the real parse→collapse→upsert pipeline (stored temps + `inherits` survive a diverged-key-only user preset), 4 cases in `decodedTagToFilament.test.ts` (recommended-wins, fallback, string coercion, junk rejection).
- Full suite green (3280 tests) pre-rebase; 253 targeted tests green post-rebase onto the #1016 merge. `tsc` + ESLint clean.

### DIFF
--- a/src/lib/decodedTagToFilament.ts
+++ b/src/lib/decodedTagToFilament.ts
@@ -1,6 +1,19 @@
 import type { DecodedOpenPrintTag } from "./openprinttag-decode";
 
 /**
+ * GH #1008 F6: coerce an `aux` value to a finite number. Aux values ride
+ * unvalidated client JSON (`Record<string, unknown>`), so they may arrive as
+ * strings or junk — only numbers and non-empty numeric strings pass; anything
+ * else (booleans, objects, arrays, "", NaN, Infinity) returns null.
+ */
+function auxTemp(v: unknown): number | null {
+  if (typeof v !== "number" && typeof v !== "string") return null;
+  if (typeof v === "string" && v.trim() === "") return null;
+  const n = Number(v);
+  return Number.isFinite(n) ? n : null;
+}
+
+/**
  * Map a tag decoded by `POST /api/nfc/decode` (a `DecodedOpenPrintTag`) into a
  * Filament DB creation payload — the server-side mapper behind create-from-scan
  * (mobile Phase 2, plan §4.4). The phone never reproduces this mapping: it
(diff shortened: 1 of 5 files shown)


hyiger:
@codex review

PR #1018, opened 2026-07-13
fix(slicer): stop Orca/Bambu syncs pinning variant inheritance + hot-plate bed guard (#1008 F2+F4+F5)

hyiger:
Part of #1008 (findings 2, 4, 5). With #1016 (F1) and #1017 (F3+F6), this completes the issue.

## F2 (P2) — `POST /api/filaments/{id}/orcaslicer` wrote flattened variant values verbatim
The only slicer sync path with no inheritance split: the Orca export flattens a variant through `resolveFilament`, and the sync-back `$set` the parent's density/cost/temps/settings onto the variant as local overrides — silently severing GH #106 live inheritance (parent edits stopped propagating). Temps are now written as **dotted** `temperatures.<sub>` paths (the split helper only splits dotted keys — a nested object passed through whole; dotted paths also stop update-validators dragging every *stored* temp into validation, the #859 class), the full active parent is fetched once (feeding both the existing nozzle-range inversion check and the split), and the update runs through `splitInheritedImportSet` with `$set`/`$unset` composition — exactly the PrusaSlicer per-id route's pattern (#951). The response `updated` reports the post-split keys.

## F4 (P3) — Bambu sync pinned inherited `temperatures` and `bedTypeTemps`
`setIfNotInherited` guarded scalars only. The temps merge now resets each parent-equal merged subfield to **null in the nested object** (the create-path-safe inherit sentinel — the bulk route spreads the update into `Filament.create`, where dotted keys would be silently dropped by strict mode, so no `$unset`); a merged `bedTypeTemps` array that deep-equals the parent's effective array (order-insensitive) collapses to `[]` (array-fallback inheritance per GH #106/#477) instead of materializing a copy.

## F5 (P3) — `hot_plate_temp` doubles as `temperatures.bed` (import-side guard, per product decision)
Export unchanged (the plate value correctly wins in the preset — Bambu's default plate IS the hot plate). On import/sync, when the parsed profile carries a "Hot Plate" `bedTypeTemps` entry and the target already has an **effective** bed temp (own ?? parent — an inheriting variant counts), the hot-plate value no longer overwrites `temperatures.bed`/`bedFirstLayer`; it lands only in the Hot Plate entry. Both DB values now survive a re-sync; fresh imports still seed bed from `hot_plate_temp`. One pre-existing route test pinned the old overwrite (the exact data loss the finding reports) and was updated with the rationale in-test.

## Tests
4 new F2 route tests (no-pin + parent-edit propagation, stale-pin `$unset` + divergent write, empty-`$set`+`$unset` edge, standalone no-regression), 21 new `bambuStudioApply` unit tests (F4 temps/bedTypeTemps + all F5 cases). Full suite green (3282 tests) pre-rebase; 319 targeted tests green post-rebase onto the #1016 merge. `tsc` + ESLint clean.

### DIFF
--- a/src/app/api/filaments/[id]/orcaslicer/route.ts
+++ b/src/app/api/filaments/[id]/orcaslicer/route.ts
@@ -11,6 +11,7 @@ import { errorResponse, errorResponseFromCaught } from "@/lib/apiErrorHandler";
 import { assertSameOriginRequest } from "@/lib/requestGuard";
 import { mergeSlicerSettings } from "@/lib/slicerSettings";
 import { isUpdateNozzleRangeInverted } from "@/lib/temperatureRange";
+import { splitInheritedImportSet } from "@/lib/importFilaments";
 
 /**
  * Top-level body keys that map to structured Filament DB fields.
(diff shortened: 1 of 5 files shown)


hyiger:
@codex review

hyiger:
@codex review

PR #1022, opened 2026-07-22
fix(slicer): compatibleNozzles ticks no longer hide presets in PrusaSlicer (#1021)

hyiger:
Fixes #1021.

## Root cause
Three mechanisms chained: (1) since #872, every exported preset derived `compatible_printers_condition = nozzle_diameter[0]==D [or …]` from the filament's **Compatible Nozzles** checkbox diameters; (2) PrusaSlicer hides presets whose condition doesn't match the active printer profile's nozzle diameter (unless "show incompatible presets" is on); (3) variants inherit the parent's `compatibleNozzles` when their own list is empty (`resolveFilament`, empty === inherit per GH #106).

Net effect, exactly as reported: a parent whose ticked diameters exclude the active printer's nozzle vanishes from PrusaSlicer **along with all its variants**; unlinking a variant makes the ex-variant reappear (its own empty list → no condition) while the ex-parent stays hidden. Filaments with no variants disappear the same way via their own ticks. Not a v1.67 regression — designed-in since #872 (~v1.63), but the reporter experience shows the default is surprising: ticking compatible nozzles reads as bookkeeping, not "hide me everywhere else".

## Fix (option B, per maintainer decision)
Remove the `compatibleNozzles` → condition derivation entirely. The nozzle-diameter condition remains **only where it's structurally meaningful**: the per-nozzle calibrated fan-out presets, whose baked values are tuned for exactly one nozzle. Unchanged: round-tripped user pins (non-empty strings) and PrusaSlicer's `nil` marker (null) still pass through the settings bag; the empty `compatible_printers` / `compatible_printers_condition` "no restriction" defaults still apply — CLAUDE.md's existing "both empty by default" description now holds without exceptions for uncalibrated presets.

## Tests
The four derivation tests are rewritten as #1021 regression pins (ticks yield an empty condition in every input shape; an empty round-tripped condition stays empty; the derivation's input-sanitizing branches are gone with it). Pin-wins, nil-preserved, calibration-bake, and per-nozzle fan-out bundle assertions unchanged. Full suite green (3311 tests); `tsc` + ESLint clean.

### DIFF
--- a/electron/sync-service.ts
+++ b/electron/sync-service.ts
@@ -654,7 +654,13 @@ export class SyncService extends EventEmitter {
       const toEnqueue = Array.from(merged.values());
       let enqueued = toEnqueue.length === 0;
       let queueDbUsed: Db = localDb;
-      if (toEnqueue.length > 0 && !this.aborted) {
+      // Codex P2 r28: the enqueue is deliberately NOT abort-gated. It
+      // persists the cleanup INTENT for copies the (already completed)
+      // filament sync made this cycle — skipping it on a late destroy()
+      // would strand equal-timestamp pairs that no replacement service could
+      // ever rediscover through LWW. Only the destructive CLEARS below honor
+      // the abort; this is bookkeeping for writes that already happened.
+      if (toEnqueue.length > 0) {
         // Codex P2 r27: durable intent must survive SERVICE RECREATION, not
         // just this process — try the local queue, then fall back to the
         // REMOTE db's queue (the drain reads both). Only when BOTH databases
@@ -683,7 +689,7 @@ export class SyncService extends EventEmitter {
           );
         }
       }
-      if (enqueued) {
+      if (enqueued && !this.aborted) {
         for (const queued of toEnqueue) {
           if (this.aborted) break;
           try {
--- a/electron/sync-service.ts
+++ b/electron/sync-service.ts
@@ -584,11 +584,31 @@ export class SyncService extends EventEmitter {
                 .toArray()
             : [];
           if (deriveLegacyNozzleCondition(nozzleDocs) !== entry.observed) continue;
-          // Same clock-skew guard as the in-transit bump: the copy carried
-          // the SOURCE's updatedAt, which a fast-clocked peer may have
-          // stamped in our future — the cleared doc must land strictly newer.
-          const observedTs =
-            entry.observedUpdatedAt instanceof Date ? entry.observedUpdatedAt.getTime() : 0;
+          // Codex P1 r24: bumping the target makes the COPIED SNAPSHOT
+          // authoritative for the pair, so first prove the SOURCE row hasn't
+          // moved since the copy — a concurrent source edit would otherwise
+          // be replaced wholesale by the stale snapshot on the next cycle.
+          // If it moved, skip entirely: the newer source replaces the target
+          // next cycle and is re-judged in that transit.
+          const sourceRow = await sourceSide.collection("filaments").findOne(
+            { syncId: entry.syncId },
+            { projection: { updatedAt: 1, "settings.compatible_printers_condition": 1 } },
+          );
+          const sourceCondition = (
+            sourceRow?.settings as { compatible_printers_condition?: unknown } | undefined
+          )?.compatible_printers_condition;
+          const observedTs = SyncService.readTimestamp(entry.observedUpdatedAt) ?? 0;
+          if (
+            !sourceRow ||
+            sourceCondition !== entry.observed ||
+            (SyncService.readTimestamp(sourceRow.updatedAt) ?? 0) !== observedTs
+          ) {
+            continue;
+          }
+          // EXACTLY observed+1ms, not wall-clock (Codex P1/P2 r24) — the
+          // same rationale as the in-transit bump: smallest stamp that beats
+          // the stale source while losing to any genuine later edit, with
+          // the timestamp parsed the way LWW itself parses it.
           await targetSide.collection("filaments").updateOne(
             {
               syncId: entry.syncId,
@@ -598,7 +618,7 @@ export class SyncService extends EventEmitter {
             {
               $set: {
                 "settings.compatible_printers_condition": "",
-                updatedAt: new Date(Math.max(Date.now(), observedTs + 1)),
+                updatedAt: new Date(observedTs + 1),
               },
             },
           );
@@ -1570,12 +1590,18 @@ export class SyncService extends EventEmitter {
             // cleaned doc back over the source — source-side convergence with
             // no bespoke second write, retried by construction every cycle
             // (an equal-timestamp pair would freeze the divergence instead).
-            // A later user edit on either side is newer still and wins.
-            // max(now, source+1ms): a peer whose clock runs AHEAD of this
-            // machine stamped the source in our future — a plain `now` would
-            // land OLDER and the stale source would win the pair right back.
-            const sourceTs = doc.updatedAt instanceof Date ? doc.updatedAt.getTime() : 0;
-            doc.updatedAt = new Date(Math.max(Date.now(), sourceTs + 1));
+            // EXACTLY source+1ms, deliberately NOT wall-clock (Codex P1 r24):
+            // only intra-pair ordering drives the convergence, and +1ms is
+            // the smallest stamp that beats the stale source while losing to
+            // ANY genuine later edit on either side (a same-peer edit after
+            // the source stamp always carries a later same-clock timestamp;
+            // a wall-clock bump could leap PAST such an edit and erase it).
+            // Parsed via readTimestamp (Codex P2 r24) — raw rows can store
+            // updatedAt as an ISO string / epoch number, and treating those
+            // as 0 would land the bump in 1970 and hand the pair straight
+            // back to the stale source.
+            const sourceTs = SyncService.readTimestamp(doc.updatedAt) ?? 0;
+            doc.updatedAt = new Date(sourceTs + 1);
           }
         } else if (doc.parentId != null && typeof doc.syncId === "string" && doc.syncId) {
           deferredParentChecks.push({
--- a/electron/sync-service.ts
+++ b/electron/sync-service.ts
@@ -8,6 +8,19 @@ import {
   type MinimalDb,
 } from "../src/lib/legacyNozzleConditions";
 
+/** GH #1021 r25/r26: one pending legacy-condition transit clear — direction,
+ * syncId, the observed condition + updatedAt (the conditional-write filter),
+ * and the provenance to revalidate on every attempt (own refs, else the
+ * source parentId). Persisted in the local `_migrations` queue. */
+interface LegacyTransitEntry {
+  d: "toLocal" | "toRemote";
+  s: string;
+  c: string;
+  u: unknown;
+  p: unknown;
+  r: unknown[] | null;
+}
+
 /**
  * Recognise a duplicate-key error specifically on the `syncId` index,
  * so the local-only push / pull paths can treat a concurrent peer
@@ -157,6 +170,11 @@ export class SyncService extends EventEmitter {
   };
   private intervalId: ReturnType<typeof setInterval> | null = null;
   private syncing = false;
+  /** GH #1021 r26: candidates whose durable enqueue failed — carried across
+   * cycles in memory so the enqueue is retried until it lands (an
+   * equal-timestamp pair never re-copies, so the transform alone cannot
+   * rediscover them). Cleared once the batch enqueue succeeds. */
+  private pendingLegacyCandidates: LegacyTransitEntry[] = [];
   // #823: set by destroy() so an in-flight cycle stops converging both DBs
   // after the user switches connection mode. Checked before each collection
   // step and each repair pass; the collection already executing finishes (its
@@ -335,11 +353,13 @@ export class SyncService extends EventEmitter {
         }
       }
 
-      // GH #1021 (Codex P2 r23 / r25): drain the durable transit-clear queue.
-      // A prior cycle's pair-clear that failed halfway left its entry queued
-      // in the local `_migrations` collection; re-attempt it every cycle (the
-      // conditional filters make retries safe forever) and dequeue once
-      // neither side matches the observed state anymore.
+      // GH #1021 (Codex P2 r23 / r25 / r26): drain the durable transit-clear
+      // queue. A prior cycle's pair-clear that failed halfway left its entry
+      // queued in the local `_migrations` collection; re-attempt it every
+      // cycle and dequeue once neither side matches the observed state.
+      // Every replay REVALIDATES the provenance persisted with the entry
+      // (r26) — parent/nozzle drift since the enqueue drops the entry
+      // without touching either row.
       if (!this.aborted) {
         const pendingDoc = await localDb.collection("_migrations").findOne(
           { _id: "legacyTransitClears" as never },
@@ -360,9 +380,18 @@ export class SyncService extends EventEmitter {
             ).catch(() => {});
             continue;
           }
-          const entry = raw as { d: string; s: string; c: string; u: unknown };
+          const entry = raw as { d: string; s: string; c: string; u: unknown; p: unknown; r: unknown };
           try {
-            if (await this.attemptTransitClearPair(localDb, remoteDb, entry)) {
+            if (
+              await this.attemptTransitClearPair(localDb, remoteDb, {
+                d: entry.d,
+                s: entry.s,
+                c: entry.c,
+                u: entry.u,
+                p: entry.p ?? null,
+                r: Array.isArray(entry.r) ? entry.r : null,
+              })
+            ) {
               await localDb.collection("_migrations").updateOne(
                 { _id: "legacyTransitClears" as never },
                 { $pull: { entries: entry } } as never,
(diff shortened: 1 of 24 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

Round-5 response (both P1s), commit 92378d37:

1. **Claim-first serialization** — the cleanup moved to a shared helper, `src/lib/legacyNozzleConditions.ts`: the durable `_migrations` marker is now inserted (unique `_id`) **before** the destructive `updateMany`, so processes racing the first connect serialize on the insert — the loser skips entirely and can never erase a pin authored after the winner's run. On a transient clear failure the claim is released so the next connect retries; an uncompleted claim is never taken over (crash-mid-clear leaves recoverable residual legacy values, strictly safer than any re-run that could eat a post-upgrade pin).

2. **Hybrid remote propagation** — `electron/sync-service.ts` now runs the same marker-guarded helper against the **remote** DB at the top of each `sync()` cycle (best-effort, non-fatal). The helper is deliberately driver-level (takes a bare `Db`) so dbConnect and the sync service share one implementation; each side cleans its own database, which is the only correct posture given the raw update preserves `updatedAt` (LWW can't carry it) and an Atlas edit would otherwise push the stale value back.

Tests: new `tests/legacyNozzleConditions.test.ts` (claim race via duplicate-`_id`, skip-if-claimed, claim release + retry, release-failure crash state — helper at 100% stmts/branches), mongodb.test.ts transient test retargeted at the raw driver handle, and a sync-service integration test pinning the remote clear + post-cleanup pin survival across cycles. Full suite 3324/3324 green.

hyiger:
@codex review

hyiger:
@codex review

Round-6 response (all four P1s), commit f298cccd:

1. **Provenance, not syntax** — the cleanup now clears a row only when its stored condition **byte-equals the removed exporter's derivation** from the row's effective `compatibleNozzles` (own list, else the parent's — the same `resolveFilament` resolution the exporter used). The derivation is frozen verbatim as `deriveLegacyNozzleCondition()` so it keeps reproducing the historical bytes. A pre-upgrade user pin that doesn't match its ticks' derivation survives; the residue in both directions (a pin byte-identical to the current derivation is indistinguishable and clears; a machine value whose ticks were edited post-round-trip survives, visible and hand-clearable) is documented in the helper header.

2. **Live claims make contenders wait** — an uncompleted marker no longer skips: the contender polls until the winner completes (→ already-done), the claim is released (→ takes its own attempt), or `waitMs` expires (→ throws `LegacyCleanupInProgressError`, so dbConnect keeps its retry state and sync aborts the cycle — nothing treats the DB as clean while an updateMany is in flight). Only a claim older than `staleMs` (crashed claimer) skips permanently; there is still no takeover.

3. **Both sides cleaned in sync()** — the cycle now runs the helper against the **local** DB too, before any collection sync (hybrid startup fires the first cycle before the Next server's dbConnect, so a locally newer legacy doc would otherwise LWW-push the stale value over the just-cleaned remote and freeze at equal timestamps).

4. **Cleanup failure = prerequisite failure** — the try/catch was removed; a cleanup failure on either side throws into `sync()`'s outer catch, aborting the cycle (error status, retry next cycle) instead of syncing stale values around a half-cleaned pair.

Tests: helper suite rewritten for provenance + wait semantics (user-pin mismatch, no-ticks pin, variant-through-parent provenance, wait-completes/-released/-timeout, stale claim, race loop) — 100% stmts/branches/funcs/lines on the helper; mongodb.test.ts pins PreUpgradePin survival through dbConnect; the sync integration test covers both sides' cleanup plus a surviving remote user pin. Full suite 3330/3330 green.

hyiger:
@codex review

Round-7 response (all three P1s), commit 1612c53b:

1. **Nozzle refs resolved for provenance** — you were right that this was a silent no-op: `compatibleNozzles` stores ObjectId refs and the removed exporter populated them before reading `.diameter`. The helper now resolves the effective ref lists through one indexed join against the `nozzles` collection (raw BSON `$in`, string-keyed dedupe) and feeds `deriveLegacyNozzleCondition` the populated shape; a dangling ref contributes nothing, exactly like `populate()` yielding null. Tests reseeded onto the real ref shape, including dangling/junk/duplicate refs and variant-through-parent ref resolution.

2. **Per-row conditional clears** — each destructive write is now an `updateOne` filtered on `{_id, condition: <the exact value the scan observed>}`, so a filament edited by an ordinary client between scan and write keeps its new value (claiming only serializes cleanup runners). A TOCTOU test pins this: a condition rewritten after the scan survives with `cleared: 0`.

3. **dbConnect rethrows** — a cleanup failure (including `LegacyCleanupInProgressError`) now fails the CURRENT request instead of serving against a not-terminally-cleaned DB; the flag-false path still retries on the next connect. The other startup migrations stay best-effort deliberately — they're safe to re-run, this one gates correctness (comment in the catch says so).

Full suite 3331/3331 green; helper at 100% statements/branches/functions/lines.

hyiger:
@codex review

Round-8 response, commit 8d99dbdb — durable per-row progress (your first suggested remedy; transactions are out because the embedded local mongod runs standalone, the same reason print-history carries a sequential fallback):

- The claim document now carries `clearedIds`. Each row is **recorded there BEFORE its conditional clear** (rolled back best-effort only if that write throws), and a resume filters recorded rows out of the candidate scan — so a retry never re-examines a row whose clear went through, and a provenance-matching pin authored on it between failure and retry survives. The residual trade (hard crash exactly between record and write → that row is skipped on resume and keeps its legacy value, visible and hand-clearable) is documented as strictly safer than any path that could erase a pin.
- A failed attempt **durably marks the claim `released`** (progress kept) instead of deleting it; the next attempt — this process or any other — resumes via a **CAS takeover on the observed `claimedAt`**. Stale and malformed claims resume the same way: per-row progress is exactly what makes takeover safe, so the former skip-forever crash state (and the `claimed-elsewhere` result) is gone.
- Live claims still wait → throw `LegacyCleanupInProgressError` on timeout; dbConnect still rethrows; sync still aborts its cycle.

Tests: a direct round-8 regression (pin authored on a cleared row survives the resume while the remaining machine row still clears), released/stale/malformed/missing-`claimedAt` takeovers, a lost-takeover-CAS race, rollback of the failed row's record, and the hard-crash worst case (release AND rollback writes fail → waiters throw, stale takeover skips the recorded row and completes). Helper at 100% statements/branches/functions/lines; full suite 3333/3333 green.

hyiger:
@codex review

Round-9 response (both P1s), commit e8d1b6c0:

1. **Fencing token** — every claim acquisition (fresh insert or CAS takeover) mints a unique `claimToken`; every subsequent marker write (per-row progress record, rollback, release, completion) filters on it, checked via `matchedCount` ("does the claim still carry OUR token", not "did the write modify anything"). A holder taken over past `staleMs` is fenced out at its next record write — BEFORE the row's destructive clear — and aborts with `LegacyCleanupInProgressError`; its release/completion writes no-op so the new holder's state stands, and completion reports terminal state only when the marker actually reflects it.

2. **Version predicate** — the scan now also observes `updatedAt` and the per-row clear filter re-asserts it (`?? null` matches rows that never had one), so an intervening app save blocks the clear even when the re-pinned text is byte-identical.

Tests: identical-text re-pin survival (fresh `updatedAt`, same text → `cleared: 0`), fenced-out-at-record abort (row untouched, new holder's claim not marked released), takeover-before-completion surfacing the in-progress error, and an `updatedAt`-carrying clearing row covering the non-null predicate side. Helper at 100% statements/branches/functions/lines; full suite 3336/3336 green.

hyiger:
@codex review

Round-10 response (both P1s + the P2), commit c47a9660:

1. **Ingestion guard** — the migration alone can't stop a pre-upgrade fork/INI from re-sending the stamped condition after the one-shot is spent. New `src/lib/stripLegacyNozzleCondition.ts` strips an incoming `compatible_printers_condition` (→ `""`) when it provenance-matches the target's effective ticks (own → parent refs resolved through one indexed nozzles query, only on the rare regex hit), wired at all three write boundaries: the per-id PrusaSlicer sync merge, the per-id OrcaSlicer sync merge (the shared bag rides every export as passthrough), and `buildIniUpdate` — the single choke point behind both bulk INI importers' update/resurrect/create-race paths. Fresh creates have no ticks to test against and stay unguarded (nothing provable). The predicate is exported as `isLegacyMachineNozzleCondition` from the migration module so the two halves share one definition.

2. **Completion-failure release** — the completion write is now wrapped: a transient failure releases the claim (token-fenced, best-effort) before rethrowing, so one blip no longer turns into wait-then-fail on every dbConnect until `staleMs`. All row work is recorded, so the resumed retry re-runs completion against an empty candidate set.

3. **Boundary ownership re-check (P2)** — ownership is re-asserted (marker read, token compare) between the progress record and the destructive write; the residual single-write window is documented as benign-convergent — the condition+updatedAt predicate means a late clear can only apply the exact clear the migration intended, and any user save in between bumps `updatedAt` and blocks it.

Tests: new strip suite (own/parent/dangling refs, populated-doc entries, no-ticks/non-grammar no-ops), route-level pins for all three boundaries, completion-blip + completion-crash cases, boundary re-check (token swap + vanished marker). Both src/lib files at 100% statements/branches/functions/lines; full suite 3349/3349 green.

hyiger:
@codex review

Round-11 response (all three P1s), commit 22111108:

1. **Restore re-cleanup** — new `rerunLegacyNozzleCleanupAfterRestore()` (src/lib/mongodb.ts): invalidates the durable marker (the restored dataset is a different world, so old observations are deliberately discarded wholesale) and re-runs the cleanup against the restored rows; the snapshot restore route calls it after a successful restore. A transient failure there leaves the process-local flag false, so the next dbConnect retries under the r7 prerequisite posture.

2. **Incoming-only strips** — both per-id sync strips are now gated on the sync payload actually owning `compatible_printers_condition` (`merge.settings` is seeded from the stored bag, so a partial sync must never re-judge a stored post-cleanup pin). The INI path was already incoming-only by construction — `collapsed.settings` is the section's own keys.

3. **Full-candidate observations** — the marker now records `{id, condition, updatedAt}` for EVERY examined syntactic candidate (matches and non-matches alike) in one fenced batch write before any destructive write. On resume a recorded row is eligible only while byte-identical to its observation; anything saved since (re-pin, tick edit) bumps `updatedAt`, mismatches, and is skipped forever. This replaces the r8 per-row record/rollback outright — a thrown clear leaves the row recorded and the retry finishes it while untouched (strictly better than the old skip-forever residue), while touched rows are never re-judged by any later attempt.

Tests: restore re-clean (stale completed marker + pre-upgrade snapshot rows, non-matching pin preserved), partial-sync pin preservation on both sync routes, the generalized resume test (re-pinned / changed-since-examined / untouched rows), the hard-crash case reworked to the finish-untouched-row outcome, and malformed observation entries ignored. Both src/lib files at 100% statements/branches/functions/lines; full suite 3351/3351 green.

hyiger:
@codex review

Round-12 response (both P1s), commit c3806b75:

1. **Snapshot cleanup provenance** — the snapshot GET now stamps `legacyNozzleCleanupComplete` from the durable marker, and restore honors it: a post-cleanup backup's byte-identical pure nozzle condition is a user pin and is NOT re-judged (the marker is just upserted to completed so no later first-connect re-judges the restored rows either); pre-cleanup or unknown-provenance backups — including every older snapshot version, where the field is absent — get re-cleaned as before.

2. **Race-safe ordering** — the re-clean deletes the marker BEFORE flipping the process-local flag. A dbConnect racing the old window could observe (flag false + completed marker), return already-done, and flip the flag back to true — cancelling the retry if the cleanup then failed. With the marker gone first, a racer either early-returns on the still-true flag or joins the cleanup through the claim protocol; a failed re-clean now reliably leaves the flag false so the next connect retries.

Tests: GET provenance stamping (marker present/absent), a post-cleanup snapshot restore preserving the byte-identical pin + marker upsert, and a deterministic race regression that fires a dbConnect inside the marker delete, fails the cleanup, and asserts the flag stays false with the next connect finishing the clean. Full suite 3354/3354 green; both src/lib helper files still 100% on all four coverage metrics.

hyiger:
@codex review

Round-13 response, commit 61606083:

**Snapshot schema bumped to v5** — no collection change; the bump exists precisely so a pre-#1022 build rejects a provenance-carrying file via its GH #953 newer-version guard (`version > CURRENT` → 400 before any wipe, verified present in released builds) instead of restoring it, dropping `legacyNozzleCleanupComplete`, and leaving a post-upgrade migration to re-judge a byte-identical pin. The version-history comment documents v5, the #953 rejection test now probes with v6, and the GET provenance test pins `version === 5`. Full suite 3354/3354 green.

hyiger:
@codex review

Round-14 response, commit fe690ead:

**Bambu import paths guarded** — `prepareBambuUpdate` (the shared apply behind both `POST /api/filaments/bambustudio` and `POST /api/filaments/{id}/bambustudio`) now runs the same `stripLegacyMachineCondition` guard as the other three boundaries: incoming-only (gated on the parsed profile owning `compatible_printers_condition`, so a profile omitting the key never re-judges a stored post-cleanup pin), provenance-matched against the target's effective ticks, creates unguarded. Both routes' hand-picked existing-doc projections now carry `compatibleNozzles` so the guard can see the ticks — the per-id route's projection was the reason a naive helper call silently no-opped, which the new route-level test caught.

Tests: per-id Bambu strip + mismatched-pin preservation (71/71 in the Bambu files). Note: the local full-suite run was disrupted by machine overload (mongod launch/hook timeouts across 60 files including DB-free pure tests) — targeted files are green and CI runs the full gate on clean runners.

hyiger:
@codex review

Round-15 response (all three P2s), commit ad7b0286:

1. **Parent-drift-safe provenance** — observations now record the derivation at examination time (`d`), and a resumed row is eligible only when its current derivation equals the recorded one (covers both directions of parent drift across attempts; entries without `d` are unverifiable and skip). Within an attempt, parent-provenance rows get a **pre-clear revalidation** — re-read the ACTIVE parent, re-derive, skip on drift — with the same benign-convergent residual posture as the r10 ownership re-check. Own-tick rows need neither: their tick edits bump the child's own `updatedAt`, which the conditional write re-asserts atomically.

2. **Active-parent lookup** — the cleanup's parent join now filters `_deletedAt: null`, matching `stripLegacyMachineCondition` and the export path's actual resolution (a deleted parent derives nothing, so a matching pin on an orphaned variant is preserved).

3. **Non-mutating INI strip** — `buildIniUpdate` strips a per-invocation clone of the section's settings, so a strip decision based on one phase's target can't leak into the trashed-row / create-race fallbacks or the create path; a route-level test pins that the caller's payload object is untouched.

Tests: drift-resume skips in both directions, deleted-parent skip, mid-run parent-edit and parent-delete revalidation skips, a parent-provenance resume clear, and the INI non-mutation pin. Both src/lib files at 100% statements/branches/functions/lines; full suite 3360/3360 green on an idle machine (the earlier run's failures were confirmed environmental — they did not reproduce).

hyiger:
@codex review

Round-16 response (the P1 + the P2), commit 6f088c1c:

1. **Honest invalidation** — the restore hook's durable invalidation is now one atomic marker conversion (`completed` → `released`, prior-world observations wiped, `claimToken` unset so an in-flight pre-restore runner is fenced out). If that single write fails, nothing has changed, and the new typed `RestoreCleanupInvalidationError` surfaces: the snapshot route returns **500** telling the user to re-run the idempotent restore, instead of reporting success with a retry that could never happen (flag true + completed marker survive even restarts). The post-cleanup branch's marker upsert has the same posture. Failures AFTER the durable invalidation keep the success-with-retry behavior — which is now truthful, since the released marker + false flag genuinely retry.

2. **Fresh-doc revalidation for every row** — the pre-clear revalidation now re-fetches the nozzle DOCS fresh for own-tick rows too (a referenced nozzle's diameter edit or purge changes the derivation without touching the filament row); parent-provenance rows additionally re-read the active parent's ref list. Any derivation drift skips the row.

Tests: invalidation-failure unit coverage (typed error, marker + flag untouched, both branches), a route-level 500 with the data restored and the stale marker intact, an own-tick mid-run diameter-edit skip, and the r12 race regression retargeted at the invalidation write. Both src/lib files at 100% statements/branches/functions/lines; full suite 3363/3363 green.

hyiger:
@codex review

Round-17 response, commit ecaa3042:

**In-transit strip on the hybrid LWW path** — `buildFilamentRefsTransform` now applies the same provenance-matched strip as the slicer/import boundaries to every copied filament (both the LWW replace and the create paths run the transform), resolved against the **source** side — whose ticks are the world the stamp was made in: own refs, else the ACTIVE source parent's, with diameters from the source-side nozzle map. It runs before the ref remap (it needs source ids); missing provenance (dangling refs, deleted parent) strips nothing — the conservative direction — and only the in-transit copy is stripped, never the source row (its own database's cleanup owns it). The nozzle/filament prefetches carry `diameter` and `compatibleNozzles`/`_deletedAt` projections to feed it, refs only.

Test: with both sides' markers completed, a newer remote doc carrying the machine condition lands locally stripped (sibling keys intact), a non-matching pin rides through verbatim, and the source row is untouched. Full suite 3364/3364 green.

All previously-raised review threads (rounds 1–17) are addressed in HEAD and have been marked resolved.

hyiger:
@codex review

Round-18 response (the P1 + the P2), commit f0d52573:

1. **Full-document create guard** — `POST /api/filaments` now runs the same provenance-matched `stripLegacyMachineCondition` before `Filament.create`. The shared-catalog import flow (`buildFilamentImportBody`) sends exactly the vulnerable shape — remapped tick refs plus the source's settings bag verbatim — and the body is the incoming payload by construction, so the strip is incoming-only; Mongoose casts the body's string refs in the helper's lookups. Creates without ticks (INI/Bambu bulk creates, create-from-tag) still have nothing provable; a non-matching pure nozzle condition persists as a user pin. Route-level test pins both directions.

2. **Abort re-check** — the sync cleanups are now gated on `this.aborted` after the clients connect, so a `destroy()` landing mid-connect can no longer let destructive cleanup writes hit the abandoned databases (matching the mode-switch contract every later step honors).

Full suite 3365/3365 green. All previously-raised review threads are addressed in HEAD and marked resolved.

hyiger:
@codex review

Round-19 response (the P1 + the P2), commit 050a1bd1:

1. **Source row cleaned too** — the transform records every in-transit strip (`{direction, syncId, observed value, observed updatedAt}`), and after the filaments collection sync the cycle clears each SOURCE row with the same conditional predicate as the one-shot cleanup (exact value + exact `updatedAt`; raw driver, no timestamp bump) — both sides then converge on the stripped value at equal timestamps instead of the source serving the stale condition to its own exports/snapshots forever. Best-effort per row: on a transient failure the target copy is already clean, and any later source edit re-triggers a copy + a fresh strip/clear.

2. **Per-side abort checks** — `this.aborted` is re-checked before EACH side's cleanup (and before each source-side clear), so a `destroy()` landing while the first side's cleanup awaits can no longer let the second side's destructive write proceed.

The r17 transit test now asserts the source row is cleaned (single key, sibling intact), the pair converges byte-equal at the same `updatedAt`, and a non-matching pin stays on both sides. Full suite 3365/3365 green. All prior review threads resolved.

hyiger:
@codex review

Round-20 response, commit 3c788038:

**Atlas imports guarded** — `POST /api/filaments/import-atlas` now resolves provenance against the SOURCE database BEFORE force-emptying the refs: the remote doc's own ticks, else its ACTIVE source-side parent's, with diameters read from the source's `nozzles` collection. A provenance-matching machine condition is stripped into a cloned bag; a non-matching pure nozzle condition imports as a user pin; only regex hits pay the extra source reads. The source Atlas stays read-only by contract — its own deployment's migration owns cleaning it.

Test: new memory-server-as-remote suite (GH #626 harness pattern) covering the own-ticks strip, mismatched-pin pass-through, and parent-provenance strip. Full suite 3366/3366 green. All prior review threads resolved.

hyiger:
@codex review

Round-21 response, commit adfd3c89:

**Loop-until-dry cleanup** — the scan/record/clear cycle now loops until a whole pass finds nothing new to examine and nothing to clear (that empty pass IS the verification scan); earlier passes' observations fold into the in-memory map so later passes surface only genuinely new rows. A database still receiving legacy writes at the pass cap (5) throws the transient in-progress error — claim released, retried later — instead of completing over unexamined rows. The residual (a row landing between the final empty scan and the completion write) is caught by the hybrid transit strip + source-side clear on the next sync cycle (r17/r19).

Tests: a mid-cleanup arrival is examined and cleared before completion; a pathological continuous writer hits the cap, throws, and leaves the claim released. Helper at 100% on all four metrics; full suite 3368/3368 green. All prior threads resolved.

hyiger:
@codex review

Round-22 response, commit e1bd5d3e:

**Deferred parent-provenance revalidation** — own-tick rows keep the in-transit strip (their diameters come from the nozzle map, which is stable for the whole pass since nozzles sync before filaments), but parent-provenance rows are now recorded by the transform and revalidated AFTER the filaments collection sync against the **current** source parent + fresh nozzle docs. Only when the condition still derives from the ticks as they now stand is it the machine stamp — then both the transferred copy and the source row are conditionally stripped (exact value + exact updatedAt, no timestamp bump); a non-matching current derivation preserves the value on both sides. The stale pre-fetched filament tick maps are gone entirely.

Tests: the deferred strip lands on both sides in the plain case, and the exact r22 regression — the same pass moving the parent's ticks away from the child's condition — preserves the pin on both sides while the parent converges to the newer tick set. Full suite 3370/3370 green. All prior threads resolved.

hyiger:
@codex review

Round-23 response (the P1 + the P2), commit 3417f42d:

1. **Single-write convergence** — the bespoke source-side clears are gone entirely. Every strip (in-transit and deferred) is now ONE write that bumps the cleaned doc's `updatedAt` to `max(now, source updatedAt + 1ms)` (clock-skew-safe), making the cleaned side strictly newer — the engine's own next LWW cycle `replaceOne`-propagates the cleaned doc onto the source. Source-side convergence is retried by construction every cycle; a two-write partial state cannot exist.

2. **Parent-state binding** — with the deferred clear reduced to a single conditional write immediately after the revalidation, the residual revalidate→write window matches the one-shot cleanup's accepted benign-convergent posture (the filter re-asserts the exact child value + `updatedAt`; the value provably derived from the ticks at write time) — and because the pair no longer sits at equal timestamps, a later user re-pin on either side is newer and wins through normal LWW instead of freezing.

Tests: transit and deferred cases assert the second cycle's LWW propagation cleans the source (both sides converge byte-equal at the bumped timestamp); the r22 parent-drift regression still preserves the pin on both sides. Full suite 3370/3370 green. All prior threads resolved.

hyiger:
@codex review

Round-24 response (the P1 + the P2), commit 6e66e3e1:

1. **Source proven unchanged + minimal bump** — the deferred clear now re-reads the SOURCE row and requires it to still carry the observed condition AND the observed `updatedAt` before writing; if it moved, the clear is skipped entirely (the newer source replaces the target next cycle and is re-judged in that transit). Both bumps are now exactly `source + 1ms` instead of `max(now, …)`: only intra-pair ordering drives the convergence, and +1ms is the smallest stamp that beats the stale source while **losing** to any genuine later edit on either side — a wall-clock bump could leap past a concurrent edit and erase it wholesale, which is precisely the hazard you flagged.

2. **LWW's own parser** — both bump sites now parse `updatedAt` through the class's `readTimestamp` (Date | ISO string | epoch number), so a string-stamped raw row can no longer collapse to 0 and land the bump in 1970.

Full suite 3370/3370 green; the transit + deferred convergence tests pin the two-cycle propagation at the bumped timestamp. All prior threads resolved.

hyiger:
@codex review

Round-25 response, commit ad0a06b0 — took the field-level alternative you offered:

**No synthetic timestamps anywhere.** The transform no longer strips in transit — copies ride verbatim, and it only records syntactic candidates (source-side provenance ids captured pre-remap). Post-sync, each candidate is judged against FRESH source-side state and, when confirmed, BOTH sides get a field-level conditional clear: exact syncId + condition + updatedAt, setting only the condition to `""` with timestamps untouched. The stale snapshot is never authoritative, and no stamp exists to tie a genuine +1ms edit — an edited row misses its filter and normal LWW owns the pair.

**Durable retry** — the pending pair is `$addToSet`-ed into the local `_migrations` collection BEFORE the clears and dequeued only once neither side matches the observed state; every cycle re-drains the queue at start (the conditional filters make retries safe forever; malformed entries are dropped rather than retried eternally). A partial failure can no longer freeze a pair at equal timestamps.

Tests: same-cycle both-sides convergence at the source's own `updatedAt` with an empty queue; a seeded frozen partial state (target cleared / source stale at equal timestamps + queued entry) drained and dequeued; the deferred parent-provenance and r22 parent-drift regressions unchanged. Full suite 3371/3371 green. All prior threads resolved.

hyiger:
@codex review

Round-26 response (both P2s), commit 773fae47:

1. **Provenance rides the queue** — every entry now persists what the transit observed (own refs, else the source `parentId`), and every attempt — same-cycle or drain replay — re-derives from fresh source-side parent/nozzle state before any write. Drifted or unverifiable provenance drops the entry without touching either row; entries without provenance are dropped rather than replayed blind.

2. **No clears before durable intent** — all of a cycle's candidates (merged with any carried over from a failed cycle) are enqueued in ONE batch `$addToSet` before any clear. If that single write fails, every clear is skipped this cycle and the candidates carry over in memory (`pendingLegacyCandidates`) so the next cycle retries the enqueue — closing the path where an equal-timestamp pair could never be rediscovered after a lost candidate.

Shared `LegacyTransitEntry` shape across the transform recording, the queue, and `attemptTransitClearPair`. The drain test entries now carry provenance, with a new drifted-provenance case pinning drop-without-clearing + dequeue. Full suite 3371/3371 green. All prior threads resolved.

hyiger:
@codex review

Round-27 response (both P2s), commit e657e1b3:

1. **Durable intent survives service recreation** — the candidate enqueue tries the local queue and falls back to the REMOTE database's queue; the drain reads BOTH queues every cycle (dequeuing from whichever holds the entry; the same-cycle loop pulls from the db that accepted the batch). Candidates stay in memory only when BOTH databases refuse the write — and both just accepted the entire collection sync moments earlier, so a double refusal means the cycle itself is failing everywhere.

2. **Partial pairs reconciled before drift drops** — `reconcilePartialTransitPair` restores the surviving observed value onto the side showing our own clear signature (`""` at exactly the observed `updatedAt` — user edits miss these filters) before a drifted/unverifiable entry is dropped, so a target-`""`/source-pin pair can no longer freeze at equal timestamps. Reconcile failures propagate so the entry stays queued and retries.

Tests: the drift case now asserts the restored pair on both sides at the original timestamp; a new remote-queue drain case pins the fallback surviving service recreation. Full suite 3372/3372 green. All prior threads resolved.

hyiger:
@codex review

Round-28 response, commit f3040636:

**Cleanup intent persists through aborts** — the candidate enqueue is now deliberately NOT abort-gated: it records the cleanup intent for copies the already-completed filament sync made this cycle (bookkeeping for committed writes, not a new destructive operation), with the r27 local→remote fallback and the in-memory carryover running regardless of the abort flag. Only the destructive clears honor the abort (gated on `enqueued && !aborted` plus the per-entry break). A late `destroy()` can no longer strand equal-timestamp pairs that a replacement service could never rediscover.

Full suite 3372/3372 green. All prior threads resolved.

hyiger:
@codex review

PR #1031, opened 2026-07-25
fix(ci): unblock the audit gate — resolve 4 advisories, allowlist the unfixable brace-expansion (#1025)

hyiger:
Fixes #1025

Five advisories published **2026-07-24** turned `scripts/audit-gate.mjs` red on `main`. Because `release.yml` and `docker.yml` both gate on `ci-gate.yml` *before* their build matrices (#852), this blocked every PR **and the not-yet-cut v1.68.0 tag**.

## Before / after

```
$ node scripts/audit-gate.mjs          # on main
Audit gate FAILED — advisories not in .audit-allowlist.json:
  app-builder-lib [high] · brace-expansion [high] · builder-util-runtime [high]
  mongoose [moderate]    · postcss [high]
exit 1

$ node scripts/audit-gate.mjs          # this branch
Audit gate passed (3 allowlisted advisories tolerated).
exit 0
```

## What actually fixed it

**Four of the five needed no manifest change at all** — only a lockfile re-resolve, because `package.json`'s ranges already permitted the patched versions:

| package | before | after | advisory |
|---|---|---|---|
| postcss | 8.5.14 | **8.5.23** | GHSA-r28c-9q8g-f849 |
| mongoose | 9.5.0 | **9.7.4** | GHSA-664h-wqgq-64gw |
| app-builder-lib | 26.8.1 | **26.15.3** | GHSA-7g7r-gx96-252g |
| builder-util-runtime | 9.5.1 | **9.7.0** | GHSA-p2f4-r6v6-j797 |
| electron-updater | 6.8.3 | **6.8.9** | (carries the above) |

`next` stays at **16.2.11** — no downgrade.

> npm's reported `fixAvailable: {name: "next", version: "9.3.3", isSemVerMajor: true}` for postcss is a **red herring**. npm computes it against `next`'s hard-pinned `postcss: 8.4.31` and ignores the `overrides` block that has been substituting postcss since #221. `npm audit fix --force` would have downgraded Next by 7 majors.

Manifest edits are **floor bumps only**, so a future clean install can't resolve back below a patched version:

- `overrides.postcss` `^8.5.10` → `^8.5.18`
- `overrides."brace-expansion@>=3.0.0"` `^5.0.7` → `^5.0.8`
- `dependencies.mongoose` `^9.3.3` → `~9.7.4`

## Why mongoose is pinned `~9.7.4` and not `9.8.0`

9.7.4 is the minimal delta past the advisory floor (`>=9.7.2`) and **holds the MongoDB driver at 7.2.0**; `^9.8.0` would pull driver 7.5 plus `fix(subdocument): don't minimize empty document array elements to null`, which touches doc-array semantics this schema leans on hard (`spools`, `calibrations`, `amsSlots`). For a PR whose job is to unblock a release tag, minimal surface is the right risk posture. A move to 9.8.0 can land separately and fail loudly on its own.

## Why brace-expansion is allowlisted rather than fixed

It is **genuinely unfixable**, and each claim in the allowlist justification was verified:

1. **No backport exists.** The advisory range is `<=5.0.7`, and `1.1.16` / `2.1.2` are the *newest* releases in their lines (checked against the registry). Only 5.0.8 is patched.
2. **The remaining copies can't be overridden.** They're pinned by `minimatch@3.1.5` (`^1.1.7`) and `minimatch@5.1.9` (`^2.0.1`). brace-expansion 5.x moved its CJS export from a callable default to a named object — 5.0.8 exports `{EXPANSION_MAX, EXPANSION_MAX_LENGTH, expand}`, while `minimatch@3.1.5` does `var expand = require('brace-expansion')` and then calls it. Forcing 5.x there throws `expand is not a function`, breaking **lint and electron-builder packaging**. This is precisely what the existing `brace-expansion@>=3.0.0` override scoping protects — it upgrades only consumers that already expect the named-export shape. **Do not widen it.**
3. **Exposure is nil.** Every glob pattern in the repo is first-party (`eslint.config.mjs`, `electron-builder.yml`). The only production chain is `exceljs@4.4.0 → archiver@5.3.2 → {glob@7 → minimatch@3.1.5, readdir-glob@1.1.3 → minimatch@5.1.9}`, and the vulnerable path requires `archive.glob()` — which exceljs never calls (no `.glob(` anywhere in `exceljs/lib`). The app uses only the jszip reader/writer in `src/app/api/filaments/{export,import}-xlsx/route.ts`.

> npm's `fixAvailable: eslint@10.8.0` for this advisory is **also wrong** — the advisory was verified to *persist* after that semver-major (the vulnerable nodes are pinned by three independent chains, not just eslint). Pure risk for zero gate benefit, so `eslint` stays at `^9`.

## Verification

Full `ci-gate.yml` replicated locally, all green:

| step | result |
|---|---|
| `npm ci` (lockfile ↔ manifest sync) | accepted, no drift |
| `npm run lint` | clean |
| `npm run electron:typecheck` | clean |
| `npm run test:coverage` | **3372/3372**, 163 files — 99.55% lines / 99.05% statements / 97.72% functions / 97.56% branches (all above thresholds) |
| `node scripts/audit-gate.mjs` | **exit 0** |
| `npm run build` | succeeds |

Not run locally: the standalone smoke job (needs the mongo service) — CI covers it.

## Residual risk — please read

**`electron-builder` 26.8.1 → 26.15.3 is the meaningful risk in this PR, and it is not covered by any check above.** 26.15.0 is the exact release that fixes the advisory *and* the one that does `feat(migration): fully replace Go binary app-builder-bin with TS implementation` — there is no lower version carrying the fix. The dependency delta confirms the backend swap: **removed** `app-builder-bin`, `7zip-bin`, `@develar/schema-utils`, `dmg-license`, `iconv-corefoundation`; **added** `pkijs`, `asn1js`, `@noble/hashes`, `@peculiar/webcrypto`, `ajv`, `unzipper`.

26.15.2 also ships `fix(mac): use native zip for macOS zip target to preserve .framework symlinks` — and that zip is exactly the artifact `electron-updater`'s `MacUpdater` consumes (#698). `electron-updater` 6.8.3 → 6.8.9 likewise modifies `MacUpdater.ts`, `AppUpdater.ts`, `BaseUpdater.ts`, `LinuxUpdater.ts`, `GitHubProvider.ts` — a surface this repo has burned on twice (#586 Windows `latest.yml`, #698 `latest-mac.yml` merge).

Nothing here can be validated without an actual signed release build. **Recommendation: after this merges, watch the first v1.68.0 release run closely and confirm `latest-mac.yml` exists and lists both `-mac-arm64.zip` and `-mac-x64.zip` before trusting auto-update** (the check CLAUDE.md already prescribes):

```bash
gh release download v1.68.0 -p latest-mac.yml && cat latest-mac.yml
```

## Out of scope

`packages/mobile` has its own red prod audit (4 high, all `fixAvailable: true`). `mobile.yml` is path-filtered and not a required check, so it blocks neither `main` nor the release tag — worth a separate PR.

### DIFF
--- a/.audit-allowlist.json
+++ b/.audit-allowlist.json
@@ -14,6 +14,13 @@
       "title": "Immutable.js Map/Set hash-collision algorithmic complexity DoS",
       "justification": "Same pinned immutable@3.x via swagger-ui-react and same first-party-spec-only exposure as GHSA-v56q-mh7h-f735.",
       "reviewBy": "2026-10-01"
+    },
+    {
+      "id": "GHSA-mh99-v99m-4gvg",
+      "package": "brace-expansion",
+      "title": "brace-expansion DoS via unbounded expansion length (OOM crash)",
+      "justification": "NO PATCHED RELEASE EXISTS below 5.0.8: the advisory range is <=5.0.7, and 1.1.16 / 2.1.2 are the NEWEST releases in their lines (verified against the registry) — there is no 1.x/2.x/3.x/4.x backport. The remaining vulnerable copies are pinned transitively by minimatch@3.1.5 (^1.1.7) and minimatch@5.1.9 (^2.0.1), and cannot be overridden: brace-expansion 5.x moved its CJS export from a callable default to a named object ({EXPANSION_MAX, EXPANSION_MAX_LENGTH, expand}), so forcing 5.x into minimatch@3.1.5 (which does `var expand = require('brace-expansion')` then calls it) throws 'expand is not a function' — breaking lint AND electron-builder packaging. This is exactly why the overrides entry is scoped 'brace-expansion@>=3.0.0': it upgrades only consumers that already expect the named-export shape. Do not widen it. Exposure: NIL. Every glob pattern in the repo is first-party (eslint.config.mjs, electron-builder.yml) — no user input reaches a brace expansion. The only production chain is exceljs@4.4.0 -> archiver@5.3.2 -> {glob@7 -> minimatch@3.1.5, readdir-glob@1.1.3 -> minimatch@5.1.9}, and the vulnerable path requires archive.glob(), which exceljs never calls (verified: no '.glob(' anywhere in exceljs/lib) — the app uses only the jszip writer/reader (workbook.xlsx.writeBuffer / workbook.xlsx.load) in src/app/api/filaments/{export,import}-xlsx/route.ts. NOTE: npm reports fixAvailable eslint@10.8.0 for this advisory; that is WRONG — verified in a sandbox lockfile that the advisory PERSISTS after the eslint semver-major, so the upgrade is pure risk for zero benefit. Re-check when minimatch ships a release depending on brace-expansion >=5.0.8.",
+      "reviewBy": "2026-10-01"
     }
   ]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "electron-updater": "^6.8.3",
         "exceljs": "^4.4.0",
         "mongodb-memory-server-core": "^11.0.1",
-        "mongoose": "^9.3.3",
+        "mongoose": "~9.7.4",
         "next": "^16.2.4",
         "qrcode": "^1.5.4",
         "react": "19.2.4",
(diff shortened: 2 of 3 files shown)


hyiger:
@codex review

PR #1033, opened 2026-07-25
fix(api): bound usage grams so one oversized entry can't null out analytics (#1030)

hyiger:
Fixes #1030

`grams` was validated as finite and non-negative at both write boundaries but never bounded **above**, and neither schema capped it. `Number.isFinite` excludes only `Infinity`/`NaN`, so `JSON.parse('{"grams":1e308}')` yielded a finite value that cast, validated and persisted.

## The failure is worse than one wrong number

Triage caught something the original issue missed. Analytics sums raw doubles; once a day's sum overflows to `Infinity`, the Hamilton apportionment computes

```
ideal = (raw / rawDaySum) * dayGrams  =  (raw / Infinity) * Infinity  =  0 * Infinity  =  NaN
```

for **every segment of that day** — and `JSON.stringify` renders both `Infinity` and `NaN` as `null`. Reproduced end to end:

```
two 1e308 entries + one innocent 42 g entry, same UTC day
-> {"grams":null,"byFilament":[{"id":"poisonA","grams":null},
                               {"id":"poisonB","grams":null},
                               {"id":"innocent","grams":null}]}
```

So an unrelated, correctly-sized filament that merely printed on the same day also came back as `null`, breaking the numeric contract `public/openapi.json` declares. (One oversized entry alone doesn't overflow — it takes two. Both are one POST each.)

## Two layers

**1. Write boundaries** reject `grams > MAX_USAGE_GRAMS` (1,000,000 g = 1 tonne) — `POST .../spools/{spoolId}/usage` and `POST /api/print-history`.

The bound is 1000× a standard 1 kg spool and ~50× the largest FDM spool sold, so no real entry approaches it — while a spool's entire ledger (1000-entry cap × 1e6) tops out at 1e9 g, far inside `Number.MAX_SAFE_INTEGER`. No aggregate can lose precision, let alone overflow.

Deliberately **not** tightened to catch unit-confusion bugs: a fork posting mm of filament (~330,000 mm/kg) or mm³ (~806,000 mm³/kg) lands just under this cap. A 100 kg bound would catch those, but it's close enough to plausible bulk backfill logging to risk rejecting a real entry. Overflow safety and unit validation are separate concerns and shouldn't be smuggled together by lowering the constant.

**2. Analytics sanitises at ingestion** (`safeGrams`), not at the six emit points. With every input finite, `rawDaySum` cannot overflow, so the NaN cascade is unreachable *by construction* and the existing `day.grams === Σ segments` invariant still holds. This is the layer that covers rows predating the cap or arriving through paths that bypass the routes — snapshot restore, hybrid sync. A clamped row renders **at the cap rather than vanishing**, so corruption stays visible instead of silently under-reporting; NaN/negative contributes 0.

## What I deliberately did *not* do

The issue proposed mirroring the bound as `max:` on both schema paths. **I skipped it, and this is the main thing to sanity-check me on.**

It would add exactly one covered path (snapshot restore) over the route caps, still miss hybrid sync, and — because Mongoose validates on every subsequent save — would **brick any spool already holding an over-cap entry** unless shipped alongside a one-time clamp migration that silently rewrites a user's ledger. Triage flagged that migration as a data-mutation decision it shouldn't make unilaterally; I agree, so I took the path that makes the user-visible symptom safe without touching anyone's data.

The ingestion sanitiser covers the same ground defensively. Happy to add the schema max + migration as a follow-up if you'd rather have it.

## Tests

**They assert the response stays numeric, not just that the 400 fires** — a status-only test would not catch a regression in the ingestion clamp, which is the half that protects existing data.

Verified non-vacuous by neutering each half:

| neutered | failure |
|---|---|
| `safeGrams` | `expected null not to be null`, `expected 5 to be 10` |
| route caps | `expected 201 to be 400` (×3) |

Coverage:
- `tests/analytics-route.test.ts` — the two-overflowing-entries case (totals + every day segment stay finite, **the innocent filament's 42 g survives intact**, the clamped row stays visible), plus negative/non-numeric stored values contributing 0.
- `tests/spool-subroute.test.ts` — 400 with nothing persisted (no ledger entry, no weight debit), cap+1 rejected, cap exactly accepted, ordinary log still fine.
- `tests/print-history.test.ts` — 400 and atomic (no PrintHistory row, no debit, no ledger entry), cap accepted, ordinary job accepted.

| check | result |
|---|---|
| `npm run lint` | clean |
| `node scripts/audit-gate.mjs` | exit 0 |
| `npm test` | **3407/3407**, 163 files |

`public/openapi.json` updated so the spec matches the handlers (`maximum` + description on both request-side `grams` schemas).

## Note

Triage also flagged that `spool.totalWeight` has the identical unbounded-magnitude hole (`src/models/Filament.ts` has `min: 0` only). Out of scope here — this issue is about the usage ledger — but worth its own issue if you want it closed.

### DIFF
--- a/public/openapi.json
+++ b/public/openapi.json
@@ -4563,7 +4563,9 @@
                 "properties": {
                   "grams": {
                     "type": "number",
-                    "exclusiveMinimum": 0
+                    "exclusiveMinimum": 0,
+                    "maximum": 1000000,
+                    "description": "Grams consumed. Bounded above (GH #1030) so a pathological magnitude cannot overflow the analytics aggregates; an over-cap value is rejected with 400."
                   },
                   "jobLabel": {
                     "type": "string",
@@ -4817,7 +4819,9 @@
                         },
                         "grams": {
                           "type": "number",
-                          "minimum": 0
+                          "minimum": 0,
+                          "maximum": 1000000,
+                          "description": "Grams consumed by this filament. Bounded above (GH #1030) so a pathological magnitude cannot overflow the analytics aggregates; an over-cap value is rejected with 400."
                         }
                       }
                     }
--- a/src/app/api/analytics/route.ts
+++ b/src/app/api/analytics/route.ts
@@ -4,6 +4,7 @@ import Filament from "@/models/Filament";
 import PrintHistory from "@/models/PrintHistory";
 import { getErrorMessage, errorResponse } from "@/lib/apiErrorHandler";
 import { displayColor } from "@/lib/filamentColors";
+import { MAX_USAGE_GRAMS } from "@/lib/capUsageHistory";
 
 /**
  * GET /api/analytics?days=30 — usage analytics aggregation.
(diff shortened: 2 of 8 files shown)


hyiger:
@codex review

PR #1034, opened 2026-07-25
fix(i18n): make interpolate() single-pass so a param value can't inject a token (#1029)

hyiger:
Fixes #1029

`interpolate()` folded over `Object.entries(params)`, re-scanning the **output** of each pass. A `{token}` sitting inside an already-substituted **value** was therefore picked up as a live token by a later param and replaced:

```js
interpolate("Delete {name}? ({count} spools)", { name: "PLA {count}", count: 7 })
//  was -> "Delete PLA 7? (7 spools)"
//  now -> "Delete PLA {count}? (7 spools)"
```

Filament / printer / location names are user-controlled and do flow into multi-param strings, so the input is reachable.

## Scope correction — the issue's stated reachability was wrong

I wrote in the issue that the delete/retire confirm dialogs were the likely victims. **They aren't** — they're all single-param and therefore immune. The genuinely multi-param, user-controlled sites are:

| site | params | user-controlled |
|---|---|---|
| `bambuImport.success` | `{ verb, name }` | `name` |
| `trash.bulkFailures` | `{ count, names }` | `names` |
| `prusament.import.createDescription` | `{ material, color }` | both |

Each is safe today **only by insertion-order luck** — a value can only inject a token belonging to a *later* param, and in all three the user-controlled value happens to come last. That accidental protection is precisely what kept the bug invisible, and it evaporates the moment someone reorders an object literal.

So: **latent, not live.** The fix removes the order dependence entirely rather than continuing to rely on it. Worth knowing when you weigh the priority.

## The fix

One `TOKEN_RE = /\{(\w+)\}/g` pass over the template, looking each token up in `params`. An inserted value is never re-examined, so reentrancy is structurally impossible.

Three things fall out of the same change:

- The param name is no longer spliced into a `RegExp` **unescaped** — `{aXb}` used to match a param keyed `"a.b"`, with `.` acting as a wildcard.
- One regex compiled per module instead of one per param per `t()` call.
- `hasOwnProperty` (not `!== undefined`) keeps an unknown token verbatim *and* stops `{constructor}` / `{toString}` resolving off `Object.prototype`.

**The GH #1007 F1 guarantee is preserved and explicitly re-pinned** — the replacer stays a *function*, so `$$` / `$&` / `` $` `` / `$'` in a value are inserted literally. There's a dedicated regression test for it, because a careless single-pass rewrite would be the obvious way to reintroduce that bug.

## On the `\w+` token class

`\w+` is exactly the grammar both locale files use — verified against **all 61 distinct tokens** in `en.json` + `de.json`:

```
$ grep -oE '\{[^}]*\}' src/i18n/locales/{en,de}.json | sed 's/.*://' | sort -u | wc -l
61
$ ... | grep -vE '^\{[A-Za-z0-9_]+\}$'
(none)
```

Triage offered `[\w.-]+` as a future-proofing alternative. I took `\w+` **plus a locale-token invariant test**, so a future dotted or hyphenated key fails **loudly** in CI rather than silently not substituting at runtime. The test says so in a comment: widen the class if it trips, don't delete the test.

## Tests

**Verified non-vacuous** — all 5 new behavioural tests fail against the old implementation:

```
× renders a {token} inside a param VALUE literally
    expected 'Delete PLA 7? (7 spools)' to be 'Delete PLA {count}? (7 spools)'
× is independent of param key order (both directions)
× does not let a value inject a token that appears EARLIER in the template
    expected 'Updated "PLA Updated"' to be 'Updated "PLA {verb}"'
× handles a value that is exactly another token
    expected 'X/X' to be '{b}/X'
× does not mis-match a param name containing regex metacharacters
    expected 'x PWNED y' to be 'x {aXb} y'
```

Plus: order-independence asserted in **both** directions, the `$`-pattern regression guard, prototype-key safety, and statelessness across calls (a module-scoped `g` regex would misbehave if it were ever switched to `.test()`/`.exec()`, which advance `lastIndex`).

| check | result |
|---|---|
| `npm run lint` | clean |
| `npm test` | **3417/3417**, 163 files |

## Docs

`CLAUDE.md`'s i18n bullet documented only the function-replacement half of the contract and warned against "simplifying" it. That warning stays valid but was now describing half the rules — it covers both, with the reasoning for each.


hyiger:
@codex review

PR #1035, opened 2026-07-25
fix(nfc): bound the NTAG read page so a hostile CC can't splice pages 0-3 into the NDEF image (#1028)

hyiger:
Fixes #1028

A Type-2 READ BINARY pseudo-APDU carries the start page in a **single byte**:

```ts
Buffer.from([0xff, 0xb0, 0x00, startPage, 0x10])
```

`Buffer.from(array)` applies `& 0xff` **silently**, so page 256 emits `ffb0000010` — a read of pages 0–3 (UID, BCC, internal byte, static **LOCK** bytes, CC). `assembleNtagImage` derived its extent from the **tag-controlled** Capability Container byte and walked `page += 4` unbounded, so a CC byte ≥ `0x7f` drove the final burst to page 256 and copied that head back into the **tail** of the NDEF image — silent corruption, not an error.

`assembleNtagImage` feeds the read path, the OpenTag3D record check **and the post-write verify**, so a corrupted tail can also make a write verify reach the wrong conclusion.

## Threshold correction

The issue's original table said `>= 0x80`. **It's `>= 0x7f`** — `0x7f` → 1016 bytes → 1032-byte buffer → 64 bursts → final page 256. `0x7e` (1008 bytes, final page 252) is the last safe value. Caught by triage, verified independently, and the issue has been corrected.

No real NTAG21x reaches it — the largest legitimate `mlen` is `0x6d` (NTAG216, 872 bytes, final page 220) — so this needs a malformed / foreign / partially-written / hostile tag. Hence P3.

## Provenance

The write side has had exactly this guard since #927. `git log -L 793,821` shows commit `b1179fff` added `writeNtagPage`'s `[3,255]` bound **while extracting this very loop** — and never added the read-side counterpart.

## Four parts

**1. `readNtagBurst` refuses a page outside `[0, 255]`.** Page 0 *is* legal for a read (the head burst), unlike the write bound which starts at 3 because pages 0–2 hold the UID and static lock bytes.

**2. The assembly loop is bounded explicitly** (`page <= NTAG_MAX_PAGE`) rather than relying on that throw being caught and turned into a `break` — too subtle to survive future edits.

It also gains `if (copyLen === 0) break;`, and **this line is load-bearing**: a zero-length burst leaves `written` unchanged and spins forever, and the accidental wrap to page 0 was what kept data flowing and *masked* it. Bounding the page **without** the zero-progress guard converts an accidental save into a permanent NFC deadlock. For the same reason, do not "fix" this by clamping (`page = Math.min(page, 255)`) — clamping keeps the loop alive and is exactly the deadlock case.

**3. The ceiling was mis-derived.** `NTAG_MAX_NDEF_BYTES = 1024` was commented as *"a generous ceiling that bounds the read loop"* — it does not bound it. One Type-2 sector is 256 pages × 4 = 1024 bytes **total**, and the data area starts at page 4, so the largest addressable NDEF extent is (256−4) × 4 = **1008**. Being off by exactly the 16-byte head is precisely what let the loop run one burst past the addressable end.

> **I diverged from triage here.** It recommended `872` (`NTAG216_MAX_NDEF_BYTES`, aligning read with the write path's clamp). I chose **1008**: the page bounds in parts 1–2 are the structural guarantee, so the constant only controls permissiveness — and 1008 is a pure arithmetic correction with **zero behavioural narrowing**, whereas 872 would additionally truncate a hypothetical foreign/emulated Type-2 device claiming 873–1008 bytes. Easy to change to 872 if you'd rather have the alignment; the guards make it correct either way.

**4. `ndefBytesFromCc` tolerates a short head.** Two of the three callers check `head.length >= 16` first; the **write-verify caller does not**, so a truncated burst produced `Buffer.alloc(16 + NaN)` → a raw `RangeError`.

## Behavioural posture

Unchanged apart from the fix: an over-large CC now yields a **short image** (callers already use the returned `written`) instead of a wrapped read. Truncating silently matches the existing NAK-driven `break`; failing loudly could regress tags that work today.

## Extraction

The arithmetic moves to `src/lib/ntagPages.ts` — pure, and inside the **coverage-gated** tree — mirroring what #973 did for the GET_VERSION parse (`src/lib/ntagVersion.ts`). `electron/` has no test file at all and `nfc-service.ts` is its largest module; this is its *second* page-arithmetic safety bug, so the numbers belong somewhere they can be pinned.

## Tests

- **Exhaustive over all 256 CC values** — no burst page may exceed 255, for any byte a tag can present.
- Pins the boundary table, including the corrected `0x7f` threshold and every real chip (`0x12`/`0x3e`/`0x6d` unchanged at 9/31/55 bursts).
- Asserts no truncation for real chips (bursts × 16 ≥ declared extent).
- Hostile-head cases: short head, zero CC, `NaN`/negative/`Infinity`.
- Asserts **the defect mechanism itself**, so it can't be argued away:

```ts
expect(Buffer.from([0xff, 0xb0, 0x00, 256, 0x10]).toString("hex")).toBe("ffb0000010");
expect(Buffer.from([0xff, 0xb0, 0x00, 256, 0x10])[3]).toBe(0);
```

**Verified non-vacuous** — restoring the 1024 ceiling + unbounded walk fails 3 tests, including `expected 256 to be less than or equal to 255`.

| check | result |
|---|---|
| `npm run lint` | clean |
| `npm run electron:typecheck` | clean |
| `npm test` | **3419/3419**, 164 files |
| coverage thresholds | met — 99.55% lines / 99.06% statements / 97.74% functions / 97.57% branches |

## Still owed

**On-hardware confirmation.** No ACR1552U + NTAG in this environment. Mitigating: the change is arithmetic-only and **cannot affect a tag whose CC byte is ≤ 0x7e**, which covers every real NTAG213/215/216 — the burst sequence for those is byte-for-byte identical to before (pinned by the boundary table test).

### DIFF
--- a/electron/nfc-service.ts
+++ b/electron/nfc-service.ts
@@ -31,6 +31,14 @@ import {
   NTAG_NAME_TO_NDEF_BYTES,
   type NtagSizeName,
 } from "../src/lib/ntagVersion";
+import {
+  NTAG_CC_OFFSET,
+  NTAG_TLV_OFFSET,
+  NTAG_FIRST_DATA_PAGE,
+  NTAG_MAX_PAGE,
+  ndefBytesFromCc,
+  isReadablePage,
+} from "../src/lib/ntagPages";
 import { deriveBambuKeys, parseBambuBlocks, bambuToDecodedTag } from "./bambu-tag";
 import {
   classifyNfcError as classifyNfcErrorImpl,
@@ -88,12 +96,16 @@ const BLOCK_SIZE = 4;
 const DEFAULT_BLOCK_COUNT = 80;
 
 // NTAG (NFC-A / ISO 14443 Type 2) read tuning (#864). FF B0 READ BINARY returns
-// a 4-page (16-byte) burst; the Type-2 CC sits in page 3 (bytes 12–15) with the
-// NDEF TLV area starting at page 4 (byte 16). NTAG216 tops out near 872 user
-// bytes, so 1 KB is a generous ceiling that bounds the read loop.
-const NTAG_CC_OFFSET = 12;
-const NTAG_TLV_OFFSET = 16;
-const NTAG_MAX_NDEF_BYTES = 1024;
+// a 4-page (16-byte) burst; the Type-2 CC sits in page 3 (bytes 12-15) with the
+// NDEF TLV area starting at page 4 (byte 16).
+//
+// GH #1028: the offsets, the extent ceiling and the page bound now live in the
+// pure, unit-tested ../src/lib/ntagPages.ts. The old local ceiling of 1024 was
+// arithmetically unreachable — one Type-2 sector is 256 pages x 4 = 1024 bytes
+// TOTAL and the data area starts at page 4, so the real maximum is
+// (256 - 4) x 4 = 1008. Being off by the 16-byte head is exactly what let the
+// read loop run one burst PAST the addressable end and wrap the single-byte
+// page field back to 0.
 // Write-side extent caps (Codex #927) now live in the pure, unit-tested
 // resolveNtagWriteSize (src/lib/ntagVersion.ts): NTAG216_MAX_NDEF_BYTES (872,
 // the largest real NTAG) and NTAG213_NDEF_BYTES (144, safe on ANY chip).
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

PR #1038, opened 2026-07-27
feat(labels): TSPL emitter for the KNAON Y813BT dry-box label printer

hyiger:
Phase 1 of the TSPL label emitter: the pure emitter core. **No transport, no UI, no template layer** — those are phases 2–4, scoped at the bottom.

Adds support for a second label printer alongside the existing Brother PT-P710BT. The two do not overlap: **Brother / 24mm tape / spool labels** (unchanged), **KNAON / 4×6 sheet / dry-box labels** (new). So this is not a printer-kind abstraction over one shared surface — it is a separate emitter for a separate label domain.

## Task 0 is resolved

The spec's gating question was whether the firmware implements the text/barcode/2D primitives or only raster. The vendor CUPS filter only ever emits `BITMAP`, so it proves nothing either way, and many ODM firmwares ship a raster-only subset.

Printing `03_probe_full.prn` settled it: **all nine numbered items rendered**, including item 7 (`BARCODE`, Code 128) and item 8 (`QRCODE`).

That is the best-case branch. A 4×6 label is ~600 bytes rendered from the printer's font ROM at native resolution; the rasterized alternative is 812 × 1218 px 1-bit ≈ 124 KB through a threshold-and-dither path. **No QR library, no rasterization path, and no bitmap compositor is needed** — `bitmap` stays in the command union for future logo support and nothing more. The outcome is recorded at the top of the module, since it cost a printed label to establish.

## Design notes worth reviewing

**Returns `Uint8Array`, not `Buffer`.** The spec asked for `Buffer`, but justified `bitmap.data` being `Uint8Array` so the shared module stays free of Node types. The return value needs the same treatment — the module is importable by the sandboxed renderer to drive an on-screen preview from the same `LabelDocument`, so `node:buffer` must not enter the client bundle. Matches how `src/lib/labelEncoder.ts` already works.

**Four gaps in the spec's proposed type union**, all surfaced by trying to reproduce the fixtures byte-for-byte:

- `qrcode` needs `mode` (`'A'|'M'`) — the fixtures emit `QRCODE 24,560,H,6,A,0,"…"` with ecc and mode as **bare unquoted tokens**, unlike the quoted symbology on `BARCODE`
- `LabelSpec` needs optional `reference` and `direction`
- `PRINT` takes two arguments (`sets`, `copies`), not a single `copies`
- the `DENSITY`/`SPEED`/`REFERENCE`/`DIRECTION` block must be **structurally** optional, not merely value-optional — the minimal probe job that proves the parser is live is `SIZE`/`GAP`/`CLS`/`PRINT` and nothing else

**`DIRECTION` is emitted only when set.** It is absent from the vendor driver's string table, so it is unverified on this firmware and is the first line to drop when diagnosing a job that produces no output.

**Encoding folds to strict 7-bit ASCII.** The two candidate codepages disagree on exactly the characters this app produces: `°` is `0xF8` in cp437 but `0xB0` in latin1, and `³` has **no cp437 codepoint at all** (the reference generator used `cp437, errors="replace"`, so `mm³/s` silently became `mm?/s` there). All four fixtures are pure ASCII, so **they cannot validate any encoding choice**. Folding via NFD decomposition plus a transliteration table is deterministic, testable, needs no `CODEPAGE` command, and the table is a single swap point once a hardware probe settles it. U+202F is in the table specifically because it is `GROUP_SPACE`, the thousands separator the v1.66 "Space" number format emits — so it reaches a label through any formatted weight or price.

**A newline-injection bug the tests caught.** `sanitizeTsplLiteral` now strips line terminators *before* the ASCII fold. The other order is silently wrong: the fold maps control characters to `?`, leaving nothing for the strip to match, so a newline in a user-entered dry-box note would have terminated the `TEXT` command mid-literal and handed the remainder to the parser as a new command. Pinned by a test asserting an injected `\r\nPRINT 99,99` renders as five lines, not six.

## Golden fixtures

The four `.prn` files are byte-for-byte captures of jobs that printed correctly on the physical printer. Each is reconstructed as a `LabelDocument` and compared with byte equality — that is what pins the wire format.

There is no prior binary-fixture precedent in this repo (166 test files, zero non-`.ts`), so this also adds a `.gitattributes` with `*.prn -text -diff`. That rule is load-bearing: TSPL requires CRLF framing and an LF-only job is **silently discarded** by the firmware, so a line-ending-normalised fixture would still look correct in a diff while no longer matching what the printer accepts. Verified the committed blobs are byte-identical to the originals with CRLF intact.

The alternative — building fixtures in code, matching the house pattern — was considered and rejected: retyping hardware-verified ground truth reintroduces exactly the transcription risk a golden file exists to eliminate.

## Gates

- 43 tests, all passing
- **100%** statements / branches / functions / lines on the new module (thresholds are 99/98/96/96, global over `src/lib/**`)
- ESLint clean, `tsc --noEmit` clean
- Full suite green: **165 files, 3472 tests**

One thing reviewers may want to check: `\p{M}` was swapped for an explicit `[̀-ͯ]` range. Property escapes are ES2018 and this repo targets ES2017 — `tsc` validates regex *flags* against the target but not regex *bodies*, so `\p{M}` passed the typecheck and would only have surfaced at runtime.

## Follow-ups (not in this PR)

**Phase 2 — transport.** The spec recommended a `usb` native module. That looks wrong for this repo: GH #588 deliberately *removed* `serialport` from the label path in favour of shelling out to the OS print system, `printLabel(target, bytes)` is already a model-agnostic raw byte pump, and the in-repo comment on the CUPS path notes `-o raw` sends bytes "verbatim regardless of any driver attached to the queue" — so the vendor PPD cannot mangle TSPL. Reusing it avoids `@electron/rebuild`, macOS `LIBUSB_ERROR_ACCESS`, and the Windows Zadig problem entirely. (`asarUnpack` is moot regardless — `asar: false`.) The concrete blocker is that `label-printer-print` carries only `bytes` and reads one global `labelPrinterDevicePath` store key, so it cannot address two printers without a signature change across `electron/main.ts` + `preload.ts` + `src/types/electron.d.ts`.

**Phase 3 — `dryBoxLabel()` template.** Needs a new `Location.desiccantChangedAt` field; `Location` is currently name/syncId/kind/humidity/notes/`_deletedAt`. Contents manifest can come straight from `GET /api/spools/by-location`, with an "as of" date stamp since a printed manifest goes stale.

**Phase 4 — QR target.** §4.3 of the spec asks to align with the OpenPrintTag `storage_location` aux field, but **that field does not exist anywhere in this codebase** (the OPT aux region carries exactly one key, `AUX_CONSUMED_WEIGHT = 0`). So the dry-box QR is new design. Location `_id` is the only usable stable identifier — `syncId` is minted lazily by the hybrid-sync engine and stripped from client POSTs, so it is absent in most deployments. Suggested target is `<base>/inventory?location=<id>`, since `/inventory` already groups spools by location and answers "what is in this box" live; there is no location view page today, only `/locations/{id}/edit`.

**Also open:** a fifth probe fixture printing a codepage test card, to settle `°`/`³` on hardware rather than transliterating indefinitely.


hyiger:
@codex review

hyiger:
Thanks — all three reproduced before fixing, and all three were real. Fixed in 951418bc with regression tests for each.

**1. BITMAP payloads blew the call stack.** Confirmed: `RangeError` at exactly the 124,236-byte figure you cited (102 × 1218). `asciiBytes` now returns `Uint8Array` and `render()` accumulates chunks concatenated with `set()`, so nothing is spread. `renderCommand` returns `Uint8Array[]` and a bitmap's binary payload rides as its own chunk, never passing through the ASCII encoder.

**2. Valid rasters were rejected.** You're right that a byte-level scan can't distinguish a payload `0x0A` from a broken terminator. Rather than teach the scan about payload ranges, framing is now guaranteed **by construction** — every fragment is appended through a helper that follows it with CRLF, and `asciiBytes` rejects a raw terminator *inside* a command. That's strictly stronger and stays valid with binary data. `render()` asserts only the always-valid tail check; `assertCrlfFramed` keeps the full scan and stays exported for validating jobs from outside the emitter (a `.prn` read from disk), now documented as bitmap-free-only.

**3. Non-finite dimensions.** Confirmed both shapes — `SIZE Infinity mm,150 mm` and `GAP 3 mm,NaN mm` were emitted silently. All four dimensional fields now go through `Number.isFinite` before the range comparisons. One deliberate exception: a negative `gapOffsetMm` stays legal, since it's a signed alignment nudge unlike the gap itself.

Coverage remains 100% statements/branches/functions/lines; 47 tests.

@codex review

hyiger:
Fixed in e7a5b647 — you're right, and the contract was the part that was wrong. `A\rB\r\n` ends in CRLF and contains no unpaired LF, so an LF-only scan approves it while its first CR terminates nothing. Now checks both halves of the pair.

The reasoning that makes a bare CR unambiguously malformed rather than merely suspicious: in a bitmap-free job — the only kind this helper is documented as valid for — every byte is either ASCII command text or a terminator, because `asciiBytes` rejects CR and LF *inside* a command. So a CR not followed by LF has no legitimate reading. Added coverage for the bare-CR case, a trailing-CR case, a well-formed multi-line job, and all four golden fixtures pushed through the validator.

Note for anyone reading the thread: the three findings still showing at lines 565/591/633 are the **original** review re-anchored to shifted line numbers — `original_commit_id` on each is `d22f3864`, and all three were fixed in 951418bc. This CR finding is the only one raised against 951418bc.

50 tests, coverage still 100% statements/branches/functions/lines.

@codex review

hyiger:
@codex review

(Re-requesting: HEAD is now `e7a5b647`, and I don't see a review against it yet — the latest I can find is on `951418bc`.)

Issue #424, opened 2026-05-30
Mass-assignment risk: Nozzle/Printer/BedType/Location PUT pass body straight through


hyiger:
## Summary
Four entity PUT routes pass the entire request body to `findOneAndUpdate(..., body, ...)` with only `_id` / `_deletedAt` / `createdAt` / `updatedAt` / `__v` / `syncId` / `instanceId` stripped. There's no allowlist, so any future schema field is automatically client-writable. The Filament PUT pattern (which explicitly strips `_purged`, `_parent`, `_variants`, `_inherited`, `spools`) was a deliberate hardening that wasn't propagated here.

`_purged` doesn't exist on these models, so the GH #222 tombstone-hijack doesn't apply today — but the precedent is wrong and a future field added (e.g. an ownership / shared-from flag) silently becomes attacker-controlled.

## Evidence
- `src/app/api/locations/[id]/route.ts:49-53`
- `src/app/api/nozzles/[id]/route.ts:72-76`
- `src/app/api/printers/[id]/route.ts:106-110`
- `src/app/api/bed-types/[id]/route.ts:50-54`

## Suggested fix
Adopt the Filament-PUT pattern in each handler: extract a known allowlist of editable fields (`name`, `kind`, `humidity`, `notes`, etc. per model) instead of spreading the whole body. This also makes the API contract explicit.

Filed from full-codebase review on 2026-05-29.

PR #1042, opened 2026-07-29
feat(labels): TSPL transport for a second label printer + dry-box label template

hyiger:
Phases 2 and 3 of the TSPL dry-box label work, building on #1038. **Partial by design** — this is the transport, schema and template. The Settings picker, the `/inventory` print entry point and a bring-up CLI are follow-up commits on this branch; there is no end-to-end print path from the UI yet.

Makes the KNAON Y813BT an **independent second label printer** alongside the Brother PT-P710BT. They never print the same thing — Brother/24mm tape/spool labels, KNAON/4×6 stock/dry-box labels — so they get separate device settings rather than a shared "which kind" toggle.

## Phase 2 — transport

**No new native module.** The handoff spec recommended `usb` + `@electron/rebuild`, but GH #588 deliberately *removed* `serialport` from this path in favour of the OS print system, and `electron/label-printer.ts` is already a byte-agnostic raw pump (`lp -o raw` / winspool `RAW`). The in-repo comment notes `-o raw` sends bytes "verbatim regardless of any driver attached to the queue", so TSPL reaches the print head exactly as Brother raster does. This also sidesteps `LIBUSB_ERROR_ACCESS` on macOS and the Windows Zadig problem entirely. (`asarUnpack` was moot regardless — `asar: false`.)

### It fixes a wrong-printer race

`MANAGED_QUEUE = "FilamentDB_Label"` was a single module-global CUPS queue name, and `ensureManagedQueue` **rebinds** its queue to whichever device is printing. With two USB label printers both selected as raw `usb://` URIs, a Brother print and a dry-box print overlapping in time would rebind the queue away from each other — and because CUPS delivery is asynchronous (the `lp` job is spooled after the rebind but drains later), a job could be delivered to the **wrong physical printer**.

Managed queues are now per-kind, with distinct `lpadmin -D` descriptions so they aren't duplicates in the user's system printer list. **The Brother queue keeps its original name**, so existing installs are untouched by the upgrade.

### One handler table, not two copies

The three label-printer IPC handlers are registered from one table for both printers. These handlers are mirrored by hand across `electron/main.ts`, `electron/preload.ts` and `src/types/electron.d.ts` with **no compile-time link** between them, and that mirroring has already drifted in production (GH #1006 F4, where preload's local `NfcStatus` silently omitted a field). Two hand-copied ~50-line print handlers would be the same trap: a validation guard tightened on one and missed on the other is a security fix that only half-lands.

`tests/ipc-contract-parity.test.ts` is a new permanent guard on that contract — every channel preload invokes must be registered main-side, every contextBridge method must be declared on `ElectronAPI`, and both printers must expose a symmetric channel set. **Both drift classes were verified to fail it before being restored.** Worth noting the first version of that test was itself broken: it passed on a deliberately typo'd channel because the literal scan included `preload.ts` and matched the typo against itself.

Device listing is deliberately **not** duplicated — `listLabelPrinters` is printer-agnostic and both pickers share it.

## Phase 3 — schema + template

`Location.desiccantChangedAt` (optional date) answers the one question a glance at a drybox can't. Plumbed through the PUT allowlist (GH #424), POST, the `by-location` `$project`, snapshot `DATE_FIELDS` so restore revives it as a `Date` rather than an ISO string, the shared `InventoryLocation` type, and the location form as a date input.

Both write paths validate with `isValidIsoDateString` **before** the cast: Mongoose *rolls over* an impossible-but-ISO-shaped date rather than rejecting it, so `"2026-02-30"` would silently persist as March 2nd — the same trap the spool date fields hit in GH #372. There's a test asserting the stored value is untouched after a rejected write.

`src/lib/dryBoxLabel.ts` generalizes the hardware-verified `04_drybox_label.prn` geometry: the header block, contents rule, desiccant line and Code 128 keep their proven positions, while the contents list reflows to fill the sheet (16 rows on 100×150mm, vs the three the fixture hardcoded) with the footer pinned to the bottom of whatever stock is configured.

Three choices worth a reviewer's attention:

- **Overflow is reported, not truncated.** A label showing 16 of 40 spools with no indication reads as a complete inventory; past capacity the last row becomes `+N more`.
- **The manifest is stamped "as of \<date\>".** A printed contents list is a point-in-time snapshot and spools move. The app already made this call for Brother spool labels (no remaining-grams field, because a printed amount goes stale) — but for a dry box the manifest *is* the point of the label, so it prints, dated, and the QR resolves the live view.
- **`asOf`, `formatDate` and all strings are injected**, not read from the clock or locale. Calling `toLocaleDateString` inside the template would make the same document render differently on the preview canvas and the print head; reading the clock would break determinism.

## Gates

- `src/lib/dryBoxLabel.ts` at **100%** statements / branches / functions / lines, 23 tests — including the invariant that no manifest row or the barcode can land below the footer at any stock height
- ESLint, `tsc --noEmit`, and `electron:typecheck` all clean
- en/de both at 1681 keys, i18n invariant green

## Deliberately inert

`dryBoxLabel()` has no caller yet, and the `tspl-printer-*` channels have no UI. That's the same shape as #1038 — the risky, testable parts land reviewable on their own. Nothing is half-wired: there's no dead path a user can reach today.

## Follow-ups on this branch

Settings picker · `scripts/print-tspl.ts` bring-up CLI · `/inventory` group-header print entry point.

Still open: a **codepage probe** on hardware to settle `°`/`³` (the emitter currently folds to strict ASCII — deterministic and safe, but `mm³/s` renders as `mm3/s`). The probe job is written and only needs `send_tspl.py`.

### DIFF
--- a/electron/label-printer.ts
+++ b/electron/label-printer.ts
@@ -30,10 +30,17 @@ import { join, win32 } from "node:path";
 
 const execFileP = promisify(execFile);
 
-/** Heuristic match for "this device is a label printer we know about" —
- *  the PT-series (Brother raster) or the Y813BT (KNAON, TSPL). Drives a
- *  badge in the picker only; it never gates printing. */
-const PRINTER_PATTERN = /pt-?p710bt|p-?touch|brother|knaon|y-?813/i;
+/**
+ * Heuristic match for "this device is a Brother PT-series label printer".
+ *
+ * Deliberately NOT widened to cover the KNAON. The only consumer today is the
+ * Brother picker in LabelPrinterSettings, which renders this flag as a
+ * "PT-Touch" badge and saves the selection for labelPrinterPrint — whose test
+ * job is Brother raster. Badging a Y813BT there would actively invite the user
+ * to select it and print raster bytes to a TSPL printer. Per-kind badging
+ * belongs with the TSPL picker, alongside a pattern of its own.
+ */
+const PRINTER_PATTERN = /pt-?p710bt|p-?touch|brother/i;
 
 /**
  * Which label printer a job is bound for.
--- a/electron/label-printer.ts
+++ b/electron/label-printer.ts
@@ -30,13 +30,61 @@ import { join, win32 } from "node:path";
 
 const execFileP = promisify(execFile);
 
-/** Heuristic match for "this printer/device is a PT-series label printer". */
-const PRINTER_PATTERN = /pt-?p710bt|p-?touch|brother/i;
+/** Heuristic match for "this device is a label printer we know about" —
+ *  the PT-series (Brother raster) or the Y813BT (KNAON, TSPL). Drives a
+ *  badge in the picker only; it never gates printing. */
+const PRINTER_PATTERN = /pt-?p710bt|p-?touch|brother|knaon|y-?813/i;
+
+/**
+ * Which label printer a job is bound for.
+ *
+ * The app drives two physically different printers that never overlap:
+ * "brother" is the PT-P710BT on 24mm tape (spool labels, raster bytes) and
+ * "tspl" is the KNAON Y813BT on 4x6 stock (dry-box labels, TSPL bytes).
+ * The transport itself is byte-agnostic — this only selects which managed
+ * queue and description a raw `usb://` device gets bound to.
+ */
+export type LabelPrinterKind = "brother" | "tspl";
 
 /** Name of the hidden raw CUPS queue we manage for raw `usb://` devices
  *  that aren't already installed as a queue. CUPS queue names allow only
  *  letters, digits and underscores. */
 const MANAGED_QUEUE = "FilamentDB_Label";
+
+/**
+ * The TSPL printer gets its OWN managed queue, and that is load-bearing
+ * rather than tidiness.
+ *
+ * `ensureManagedQueue` REBINDS its queue to whichever device is currently
+ * printing. With both printers sharing one queue name, a Brother print and
+ * a dry-box print overlapping in time would rebind the queue away from each
+ * other — and because CUPS delivery is asynchronous (the `lp` job is spooled
+ * after the rebind but drains later) a job could be delivered to the WRONG
+ * PHYSICAL PRINTER. Per-kind queues make that impossible.
+ *
+ * The Brother queue keeps its original name so existing installs, which
+ * already have `FilamentDB_Label` bound, are untouched by the upgrade.
+ */
+const MANAGED_QUEUE_TSPL = `${MANAGED_QUEUE}_TSPL`;
+
+/** Every queue this app creates and owns. */
+const MANAGED_QUEUES: readonly string[] = [MANAGED_QUEUE, MANAGED_QUEUE_TSPL];
+
+/** True when `name` is one of OUR managed queues — an implementation detail
+ *  the picker surfaces as the underlying USB device rather than as a queue. */
+function isManagedQueueName(name: string): boolean {
+  return MANAGED_QUEUES.includes(name);
+}
+
+function managedQueueFor(kind: LabelPrinterKind): string {
+  return kind === "tspl" ? MANAGED_QUEUE_TSPL : MANAGED_QUEUE;
+}
+
+/** `lpadmin -D` description, so the two queues are distinguishable in the
+ *  user's system printer list rather than appearing as duplicates. */
+function queueDescriptionFor(kind: LabelPrinterKind): string {
+  return kind === "tspl" ? "Filament DB Dry-Box Label Printer" : "Filament DB Label Printer";
+}
 
 /** Per-subprocess timeout for listing + the CUPS `lp` print. Listing and a
  *  single 24mm label both complete well under this; the IPC handler wraps the
@@ -156,7 +204,7 @@ async function listCupsPrinters(probeUsb: boolean): Promise<LabelPrinterDevice[]
       const name = m[1].trim();
       const uri = m[2].trim();
       seenUris.add(uri);
-      if (name === MANAGED_QUEUE) {
+      if (isManagedQueueName(name)) {
         // Our own managed queue is an implementation detail — surface it as
         // the underlying USB *device* (path = the uri), so selecting it and
         // printing both route through ensureManagedQueue idempotently. Without
@@ -437,19 +485,33 @@ function isLegacySerialTarget(target: string): boolean {
  * Send the raster byte stream to the selected print target. Rejects with a
  * descriptive Error on failure; the IPC handler surfaces it to the renderer.
  */
-export async function printLabel(target: string, bytes: Uint8Array): Promise<void> {
+export async function printLabel(
+  target: string,
+  bytes: Uint8Array,
+  kind: LabelPrinterKind = "brother",
+): Promise<void> {
   if (isLegacySerialTarget(target)) {
+    // Only the Brother setting can hold a legacy serial value — the TSPL
+    // setting postdates the Bluetooth transport entirely — so the Brother
+    // copy stays, and the TSPL case gets wording that isn't a lie.
     throw new Error(
-      `"${target}" is a serial-port setting from an older version that printed over Bluetooth. ` +
-        `The PT-P710BT now prints over USB — open Settings → Label Printer and select your printer again.`,
+      kind === "tspl"
+        ? `"${target}" is a serial-port setting, which is not a valid print target. ` +
+            `Open Settings → Devices and select your label printer again.`
+        : `"${target}" is a serial-port setting from an older version that printed over Bluetooth. ` +
+            `The PT-P710BT now prints over USB — open Settings → Label Printer and select your printer again.`,
     );
   }
   if (process.platform === "win32") return printWindows(target, bytes);
-  if (isCups()) return printCups(target, bytes);
+  if (isCups()) return printCups(target, bytes, kind);
   throw new Error(`Label printing is not supported on platform "${process.platform}".`);
 }
 
-async function printCups(target: string, bytes: Uint8Array): Promise<void> {
+async function printCups(
+  target: string,
+  bytes: Uint8Array,
+  kind: LabelPrinterKind,
+): Promise<void> {
   // A `usb://…` target is a raw device → route it through our managed raw
   // queue. Otherwise it's an installed queue name. Only the usb scheme is
   // accepted — it's the only scheme listLabelPrinters ever surfaces — so a
@@ -457,8 +519,8 @@ async function printCups(target: string, bytes: Uint8Array): Promise<void> {
   // instead of being forwarded verbatim to `lpadmin -v` (GH #623).
   let queue = target;
   if (/^usb:\/\//i.test(target)) {
-    await ensureManagedQueue(target);
-    queue = MANAGED_QUEUE;
+    queue = managedQueueFor(kind);
+    await ensureManagedQueue(target, kind);
   } else if (/^[a-z][a-z0-9+.-]*:\/\//i.test(target)) {
     throw new Error(
       `Unsupported print target "${target}" — only usb:// devices and installed ` +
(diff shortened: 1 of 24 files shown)


hyiger:
@codex review

hyiger:
All three confirmed and fixed in c8b7fb69 — the third one I'd have gotten wrong on my own.

**1. QR ran off the sheet.** Verified the arithmetic: a `QRCODE` command's printed size isn't implied by its arguments, since the firmware picks the version from the payload at print time. The fixture's own 16-byte payload is 29 modules = 203 dots and fits; a realistic 67-byte deep link is 49 modules = 343 dots, which at the inherited `cell 7` runs **104 dots past** the edge of 100mm stock — silently unprinted. Worth noting the fixture's QR also already spilled 63 dots *below* its header box, which this fixes too.

Added `qrModuleCount()` to the encoder (byte-mode capacity tables, all four ECC levels, v1–v20) and `fitQr()` to the template: it picks the **strongest ECC that still prints at a scannable module size** and steps down only when geometry forces it — the trade-off the handoff spec asks for past ~60 characters — and throws rather than emit a clipped symbol. The layout below the header is now derived from the fitted QR, since the header box has to grow to contain it.

**2. Manifest rows ran off the sheet.** Confirmed: font 3 is 16 dots wide, so ~45 characters fit at x=56 on 100mm stock, and spool labels are free-form. Rows now truncate with an explicit `...`.

**3. Snapshot version.** I initially thought this was refutable — an additive, nullable, optional field seemed like it should degrade gracefully rather than make older builds reject the entire file. The repo's own version history decided it against me: v5 was bumped *"so pre-#1022 builds — which would drop the flag — reject the file via the #953 guard"*, and v4 exists because share links were silently dropped on round-trip. A refused restore is recoverable; a "restored successfully" that quietly lost data isn't. Bumped to 6 with a history entry.

While fixing 3 I found the #953 guard test hardcoded `6` as "a newer version" — this bump would have silently turned that into the *current* version and disabled the guard's own coverage. It now uses a far-future number so the next bump can't repeat it.

Coverage still 100% statements/branches/functions/lines on both `tsplEncoder.ts` (54 tests) and `dryBoxLabel.ts` (32 tests).

@codex review

hyiger:
Pushed 5d7deb29 — three more, from an adversarial self-review of the diff I ran in parallel with your round. Each confirmed by executing the committed code before fixing.

**1. `describeItem` silently dropped the material type.** The dedup guard was a bare substring test over `"<vendor> <name>"`. Verified by execution: `"Prusa Space Gray" + PA` → `"Prusa Space Gray"` — the `pa` inside `Space` swallowed it. Same for Papaya/PA, Sapphire/PP, Petrol Blue/PET. Those are real material types and ordinary colour names, and on a dry-box manifest the material is the one thing a glance needs. Now matches whole tokens by normalising separators and space-padding both sides, which also makes hyphenated `PC-ABS` match a `PC ABS` name.

**2. The header name overprinted the QR.** The previous commit truncated manifest rows but not the two header fields, which share their row with the symbol. Name is font 5 (32 dots) at x=40, QR column starts at x=560 → 16 characters fit. `"Garage Cabinet Drybox #2"` ran 208 dots into the symbol, printing a label whose QR doesn't scan. Both header fields now have their own budget ending at the QR column.

**3. A row was emitted even when none fit.** On stock too short for one manifest row (capacity 0, e.g. 65mm) the builder still emitted one at `firstRowY`, landing on the footer rule and barcode. Now emits no manifest at all.

**And one of my own tests could not fail.** The row-overflow guard built its row set with `.filter(y => y < footerTop)` and then asserted `expect(y).toBeLessThan(footerTop)` — any offending row was removed *before* the assertion, so it passed for every possible implementation, while its comment claimed it guarded exactly the bug in (3), which was live and reachable. Rows are now selected structurally (`x === rows.x` is unique to manifest rows) and the loop covers capacity-0 heights. Verified by mutation: inflating capacity by 3 now fails it.

Coverage still 100% statements/branches/functions/lines, 39 tests.

@codex review

hyiger:
All three reproduced and fixed in 939001f6. Two of them are cases where my previous fix solved half the problem — worth calling out.

**1. The contents heading and rule still printed over the footer.** I suppressed manifest *rows* on capacity-0 stock but left the heading and its rule emitted unconditionally, so at 65mm a `CONTENTS` heading lands at y=303 against a footer starting at 279. The block is now emitted as a unit — heading, rule and rows — or not at all, with a test asserting nothing lands in the footer band.

**2. Truncation measured the wrong string.** This is the one I'm glad you caught. `fitRowText` budgeted against its *raw* input, but the encoder folds to 7-bit ASCII on the way out and several substitutions are length-**increasing**: `ß`→`ss`, `°`→`deg`, `€`→`EUR`. Verified: sixteen `ß` fitted to a 16-char budget print as **thirty-two**. Entirely invisible to ASCII testing — it would have first surfaced on a German user's label, which is exactly the population the transliteration table exists for. `fitRowText` now measures the sanitized string; sanitisation is idempotent so the encoder's fold is a no-op afterwards.

**3. `PLA+` compared equal to `PLA`.** The whole-token normalisation I added last round collapsed *every* non-alphanumeric to a separator, which erased a real material distinction — a PLA+ spool whose name mentions PLA printed as plain `PLA`, understating the material. `+` now survives normalisation while hyphens and slashes still collapse, preserving the `PC-ABS` ↔ `PC ABS` match that motivated the change.

Coverage still 100% statements/branches/functions/lines, now 44 tests.

@codex review

hyiger:
All three fixed in 41320438.

**1. The QR was sized from the wrong string — the same defect as the manifest rows two commits ago, in a sibling path I didn't check.** `fitQr` measured `input.qrPayload`, but `renderCommand` sanitizes on the way out and several substitutions grow the string. Verified: twelve `£` sized a 203-dot symbol and printed a 259-dot one, 56 dots past its region.

Fixed **structurally** rather than at the call site — every payload is now sanitized once at the top of `dryBoxLabel` and the sanitized string is used for both sizing and emission, so a third measurement site can't repeat it. Sanitisation is idempotent, so the encoder's fold is a no-op afterwards. I should have done this when the row-truncation instance came up.

**2. The default barcode had no width bound.** Location names have no schema or form length limit, and Code 128 width scales with the payload: 40 characters at 2-dot bars is 950 dots against 751 available. A clipped Code 128 has no stop pattern, so it scans as *nothing* while still looking like a barcode. `fitBarcode` now tries 2-dot bars, falls back to 1, and returns null when even that overflows — the barcode is then **omitted** rather than printed broken. The QR carries the identity regardless.

**3. `PRINTER_PATTERN` no longer matches the KNAON.** Widening it in the first commit was my mistake. Its only consumer today is the Brother picker, which renders the flag as a "PT-Touch" badge and saves the selection for `labelPrinterPrint` — Brother raster. Badging a Y813BT there actively invited the user to send raster bytes to a TSPL printer. Reverted to Brother-only; per-kind badging belongs with the TSPL picker and its own pattern.

Coverage still 100% statements/branches/functions/lines, 55 tests.

@codex review

hyiger:
All four fixed in 92fdfe69. Three shared a root cause — geometry that ignored the spec it was handed.

**1. `fitQr` returned symbols below its own readability floor.** `QR_MIN_CELL` is documented as the point where a 203 dpi thermal print stops scanning, but the fallback still returned 2- or 3-dot candidates "so something prints" — handing the user a label indistinguishable from a working one until they try to scan it. Now refuses: a payload that can't reach the floor at any ECC level is too long for the label.

**2. The desiccant date shifted a day west of UTC.** Reproduced on this host: stored as midnight UTC by the date input, a locale formatter renders the *previous* calendar day. This repo already hit exactly this in v1.60.1's analytics day labels. The date is re-anchored to local midnight before formatting, and the default formatter now reads local components so it agrees with an injected one rather than disagreeing by a day.

**3. Short stock emitted negative coordinates** — 20mm gives `footerTop = -80`. Now refused with a message naming the stock.

**4. Decorations ignored the stock width** — fixed `x1=796` header box and 740-dot rules overflow 90mm (719 dots). Both derive from `widthDots`.

Fixing (3) surfaced a case the finding didn't name: the QR was sized from **width alone**, so wide-but-short stock (120×70mm) grew a 287-dot symbol that pushed the header past the footer and turned a printable label into an error. The QR is now bounded on both axes and shrinks to fit.

Coverage back to 100% statements/branches/functions/lines, 55 tests — including the footer guard's narrow reachable window (short payload on 50mm stock, where the QR shrinks fine but the minimum header still overruns the footer).

One observation for calibration: rounds 4 and 5 have been almost entirely **non-default stock** (90mm, 70mm, 20mm, 120mm). The hardware here runs 4×6 only, so these are defensive-path hardening rather than reachable risk. Still worth fixing — the overrides are a public API — but I'd weight further findings in that area lower than anything touching the default path.

@codex review

hyiger:
Fixed in f2e2a603 — and this one had precedent in the repo I failed to follow: `src/lib/labelBitmap.ts` already renders its QR with `margin: QR_QUIET_ZONE_MODULES` for exactly this reason.

Confirmed the intrusion at default stock: a 5-dot cell needs 20 clear dots per side, and the header border sat **8** below the last data module. Same failure class as the clipped symbol three commits ago — prints, looks correct, never scans.

Fixed by making the quiet zone part of what has to **fit** rather than something left over. `fitQr` now sizes `(modules + 8) * cell` against the region and returns `quietDots`/`footprintDots` alongside `sizeDots`; the header box clears the footprint; and the symbol is emitted inset by its quiet zone so the reserved space exists on the leading edges too. All four sides are asserted clear across every supported stock size.

The numbers move — the default 100×150mm label drops from a 5-dot to a 4-dot cell, and a payload that fit 160 dots now needs 200 — but the ECC step-down behaviour is unchanged and every fit stays at or above the 4-dot readability floor.

Coverage 100% statements/branches/functions/lines, 59 tests.

@codex review

hyiger:
Fixed in 49b725c8 — and this is the same defect as the previous commit, one symbology over. Reserving the QR's quiet zone while leaving the barcode's unreserved is the third time on this branch I've fixed one instance and not checked its sibling.

Verified: a 27-character payload on 90mm stock fits on bars alone (664 of 671 dots) but needs 704 with its quiet zones, so 2-dot bars were accepted and the clear space ran off the sheet. `fitBarcode` now fits bars **plus both quiet zones**, and the symbol is emitted inset by its own quiet zone so the clear space lands on the sheet rather than off the left edge. Both edges asserted clear across every stock width and name length.

Coverage 100%, 62 tests, full lint/tsc/electron-typecheck clean.

---

Calibration note, since this is round 7. Five of the last six findings have been **symbol-clearance geometry** — clipped QR, wrong-sized QR, sub-readable modules, QR quiet zone, Code 128 quiet zone — all with the same signature: prints, looks right, never scans. That class is genuinely hard to catch by reading and trivial to catch with a scanner, and the hardware here runs one stock size (4×6) while several recent findings only reachable on 90/70/20mm overrides.

I'm going to stop iterating here and wire the print path so a real label can go through a real scanner. Further review findings are welcome, but I'd weight anything on the default 4×6 path far above the override paths.

@codex review

hyiger:
Codepage probe result committed in 801156e7 (docs + the probe fixture; no behaviour change).

All Codex threads resolved — every finding across the seven rounds is fixed and covered.

@codex review

hyiger:
Fixed in 603f1a53 — and this is the **fourth** time on this branch I've fixed one instance of a flaw and left its sibling. The type dedup moved to whole-token matching three commits ago; the vendor check sitting immediately next to it kept a raw `startsWith`.

Verified by execution: vendor `Sun` vanished from `Sunset Orange`, `Pol` from `Polar White`. Both are plausible brand/colour pairs, and a dropped brand on a dry-box manifest is exactly the silent loss the type fix existed to prevent.

The vendor now has to match the name's leading **whole tokens**, reusing the `tokenized` helper from the type check.

One deliberate trade-off worth flagging: `Prusa` + `Prusament PLA` now renders `Prusa Prusament PLA` rather than collapsing, since `prusa` is only a partial token of `prusament`. A redundant word is cosmetic; a dropped vendor loses information. The case the guard was documented for — `Prusament` + `Prusament PLA` — still collapses, as do multi-word vendors.

Threads resolved. Coverage 100%, 65 tests, full gate clean.

@codex review

PR #1043, opened 2026-07-29
feat(labels): dry-box label printing end to end — /inventory entry point, Settings picker, bring-up CLI

hyiger:
The completing slice of the TSPL dry-box feature. Everything merged in #1038/#1042 was deliberately inert; this makes it reachable — **a dry-box label can now be printed from the app.**

## `/inventory` entry point

Location group headers gain a print button, gated on `kind === "drybox"` and on location grouping. The full-width collapse `<button>` becomes a flex row so the print action is a **sibling, not a nested button**. `PrintDryBoxLabelDialog` mirrors `PrintLabelDialog` (focus trap, escape, lazy device/public-URL loads, toast errors) but is data-only glue: the document comes from the pure `dryBoxLabel()` template, fed by the group the page already holds — no second fetch. Retired spools are excluded from the manifest.

## Preview from the same document

`TsplLabelPreview` maps the `LabelDocument` to SVG — **one layout pass, two painters**, so preview and print cannot drift. Geometry is exact; `TEXT` uses `textLength` to match TSPL's advance widths so truncation is exact; the QR is a real symbol (`qrcode` package, same ECC); `BARCODE` is width-correct stylized stripes. Web mode downloads the `.prn` — a working path, not an inspection aid: `lp -o raw -d <queue> file.prn` prints it unchanged (hardware-verified through the vendor PPD).

## QR target

`buildLocationDeepLink()` → `/inventory?location=<id>`. Deliberately not a location page (none exists — only `/locations/{id}/edit`); `/inventory` already answers what a scanned box label asks, live. The page reads the param once (the detail page's `?spool=` pattern — no `useSearchParams`, no Suspense boundary), expands the group, scrolls to it, rings it briefly. With only a loopback origin the dialog warns but still prints — there's no alternate payload mode, the name + barcode still identify the box, and reprinting after setting a public URL is cheap.

## Settings picker

`TsplPrinterSettings` is the Brother panel's sibling: same state machine, same GH #771 passive-list-vs-probe posture, its own store key, its own KNAON/Y813 badge pattern (client-side — the transport's `PRINTER_PATTERN` stays Brother-only per #1042 round 4). Test print sends a minimal known-good job through the real pipeline. The device-row list is **extracted to a shared `PrinterDevicePicker`** rather than duplicated — a row fix landing in one picker but not the other is exactly the hand-mirroring drift this feature was already bitten by (GH #1006 F4). The Brother BiDi warning rides a `renderRowExtra` prop.

## Bring-up CLI

`scripts/print-tspl.ts` (`npm run label:tspl`): `--file` sends a raw `.prn` (the fixture ladder 01–05), `--demo` renders a sample dry-box label through the real pipeline; sink is a file or `--printer` via the app's own transport with kind `"tspl"`. Raw jobs are CRLF-validated first — an LF-only job is silently ignored by the firmware and presents exactly like a dead printer.

## Verified in the running app against real data

Created a drybox → moved a spool in → the printed-QR deep link scrolled to and ringed the group → dialog previewed correctly (real QR, name truncated at the QR column with a visible `...`, `12% RH` subtitle, the user's date format, and the manifest row `- 3D-Fuel CHCKX PCTG HF0.4` showing whole-token type dedup on live data) → both writes reverted, DB left byte-identical.

38 new i18n keys, en+de lockstep (1718 each). Gates: full lint / `tsc` / `electron:typecheck` clean; `labelDeepLink` at 100% coverage; 183 tests across the feature's suites.

Closes out the TSPL handoff spec: emitter (#1038), transport + schema + template (#1042), UI + CLI (this PR).


hyiger:
@codex review

hyiger:
Both fixed in c8e8df44.

**1. The manifest came from the filtered view.** Confirmed — and it's the exact failure class the template's `+N more` overflow row exists to prevent, reintroduced one layer up: I chose the page's in-hand group rows over a second fetch, and those rows are the *visible* subset under type/vendor filters and search. The dialog now fetches its own unfiltered manifest from `/api/spools/by-location` on open (which is what my original recon suggested before I optimized the fetch away). The page passes only the location; retired spools stay excluded server-side.

**2. The deep link depended on the scanner's `groupBy` preference.** Confirmed — under a persisted type/vendor/none grouping the section ids are bucket values, so the QR target didn't exist. Scanning a box label is an explicit request for the location view, so the handler now forces location grouping; the scroll moved to a second effect that runs after the re-grouped DOM commits (a same-tick rAF can fire before the setGroupBy commit and find no element).

Threads resolved. Gates clean, en/de at 1719 each.

@codex review

hyiger:
The first physical label caught one more — pushed in 4b5ca8d8.

The printed footer read `Replace every 90 days or when indicator tu...` — truncated 16 characters early. The hint prints in **font 2** (12 dots/char, 62 fit) but was budgeted with the manifest's **font-3** row width (45). The truncation guard worked exactly as designed; it was measuring the wrong font. `dryBoxGeometry` now carries per-font footer budgets and each line fits with its own.

Worth noting for the thread: eight review rounds of static analysis didn't surface this, and the label did instantly — it's a *visual* defect with correct-by-construction geometry underneath. Everything else on the photo checked out: registration, orientation, density, QR + quiet zone, manifest composition, the desiccant date (no timezone shift), and the inset Code 128.

@codex review

hyiger:
Both fixed in aa411e1a. Good lens shift, worth naming: now that physical labels exist, both findings are about what happens *around* one — a label whose box has since been emptied, and a race whose loser is a reprint.

**1. Public-URL race.** `null` conflated "IPC pending" with "not configured", so a fast manifest fetch could enable Print against `window.location.origin` — localhost in packaged Electron — for a user who *had* configured a public URL. Now three-state (`undefined`/`null`/string); the document waits for both async inputs, and the web path resolves immediately.

**2. Emptied-box scan.** The by-location aggregation builds groups from spools, so an emptied (or deleted) location has no group and the deep link silently no-opped. A printed QR pointing at silence reads as "the app is broken" — it now toasts what actually happened, naming both causes.

Threads resolved. Gates clean, en/de at 1720 each.

@codex review

hyiger:
(Amended aa411e1a → bb9e7179 before review started — same content plus an exhaustive-deps lint fix on the new effect. bb9e7179 is the HEAD to review.)

@codex review

hyiger:
Fixed in b44b38bc — and nicely spotted: it's the mirror of your previous finding. The scan side of an emptied box got a toast last round; the PRINT side had no entry point at all, because /inventory's groups are built from spools and its header was the dialog's only opener. A brand-new empty drybox — the exact moment you want to label one — was unprintable.

Drybox rows on **/locations** (which lists every location regardless of contents) now carry a "Print label" action opening the same dialog. No dialog changes: it already fetches its own manifest, and an empty box prints `(empty)` — the capability existed, only unreachable.

Threads resolved. Gates clean, en/de at 1721 each.

@codex review

hyiger:
@codex review

hyiger:
Fixed in 5d28d2ad — the `join` separator was a raw 0x00 byte instead of the escape sequence `"\0"`. Same runtime string, but git classified the whole file as binary, which would have cost line diffs, blame and textual merges on every future edit. Verified the diff is textual again (`git diff --numstat` shows line counts, not `- -`).

That's the third invisible-byte defect on this feature (U+00A0 and U+2028 hit the encoder during development), so the fix came with a control-byte sweep of every file this branch touched — this was the only one.

@codex review

hyiger:
Fixed in dbc030c9. Right call — transliteration is the correct transform for display text and exactly the wrong one for a URL: folding `münchen` to `munchen` yields a QR that scans fine and points at a domain the user doesn't own.

`buildLocationDeepLink` now returns the WHATWG ASCII serialization (`new URL(...).href`) — punycoded host, percent-encoded path, the *same* URL in pure ASCII — so the emitter's mandatory fold becomes a no-op instead of a rewrite. Unparseable bases fall back to the raw construction. Tests pin the punycode host, percent-encoded path, the pure-ASCII output property, no-double-encoding, and the fallback; `labelDeepLink` stays at 100% coverage.

@codex review

hyiger:
Fixed in 7d5452f3 — in **both** panels. This feature has been bitten four times by fixing one instance and leaving its sibling, so the Brother panel got checked first this time: it *had* a clear button, but only inside the non-empty device list — partial coverage of exactly the gap you found (all devices gone → picker unrendered → selection invisible and unclearable, prints going to a dead target).

Both panels' ready state now shows an amber banner whenever the persisted selection matches no listed device — including the empty-list branch, which is why this lives in the panels rather than the shared picker (it isn't rendered in that case). Names the stale path, one-click clear; Brother reuses its existing `handleClear` with its proper toast.

Threads resolved. Gates clean, en/de at 1723 each.

@codex review

PR #1045, opened 2026-07-29
fix(slicer): disambiguate same-diameter HF/standard nozzles in the exported condition

hyiger:
Fixes #1040.

The per-nozzle fan-out groups and names by diameter+type+highFlow, but the baked `compatible_printers_condition` read only the diameter — so a 0.4-std + 0.4-HF pair exported two presets with identical conditions, each visible on both printers.

**Mechanism verified against the fork source**: `nozzle_high_flow` is a per-extruder printer option, present in `s_Preset_printer_options`, and Prusa's own MK4S system profiles use exactly `nozzle_diameter[0]==0.4 and nozzle_high_flow[0]` / `and ! nozzle_high_flow[0]` (spaced `! `, which this follows). Old slicers fail open — today's visibility.

**Emitted only when ambiguous** (product call): the HF term appears only when a diameter exists in both highFlow states within the fan-out group. Unambiguous exports are byte-identical to today, so nothing currently visible disappears (the #1021 lesson).

**The real work is round-trip persistence** — both ingestion paths now recognize the new machine shape so it can never freeze into the shared settings bag as a fake user pin (which the export gate would then stamp onto every sibling, hiding the standard preset outright):
- bulk-import collapse regex widened for the HF-tailed shapes
- the fork sync-back strips a condition shape-matching the bake for the *hinted* nozzle, via the same module that emits it — this also fixes a latent pre-existing leak where a tick-less filament persisted the plain diameter condition into the shared bag on every sync

User pins survive everywhere: compound conditions, hand-typed unspaced `!`, and machine shapes for a different diameter than the hint.

Reproduced per the issue's snapshot scenario in tests: the std/HF pair now exports
```ini
compatible_printers_condition = nozzle_diameter[0]==0.4 and ! nozzle_high_flow[0]
compatible_printers_condition = nozzle_diameter[0]==0.4 and nozzle_high_flow[0]
```

### DIFF
--- a/src/app/api/filaments/[id]/route.ts
+++ b/src/app/api/filaments/[id]/route.ts
@@ -9,7 +9,7 @@ import { errorResponse, errorResponseFromCaught, handleDuplicateKeyError, isDupl
 import { assertSameOriginRequest, assertSafeUpdateBody } from "@/lib/requestGuard";
 import { mergeSlicerSettings } from "@/lib/slicerSettings";
 import { stripLegacyMachineCondition } from "@/lib/stripLegacyNozzleCondition";
-import { resolveSyncBackColor } from "@/lib/prusaSlicerBundle";
+import { resolveSyncBackColor, isMachineDerivedPerNozzleCondition } from "@/lib/prusaSlicerBundle";
 import { splitInheritedImportSet } from "@/lib/importFilaments";
 import { escapeRegex } from "@/lib/matchFilament";
 import { assignSpoolToSlot } from "@/lib/spoolSlots";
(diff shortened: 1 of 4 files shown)


hyiger:
@codex review

hyiger:
Fixed in 7911811b. You're right that the ingestion guards only reach future syncs — an already-polluted bag kept the fix inert. The per-nozzle bake now re-derives over any machine-derivable shape (shared `isAnyMachineDerivedNozzleCondition`, also replacing the collapse's inline regex — one shape definition, three consumers). This deliberately supersedes the #1021 'post-cleanup pure nozzle condition = user pin' contract **on the bake path** — that construction never held on real installs precisely because of the leak this PR fixes; the superseded test now pins both halves (machine shape re-derives, genuine pins still win).

@codex review

hyiger:
Fixed in 3fc8b22c. Right — no exporter vintage ever wrote an index other than 0 (the bake, the derivation, and the #1021 strip all say `[0]`), so `nozzle_diameter[1]==0.4` can only be a user's multi-extruder pin. Both the diameter and HF-tail indices are now anchored to `[0]`; tests pin the predicate rejection and end-to-end pin survival across the fan-out.

@codex review

PR #1046, opened 2026-07-29
fix(printers): filament-side slot assignment now visible on the printer side

hyiger:
Fixes #1041.

**Root cause — data-model divergence, not a stale view.** An AMS slot carries two parallel refs (`filamentId` = loaded filament, `spoolId` = tracked spool). The filament-side assignment path wrote only `spoolId`; `PrinterForm` keys slot rendering on `filamentId` (spool select disabled until a filament is chosen) — so the assignment existed in the data, was fetched fresh on mount, and rendered as an **Empty slot**. Worse: "fixing" the seemingly-empty slot by re-selecting the filament silently wiped the spool ref. The reverse direction worked because the printer form writes both refs while the filament page reads only `spoolId`.

**Fix**: the shared `assignSpoolToSlot` now maintains both refs (the route already had the owning filament in hand), and every spool-keyed clear — the move pass and `clearSpoolsFromOtherPrinters` — nulls both, ending the phantom-loaded-filament staleness. `"Any spool"` dedications (filament set, no tracked spool) never match the spool-keyed filters and are untouched.

**Product calls** (confirmed): overwrite a different loaded filament silently (the assigned spool *is* what's loaded — matches the helper's existing self-healing posture); clearing a tracked spool clears the loaded filament with it.

**Legacy repair**: `amsSlotFilamentIds` startup migration backfills the owner onto orphaned slots (and clears dangling refs to deleted spools), plus a PrinterForm form-state backfill as belt-and-braces so a save persists the repaired pair.

### DIFF
--- a/src/app/api/spools/[spoolId]/assignment/route.ts
+++ b/src/app/api/spools/[spoolId]/assignment/route.ts
@@ -103,7 +103,12 @@ export async function PUT(
       return errorResponse("Printer or slot not found", 404);
     }
 
-    await assignSpoolToSlot(Printer, spoolId, { printerId, slotId });
+    // GH #1041: thread the owning filament so the slot's loaded-filament ref
+    // is maintained alongside the tracked spool — the printer form renders
+    // slots keyed on filamentId, so without it this assignment was invisible
+    // (and trample-prone) on the printer side. The positional projection
+    // above still returns the parent _id.
+    await assignSpoolToSlot(Printer, spoolId, { printerId, slotId }, filament._id);
     const assignment = await findSpoolSlot(Printer, spoolId);
     return NextResponse.json({ assignment });
   } catch (err) {
--- a/src/app/printers/PrinterForm.tsx
+++ b/src/app/printers/PrinterForm.tsx
@@ -282,7 +282,29 @@ export default function PrinterForm({ initialData, onSubmit, onDirtyChange }: Pr
     const ac = new AbortController();
     fetch("/api/filaments", { signal: ac.signal })
       .then((r) => (r.ok ? r.json() : []))
-      .then(setFilamentOptions)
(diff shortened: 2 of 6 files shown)


hyiger:
@codex review

hyiger:
Fixed in a1605dab. Good catch on the third legacy shape — the mismatched pair. Both the migration and the form fallback now compare the tracked spool's owner against the slot's filamentId and repair on disagreement (owner wins — the slot reflects what's physically loaded, same posture as the fixed path). Dangling refs drop the tracking and keep the dedication. The migration test covers all three shapes in one printer.

@codex review

hyiger:
Fixed in 99a9c732 — and this one's worth being blunt about: my round-2 "widening" replaced `$elemMatch` with the bare negated form, which under MongoDB's negated-array semantics requires **every** slot to be non-null — excluding any printer with a single empty slot, i.e. the normal partially-loaded AMS. The widening narrowed. My test missed it because its fixture had all three slots tracked; it now leads with an empty slot as the regression pin.

@codex review

hyiger:
Fixed in 6a106c15 — the arrayFilter now matches the spoolId the repair was computed from (compare-and-set), so a concurrent assignment landing between the migration's read and write makes the stale repair a no-op instead of clobbering it; the next connect re-reads and converges. Branch also updated from main post-#1045.

@codex review

hyiger:
Fixed in 3dcad54d. A CAS miss now counts (via `modifiedCount`) and any miss leaves the migration flag **unset** — the same "will retry on next connect" semantics every other step uses, so the next connect re-reads and converges. Deliberately not an in-process retry loop: a still-running pre-#1041 writer could produce new orphans indefinitely, and chasing it would livelock; deferral converges once that process upgrades or stops writing.

@codex review

PR #1056, opened 2026-08-02
feat(api): opt-in ?shape=spool responses on the five spool-mutation routes

hyiger:
Fixes #1027

## What

The five spool-mutation routes (`POST /spools`, `PUT`+`DELETE /spools/{spoolId}`, `POST …/usage`, `POST …/dry-cycles`) return the whole filament document — every spool's `photoDataUrl` blob (≤5 MB each) and full `usageHistory` — for writes that change one scalar. This is the response-side companion to the #1005 query-projection sweep, in the **opt-in** posture: `?shape=spool`.

## Contract

- **`?shape=spool`** → `{ spool: <the affected spool, complete> }`; the DELETE returns `{ deleted: true, spoolId }` (the post-`$pull` doc no longer holds the spool). The POST locates the created spool via the `instanceId` the route always stamps.
- **No param** → the full document, **byte-identical to today** — load-bearing for shipped mobile builds, which update out-of-band from the desktop and read `.spools` off the body.
- **Unrecognized value** → **400**, so a typo can't silently fall back to the multi-MB default.

Why opt-in rather than a clean break: the desktop auto-updates while the mobile app is EAS-distributed and updates manually, so new-desktop/old-phone skew is the *normal* post-release state — a changed default would corrupt an old phone's screen state after every write.

## Changes

- Pure helpers in `src/lib/spoolResponseShape.ts` (shape parse, `findSpoolById`, `findSpoolByInstanceId`); shape branches sit at the existing return sites — lock scope (`runExclusive`) and the #886/#268 `assignSpoolToSlot` orderings unchanged.
- **Web**: the detail page's six handlers merge/append/filter the affected spool into local state (`mergeSpoolIntoState`) instead of wholesale-replacing `spools`; home-list `moveSpool` + `/inventory` `updateSpool` opt in for pure bandwidth (they never read the success body).
- **Mobile**: `updateSpool`/`logUsage`/`logDryCycle` send `?shape=spool` and **handle both shapes at runtime** (`SpoolMutationResponse`) — a new phone against an old desktop gets the full doc and takes the legacy path; the usage weight refresh reads `res.spool ?? res.filament.spools.find(…)` and the returned `totalWeight` stays gross (the client subtracts tare).
- **OpenAPI**: shared `SpoolResponseShape` parameter, `oneOf` responses on all five operations, spool subdoc schema extracted to `components.schemas.Spool` (re-`$ref`'d from `Filament.spools.items`), DELETE 400 documented.
- **Docs**: new CLAUDE.md section cross-referencing #1005.

## Pre-PR adversarial review (multi-agent)

A 4-lens review (server / web-state / mobile-skew / tests+spec) with refute-style verification confirmed 4 distinct findings, all fixed in the second commit:

1. **Case-sensitivity contract break (reproduced live)**: Mongoose's ObjectId cast accepts uppercase 24-hex, so `PUT …/{UPPERCASE}?shape=spool` *committed* the write and then 404'd (`String(objectId)` is lowercase) — worst class for retrying clients (dry-cycle double-log). `findSpoolById` is now case-insensitive; regression tests at unit + route level.
2. **Duplicate-instanceId race returned the wrong spool**: `$push` appends, so the created spool is last; `.find()` first-match could 201 with a *pre-existing* spool under the documented best-effort id race. Now a tail scan, unit-pinned.
3. **DELETE's OpenAPI didn't declare the 400** its own new param description promises.
4. **Coverage gaps on the exact payloads shipping clients send**: PUT `remainingWeight` (gross-weight response the mobile refresh depends on) and `regenerate` (fresh id the SpoolCard renders) now pinned under `shape=spool`.

## Verification

- 184 files / 3889 tests green (29 new); root + mobile `tsc` clean; root + mobile lint clean.
- Live browser check against the dev server: `/inventory` inline weight edit and the detail-page SpoolCard save both fired `PUT …?shape=spool → 200` with a ~290-byte `{ spool }` body; the SpoolCard re-rendered from the merged spool (placeholder tracked the new value) with no console errors. All test writes reverted.


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

PR #1061, opened 2026-08-04
feat(spools): "Next #" button suggests the next roll number

hyiger:
Fixes #1060.

Adds a **Next #** button to the Add Spool form (both duplicated render sites, one shared handler) that pre-fills the label with the highest numeric spool label anywhere plus one — suggestion-only, freely editable, nothing auto-assigned, per the issue's hard requirement.

**The scan deliberately filters nothing** — retired spools and trashed/purged filaments' spools all count, the opposite of every other spool read in the app. Roll numbers are physical and permanent (the reporter's are hand-written on the spools): a number must never be re-suggested after its record leaves active inventory, or restoring a trashed filament collides with a reissued number. This is also why no existing endpoint could be reused — each filters exactly these rows out.

**Parsing** (pure `src/lib/nextSpoolLabel.ts`, 100% covered): integers only after trimming — no half-parsing `12a` as 12; leading zeros stripped (`0042` → suggests unpadded `43`); >15-digit strings skipped so a 200-char "number" can't poison the max.

**Endpoint**: `GET /api/spools/next-label` → `{next, max}` — label-only projection (#1005 posture), read-only-GET CSRF posture (#360), bearer gate automatic, OpenAPI documented with the scope + concurrency caveats.

**Deferred by product call**: the SpoolCard inline-editor button (its blur-commit editor makes a sibling button race-prone; rarer need) and the soft uniqueness warning (additive `?check=` follow-up if wanted).

**Verified live** against the running app: endpoint correct on a no-numeric-labels DB (`{next:1, max:null}`) and the button click fills the label input with `1`.


hyiger:
@codex review

hyiger:
Fixed in 54fe6d7d — the handler snapshots the label at click time and applies the suggestion only if the field still holds that value when the response lands; any interim typing wins. The field deliberately stays enabled during the fetch (disabling would trade the race for lockout).

Note: this branch's CI failure is the repo-wide audit gate (new upstream advisories), fixed separately in #1062 — this branch will be updated from main once that merges.

@codex review

hyiger:
Updated from main with the audit-gate fix (#1062) — CI should be green now.

@codex review

hyiger:
Fixed in f0bd3c89 — nice boundary catch. The digit-count guard made the generator and its own parser disagree: the 15-nines max suggested a 16-digit number the guard then rejected, so once created, the same number was re-suggested forever. Acceptance is now judged by **value** (`Number.isSafeInteger`), which makes the closure property hold by construction — every emitted suggestion is re-accepted — with saturation at MAX_SAFE_INTEGER instead of an unrepresentable increment. The closure property itself is now a test (suggest → create → suggest yields +1, not a repeat).

@codex review

hyiger:
Fixed in 9ad18404 — the OpenAPI description now states the value-based boundary (safe integers, oversized values skipped), the emit/re-accept closure, and the saturation at MAX_SAFE_INTEGER, replacing the superseded 15-digit rule.

@codex review

PR #1166, opened 2026-08-16
fix(form): the base Max Volumetric Speed is editable again

### LINKED ISSUE #1148: How to modify filament max vol speed ?
During the importation of a CSV I fetch a value for `max vol speed` for a filament. But I didn't find anywhere a way to update this value in the GUI. 

Example of csv file:
```csv
Name,Vendor,Type,Color,Color Name,Secondary Colors,Diameter,Cost,Density,Nozzle Temp,Bed Temp,Nozzle First Layer,Bed First Layer,Max Volumetric Speed,Spool Weight,Net Filament Weight,TDS URL,Shop URL,Drying Temp,Drying Time,Glass Transition,Heat Deflection,Shore A,Shore D,Min Print Speed,Max Print Speed,Nozzle Range Min,Nozzle Range Max,Standby Temp,Transmission Distance,Spool Type,Notes
Black,Arianeplast,PETG,#000000,Black,,1.75,,1.29,235.0,60,,,12.0,,1000.0,,https://www.arianeplast.com/3d-filament-petg/423-petg-noir-3d-filament-arianeplast-fabrique-en-france.html,65,12,80,,,,,,220,250,,,,Subgroup: basic; K-value: 0.003; Fan: 75-100%
```

hyiger:
Fixes #1148 — the reporter imported a CSV with `Max Volumetric Speed` and found no way to edit it in the GUI. The gap is real and render-only: the top-level field is fed by the CSV importer (and four other importers), shown on the detail tile, exported per-preset, and accepted by the PUT — but the form simply never rendered an input for it.

This re-adds the editor as a fourth cell in the Printing tab's **Print Speed** section, with the variant-inheritance placeholder (`parentPh`) like every sibling inheritable scalar. Clearing the field sends null, which clears the stored value and re-enables parent inheritance on variants. The field stays editable on templates (shared spec — not in `TEMPLATE_STRIP_FIELDS`), and the per-nozzle calibration cards' MVS placeholder mirrors the typed base live.

Context on the history: the top-level editor was removed once under an 'MVS is per-nozzle' stance — but the data model kept the base field and every import/export path feeds it, so the removal left an uneditable field rather than a per-nozzle-only model. Re-adding the editor was signed off as the resolution. Per-nozzle calibrations still override the base.

i18n `form.maxVolumetricSpeed` in en+de (parity test enforced). No submit/route/schema changes.


hyiger:
@codex review

hyiger:
Fixed in ba340999 — `step="any"`, accepting stored precision. Worth noting the sibling latent class this exposes: Min/Max Print Speed use `step="1"` while the same import pipeline accepts decimals there too, so a CSV-imported 25.5 would block saves the same way — pre-existing, not introduced here; I'll file a follow-up issue for the class rather than widen this PR.

@codex review

hyiger:
Rebased onto current main by merging it in — the only conflict was the append-only tail of both locale files (main's keys from the #1149/#1150/#1172 merges vs this branch's single `form.maxVolumetricSpeed`). Union-merged with main's ordering as the base: 1838 + 1 = 1839 keys in each locale, parity intact, no other files touched. The i18n parity test and typecheck/lint pass on the merged head.

@codex review

hyiger:
Fixed in fbc6dbb8 — and the class is bigger than the one field: `parentPh` reads **25** fields off `_parent`, the `?raw=true` projection carried **3**. So temperatures, drying, shore hardness, shrinkage, spool/net weights, TDS URL and MVS all rendered blank as inherited placeholders on the EDIT page, while the new-variant page (which fetches the parent doc whole) showed them — a pre-existing gap the new input merely made visible.

The projection now carries the full parentPh set, plus a source-level drift guard pinning the two lists together, since they live in different layers and diverged silently. Failing-before verified with the old 3-field select. (Aside: tsc caught my guard's `s`-flag regex — ES2018 vs the repo's ES2017 target, the same class as the documented `\p{M}` trap — now written with `[\s\S]`.)

@codex review

PR #1170, opened 2026-08-16
fix(slicer): multi-element settings arrays survive the round-trip

### LINKED ISSUE #678: Bambu/Orca import collapses multi-element settings arrays to their first element (compatible_printers round-trip data loss)

hyiger:
Fixes #678 — `unwrap()` collapsed every array bag value to its first element, so a Bambu/Orca preset compatible with several printer/nozzle combos re-exported compatible with exactly one. This implements the deferred NOTE's own prescription: store multi-element passthrough values as arrays and make every scalar-assuming consumer array-aware.

**Parser** (`bambuStudioImport.ts`): length > 1 arrays stored as arrays, each element wire-canonically wrapped (the GH #1070 rule); zero/one-element arrays keep the scalar collapse **byte-identical** — the single-element convention is the overwhelmingly common case and every existing round-trip depends on it.

**Exporters**: PrusaSlicer emits the coStrings form via a new `serializeIniValueList` (semicolon-joined, elements quoted-escaped per `escape_strings_cstyle`'s convention — a `String(array)` comma-join reads back as ONE value); Orca/Bambu pass arrays natively.

**Consumers**: `splitInheritedImportSet` compares element-wise (a parent-equal multi-value key keeps inheriting); `mergeSlicerSettings` wire-wraps array elements; the form displays `;`-joined with the same normalization feeding seed + dirty-compare (unedited fields never write, arrays preserved) and splits an edited value back; the detail amber notice joins for display. `calibrationHints.compatiblePrinters` stays first-element by design — it's only a fallback when `printer_settings_id` is absent, and multi-entry handling (punting ambiguity to `calibrationUnresolved`) is a follow-up.

**Tests**: the old collapse pin is flipped to a preservation pin; coStrings emit, inheritance equality, and a per-id Bambu route → Orca re-export round-trip. Failing-before verified: with the parser arm removed, all three preservation tests reproduce the data loss while the byte-identical single-element pin stays green. 432 tests across the seven affected suites.


hyiger:
@codex review

hyiger:
All three fixed in 1b716432.

**Array-valued gcode crash**: `extractPressureAdvance` joins an array for its M572 scan; the stored array itself is untouched.

**Unedited flattening**: rather than thirty per-field dirty guards (the per-branch-obligation trap), ONE enforcement point after all settings writes — an outgoing scalar equal to the stored array's `;`-join is the unedited case by construction (that join is exactly what the seed produced), so the stored array is restored; an edited value no longer equals the join and writes as typed. This covers the wireOrEdited fields too: unedited, they return `getSettingVal`'s join, which the restore pass maps back to the array.

**Double-encoding**: the root fix — array elements are stored RAW at both ingestion points. Arrays never ride the scalar `key = value` INI path: `serializeIniValueList` escapes at the PrusaSlicer emit (a new test pins single-encoding end to end: raw `Line1\nLine2` element → `"Line1\\nLine2";Plain`, once), and Orca emits raw arrays natively, so real newlines round-trip byte-faithfully there. My round-1 element wrapping was the mistake; the scalar wire-canonical rule is unchanged.

The two form fixes have no component-test harness (repo test scope is lib/models/routes — pre-existing); the parser/emitter behavior is test-pinned.

@codex review

hyiger:
Fixed in 47f57e52 — the restore pass now carries a per-shape unedited test: join-identity for text-backed keys, and for the four checkbox-backed keys (which seed `getSettingVal(...) === "1"` — all four, verified) the outgoing scalar is compared against that exact derivation applied to the stored array. Unedited ⇒ array restored; a genuine toggle writes the scalar the user asserted. The checkbox set carries a lockstep comment pointing at the seed lines so a fifth checkbox can't silently rejoin the flattening class.

@codex review

hyiger:
Fixed in 85b5d9f2 — `settingFlagIsOn` (pure, unit-tested) is now the single array-aware boolean derivation: FIRST element for arrays, reproducing the pre-#678 read byte-for-byte so preserving the array cannot flip any flag that used to read as on. All four reader sites route through it — the form seeds, the unedited-restore mirror (now lockstep with the seed by construction, same helper), the detail page's optTags derivation, and the OpenPrintTag sync route's tag payload. The helper's docblock names every consumer so a fifth reader can't fork the semantics.

@codex review

hyiger:
Fixed in 688728f9 — `compatible_printers_condition` is excluded from array preservation and keeps the first-element collapse. It's the semantically right cut: the key is a single expression (the multi-value work exists for genuine LISTS like `compatible_printers`), and the #1021 ingestion strip's string-only grammar is a design invariant I'd rather not widen — an array-aware strip would churn P1-hardened code to support a shape that means nothing for this key. Pinned with your exact fixture: the duplicated per-extruder expression collapses to one string, which the strip then judges normally.

@codex review

hyiger:
Fixed in d5f08cdd — `parseIniValueList` is the exact inverse of `serializeIniValueList` (round-trip-pinned over the serializer's own cases, including quoted/semicolon/newline elements), and the INI importer applies it to `compatible_printers`: more than one element reconstructs the array (raw elements, same convention as the Bambu ingestion), a single element stays scalar. `compatible_printers_condition` is deliberately excluded and pinned — one expression, per round 5 — including a fixture with a semicolon INSIDE the expression, which list-parsing would have mangled. Prusa INI → import → Orca/Bambu export now keeps list semantics end to end.

@codex review

hyiger:
Both fixed in 1f96f6bd, plus the class member the review didn't name. The quoted singleton stores the PARSED element (pinned; an unquoted scalar parses to itself, so that path is byte-identical). For the equality class: `settingValuesEqual` (unit-tested) now owns the array-aware rule at all THREE sites — `prepareBambuUpdate`'s filter (your finding), `iniImportApply`'s unset scan (same identity compare, found by the class sweep), and `splitInheritedImportSet` (whose inline copy from round 2 is replaced by the shared helper so a fourth site can't fork the rule).

@codex review

hyiger:
Fixed in 9e668847 — every list-capable key inverts on a top-level semicolon now, with the exceptions made explicit as `SCALAR_ONLY_INI_KEYS`: the wire-value texts (gcode/notes — semicolons are content there, and their legacy raw wraps can hold unescaped inner quotes that would defeat the element parser) plus the round-5 condition expression and `inherits`. Your `filament_soluble = 1;0` fixture is pinned, alongside pins that gcode/notes are never split and that a quoted singleton of a non-list key stays verbatim wire text (a literal `"nil"` must not be unquoted into the inheritance marker).

@codex review

hyiger:
Fixed in ba95e529 — `getSettingVal` returns the FIRST element for arrays (the pre-#678 read semantics for every scalar control, and consistent with `settingFlagIsOn`'s convention), so numeric inputs display 0.8 again instead of a browser-blanked control. The one genuinely list-shaped field, `compatible_printers`, moves to its own `getSettingListVal` (join for display, split-back on save) used by both its seed and its edited-only dirty check. The unedited-restore pass's generic arm now mirrors the first-element seed — each field class's unedited test mirrors ITS seed derivation exactly, which is the invariant the last three rounds converged on.

@codex review

hyiger:
Fixed in 2b92dbbe — an outgoing `undefined` normalizes back to the `""` it came from (the `|| undefined` submit writes are the only producer of undefined for these keys) before the unedited comparison, so a blank-first-element array restores like every other unedited field instead of being JSON-omitted into deletion. Non-string/non-undefined outgoing values are left alone — they aren't form-written scalars.

@codex review

hyiger:
Fixed in 77d37555 — the PA scan now reads element 0 only, completing the convention every other reader already follows (and matching pre-#678 semantics exactly, since the old collapse only ever stored element 0). With both controls deriving from the same element, your fixture seeds both blank, the submit emits undefined, and the round-10 normalization restores the array untouched. The round-2 join was the inconsistency; this closes it at the source rather than adding another restore special case.

@codex review

hyiger:
Fixed in 60638007 by taking your second option — encoding type information on export. `serializeIniValueList` now quotes EVERY element unconditionally, which makes an emitted list self-describing: canonical wire scalars escape interior quotes, so an unescaped `";"` separator can't occur inside one, and the strict all-quoted grammar (`isQuotedIniList`) is unambiguous. Generic keys list-parse only on that grammar — your `ACME;Labs` vendor stays scalar (pinned), as does arbitrary semicolon-bearing free text. `compatible_printers` keeps its looser allowlisted parse since PrusaSlicer itself may emit simple unquoted tokens there and its values are names, never free text.

@codex review

hyiger:
Fixed in adf9791b — exactly your reasoning: the r8 exclusion's rationale dissolved when the grammar became self-describing, since a bare semicolon in gcode/notes content can't match the all-quoted form. The two keys rejoin the generic inversion (SCALAR_ONLY keeps only the condition + inherits), a new pin round-trips a two-element gcode, and the r8 never-split pins still pass because their fixtures don't match the grammar. One residual stated in the docblock: a LEGACY raw-wrapped note containing a quote-aligned `";"` sequence could false-match — a shape no canonical writer produces.

@codex review

hyiger:
Fixed in 6febec01 — the emitter side of your either/or: a length-1 array collapses through `serializeIniValue` (single-element array ≡ scalar is already the parser convention at both ingestion points), a length-0 array emits an empty value, and only genuine multi-element lists use the strict quoted grammar. Pinned that `["1"]` emits `filament_soluble = 1`, not a quoted singleton.

@codex review

hyiger:
Fixed in 6b09a722 — coerced at `serializeIniValueList` itself, so no caller can reintroduce the crash, and swept the class it belongs to: every array-element consumer that inherited the string assumption. Two more had it — `settingFlagIsOn` (a numeric `1` now reads ON, restoring parity with pre-#678's `unwrap`, which did `String(value[0])`) and `getSettingVal`/`getSettingListVal` (whose declared string return type was a lie feeding form state). The singleton and Orca paths already coerced. Pinned with numeric/boolean/null fixtures.

(The commit was rebased onto the merged main, so the sha in this comment was updated from the pre-rebase one.)

@codex review

hyiger:
Fixed in 0c04d870 — every arm now stores from the already-trimmed, empty-filtered `parts`: >1 the list, exactly 1 that element, 0 an empty string (PrusaSlicer's "no restriction"). Your `A;` case stores `A`, and the same normalization also catches surrounding whitespace, which would have hidden the preset just as effectively.

(Sha updated after a rebase onto the branch's merge commit.)

@codex review

hyiger:
Fixed in 5bfbc989 — but structurally rather than by patching the comparison, because this exact mirror has now drifted from its seed twice (r9's stale join, r15's missing coercion), flattening a stored array both times. The derivation is extracted as `displayFirstElement`: `getSettingVal` seeds through it and the unedited-restore pass compares through it, so there is no second place left to forget. Your `[0.8, 1.2]` case now compares `"0.8"` against `"0.8"` and restores the array.

@codex review

hyiger:
Fixed in 8b2d5cb3 — with one correction to the diagnosis. **The infinite loop does not occur.** I probed every malformed shape (`"A" ; "B"`, `"A" "B"`, `"A"x;"B"`, unterminated quote, empty) under a watchdog: all return in 0ms, because after a quoted element the non-separator path falls into the unquoted branch, which always advances via `indexOf` or breaks.

**But the underlying observation is right, and the real defect is worse than cosmetic**: `"A" ; "B"` parsed to `["A", " ", " \"B\""]` — three printer names, two of which match nothing, hiding the preset exactly like the bugs this PR exists to fix.

So: whitespace around separators is now tolerated (the hand-formatted shape yields `["A","B"]`), and when what follows an element is neither a separator nor end-of-value, the parser returns the WHOLE value as a single element instead of inventing names — the conservative scalar fallback callers already handle; unterminated quotes take the same path. I also added a structural progress assertion so the hang you described is impossible by construction rather than by accident. Pinned both the tolerant and fallback shapes.

@codex review

hyiger:
@codex review

PR #1172, opened 2026-08-16
fix(nfc): Erase probe-derives NTAG capacity on GET_VERSION-dead readers

### LINKED ISSUE #978: NFC Erase (formatNtagImpl) hard-fails on GET_VERSION-dead readers (e.g. ACR1552U)
Follow-up from the #973 hardware investigation.

## Gap
On-hardware capture shows the **ACR1552U rejects the NTAG `GET_VERSION` command** (SW `0x6900`). The #973 follow-up made the **Write** path stop depending on GET_VERSION — it prompts for the NTAG size and the pick is authoritative (rewrites the CC). But the **Erase** path (`formatNtagImpl` in `electron/nfc-service.ts`, ~line 1460) still calls `getNtagNdefBytesViaGetVersion` and throws `NTAG_SIZE_UNKNOWN` when it returns null, with no size-picker fallback.

So on a GET_VERSION-dead reader: **Write works, Erase never can** — an incoherent split for the same reader.

## Mitigation already in place
Lower priority than it sounds: the Write path now **rewrites the capability container to the user-picked size** (`resolveNtagWriteSize` → `needsFormat`), so a tag an earlier bad write mis-sized is corrected by writing to it — the user doesn't need Erase to un-stick a tag.

## Fix sketch
Thread a user-declared NTAG size into the Erase flow: a size picker in Settings → Devices → NFC Tools (when erasing an NTAG), passed through `nfc-format-tag` IPC → `formatTag` → `formatNtagImpl`, which uses the same `resolveNtagWriteSize` decision (blank/formatted + hint) to size the fresh CC, gated by the existing `FF B0` brick-guard probe. Alternatively, until then, give `formatNtagImpl`'s NTAG_SIZE_UNKNOWN a distinct message that doesn't imply a picker exists on the Erase flow.

hyiger:
Fixes #978 — on the ACR1552U (GET_VERSION confirmed-rejected, SW `0x6900`) **Write worked while Erase never could**.

**Design (round 2, after Codex correctly killed the picker)**: when GET_VERSION is silent, `probeNtagCapacity` derives the size by descending the ladder (216 → 215 → 213) with non-mutating `FF B0` page reads — the same NAK-past-physical-end chip behaviour the Write path's hardware-proven brick guard is built on, aimed at deriving instead of validating. This closes BOTH failure directions: no oversized erase (a 213 never probes as 215/216) and no undersized one (a 215 probes as 215 no matter what any remembered default says — the round-1 picker design would have zero-filled part of the chip, stamped a smaller CC, and reported success with stale bytes left behind). Zero prompts; the 213 floor is always physically present since the head already read.

**Pure decision** (`resolveNtagEraseSize`, unit-tested): `verSize ?? probedBytes ?? refuse` — GET_VERSION stays authoritative when it answers (an erase at the true chip size is how a mis-formatted tag gets restored). A transport blip mid-probe can only under-size — recoverable by re-erasing, never writing past the chip's end.

**Posture change (signed off in triage)**: the GET_VERSION-only family proof is replaced on this path by the same guard set the hardware-proven Write path relies on — byte-0 NXP guard, CC `0xE1`/`0x00` guard, page-bound writes.

IPC and UI are untouched (round 1's picker machinery reverted as dead). `electron/` has no test harness (established scope); the size decision is pinned by the pure resolver tests. On-hardware confirmation of the Erase flow on the ACR1552U remains owed.


hyiger:
@codex review

hyiger:
Fixed in 8ed43197 by taking your first option — probe-derive the actual capacity — and it obsoletes the picker entirely. When GET_VERSION is silent, `probeNtagCapacity` descends 216 → 215 → 213 with non-mutating page reads: the same NAK-past-physical-end behaviour the Write brick guard already trusts on this hardware, aimed at deriving the size instead of validating a pick. That closes the undersize hole (a 215 probes as 215 regardless of any remembered default) and the oversize one in the same stroke, with zero prompts. The 213 floor is always physically present (the head already read); a transport blip mid-probe can only under-size, which is recoverable by re-erasing and never writes past the chip's end.

The user-declared-size machinery this replaces is reverted as dead code: IPC trio byte-identical to main again, no Settings picker flow, no shared dialog. `resolveNtagEraseSize` stays pure and tested (`probedBytes` replacing `hintBytes`); its refuse arm is now defensively unreachable with an honest message. PR title/body updated to match. On-hardware confirmation on the ACR1552U remains owed, as flagged.

@codex review

hyiger:
Fixed in 357df0ae — all three sites now describe the probe-derived sizing: the EN and DE NFC guides' erase callouts (auto-sizing, no prompt, no failure, mis-sized tags restored by the erase), and CLAUDE.md's hardware-status bullet flips from 'knowingly broken / #978 OPEN' to the shipped behavior with the on-device-confirmation-owed caveat retained.

@codex review

hyiger:
Fixed in 44e95b17 — only a DEFINITIVE chip answer demotes a rung now. `readNtagBurst`'s two failure shapes are distinguishable: an `SW=` error means the chip answered and NAK'd the out-of-range read (definitive — smaller), while a transport throw from `transmit()` proves nothing. Transport failures retry (3 attempts per rung) and, when persistent, the erase FAILS CLOSED with nothing written (`NTAG_PROBE_FAILED`, i18n-mapped in the panel like the sibling erase errors). The under-size-only residual documented in the probe docblock now applies solely to a chip that genuinely NAKs in-range reads — not to reader blips.

@codex review

hyiger:
Both fixed in 7fb2834c. The 213 rung is probed like its siblings — a definitive NAK there means a sub-213 NXP chip (NTAG210/212, which pass the head guard while the GET_VERSION parser deliberately nulls them) and the erase REFUSES with nothing written (`NTAG_UNSUPPORTED`, i18n-mapped like the sibling errors); the unconditional floor return is now unreachable and documented as such. And the stale docblock + CLAUDE.md text from round 3 now state the fail-closed semantics: the residual mis-size requires a chip that definitively NAKs a genuinely in-range page — not a reader blip.

@codex review

hyiger:
Fixed in 3bf0d6a4 — every rung now probes its chip's PHYSICAL last page (`NTAG_PHYSICAL_LAST_PAGE`: 44/134/230, config tail) instead of the user extent. The discriminating fact your finding supplied — a smaller chip's own dynamic-lock/config pages stay readable right after user memory — means the 213 rung's probe at page 41–44 lands past an NTAG212's entire readable extent (which ends at 39) and NAKs, refusing with `NTAG_UNSUPPORTED`. The constants live in the pure module with the datasheet rationale, pinned by an ordering test whose key assertion is 44 > 39; config-page reads on the true chip are non-mutating, and the erase's WRITE extent is unchanged (user pages only). Docblock + CLAUDE.md updated in the same commit.

@codex review

hyiger:
Fixed in daf1dd00 with the read/NAK asymmetry, encoded in the pure resolver: a successful read PROVES pages exist (auth cannot fake it), so probe > CC stays trusted — exactly the mis-formatted-small restore case this feature exists for — while a NAK is ambiguous (absent page vs auth-protected page, and PC/SC readers don't reliably distinguish the SWs), so NAK-derived DOWN-sizing is accepted only when the tag's own CC agrees. Your protected-216 example carries its 872-byte CC, so probed(496) < cc(872) refuses with `NTAG_SIZE_CONFLICT` and nothing written. A blank CC never conflicts; GET_VERSION stays authoritative over everything. Six resolver cases pin the matrix; CLAUDE.md documents the rule.

@codex review

hyiger:
Fixed in 2c1f7300 by conceding the general point your four findings collectively proved: every sub-216 conclusion rests on a NAK, every NAK is ambiguous, and each tie-breaker I proposed (user pick, blanket demotion, user-extent probes, the CC cross-check — which your protected+mis-formatted example broke in the second direction) was a heuristic with a hole. The design now accepts only PROOFS: GET_VERSION's answer, or reads covering the full NTAG216 extent (a successful read cannot be faked by auth). The ladder collapses to one probe — a full 216 erases hands-free on the ACR1552U; anything else refuses as `NTAG_SIZE_AMBIGUOUS` with the honest disjunction and Write NFC (explicit size + CC rewrite) named as the recovery. That narrows the GET_VERSION-dead convenience to 216 tags, but it is the sound boundary, and the issue's own mitigation note already covers the rest. Docs (EN/DE) and CLAUDE.md describe the final behavior including why each heuristic fell.

@codex review

hyiger:
Fixed in d9a4d488 — a tail read proves capacity, not writability, so the zero-fill's historical `catch{break}` (which treated any write NAK as end-of-tag) is gone: with the size PROVEN, a write failure inside the extent throws `NTAG_WRITE_REFUSED`, naming the page, the write-protection hypothesis, and — honestly — that the erase is INCOMPLETE, since the CC/TLV/prefix were already written and an overwrite has no rollback. This also fixes the same lying-success on GET_VERSION-capable readers, where the tolerant break predated this PR. i18n-mapped en+de.

@codex review

hyiger:
Fixed in c806156e — the page number is extracted from the service error and interpolated into the localized message ({page}, en+de), so the user sees exactly where the wipe stopped: pages before it cleared, data from it on remaining.

@codex review

hyiger:
Fixed in a2305e90 — the proof is now a COMPLETE 16-byte burst: an SW=9000 short payload (a shape the head/image paths already tolerate as possible) proves nothing about pages 227–230, so it returns null and the erase refuses as ambiguous, consistent with the proofs-only rule.

@codex review

hyiger:
Fixed in 25a223e8 — short SW=9000 answers join the TRANSPORT class: retried ×3, then the erase fails closed as inconclusive via `NTAG_PROBE_FAILED`, whose copy already carries the re-seat-and-retry guidance. The `NTAG_SIZE_AMBIGUOUS` negative-proof refusal is now reachable only from a definitive chip NAK, which is the only case entitled to claim one.

(Correction: this comment briefly cited a wrong commit sha from a wrong-branch slip — the fix is 25a223e8 on this branch.)

@codex review

PR #1174, opened 2026-08-16
feat(sync): surface hybrid-REMOTE name conflicts

### LINKED ISSUE #1164: Data health: surface hybrid-REMOTE name conflicts via sync status
Follow-up to #1149 (PR #1162), deferred there by design.

The Data health page scans the database the SERVER talks to — in hybrid mode, the local one. Trim-collision conflicts that exist only on the REMOTE (Atlas) side are visible to nothing but the desktop sync service: the remote never runs `dbConnect`, and the sync service's own trim pass reports its refusals only to the console (`electron/sync-service.ts` — the same conflicts gate name reconciliation via `conflictedNames`).

Shape of a fix, as scoped during #1149:
- add a `nameConflicts` field to `SyncStatus` carrying the sync-side trim pass's ACTIVE conflicts (the enriched `TrimNameConflict` shape from #1162 — id/trimsTo/reason/collidesWith — is already produced there);
- render them in the sync details dialog, and/or link to the Data health page with a remote-side section;
- resolution still happens by id-addressed writes — the sync engine propagates a rename/trash made locally, but a remote-only row has no local twin, so the surface may need to say "resolve on the machine that owns this database" for some shapes.

The page's copy already states the limitation (`health.hybridNote`), so this is an enhancement, not a gap in what #1162 claims.

hyiger:
Fixes #1164 — the deferred half of #1149.

The Data health page scans the database the **server** talks to (the local one in hybrid mode), so a trim collision living only on the Atlas side was visible to nothing. The sync pass already computes the enriched conflict — #1149's `id`/`trimsTo`/`reason`/`collidesWith` — and then discarded it: a console line plus a name-set used only for gating.

**Additive reporting; the gating logic is untouched.** `SyncStatus` gains `nameConflicts` (the enriched shape plus a `local`/`remote` tag), collected in the per-side trim loop under the *same* active-only rule the gate uses (an inactive conflict is permanent and invisible in the UI — nothing a user could act on), and published once after both sides so each cycle reports a complete set and a resolved conflict disappears rather than lingering. Only cycles that reach the trim phase publish, so a pre-trim connect failure leaves the previous, still-accurate list alone.

**Surfaces**: the sync tooltip lists them (amber, first three, link to Data health) — deliberately *not* wired into the pill's state machine, since `partial` means a collection failed to sync, which these do not. Data health gains a read-only **remote** section whose copy splits the two resolutions: a pair that also exists locally is fixed above and syncs across; a remote-only row has no local twin and must be resolved where it lives. Local conflicts aren't duplicated there.

The IPC trio stays in lockstep (main's no-service default resets the field; both hand-mirrored `electron.d.ts` types carry it, optional so an older main still typechecks against a newer renderer).

**Tests**: three integration cases — a remote-only conflict surfaces side-tagged with its enriched fields; the list refreshes wholesale (resolved → empty, not accumulating); an inactive conflict is never reported. Failing-before verified by removing the publication line (two of the three fail).

Drive-by: removes an unused `NTAG216_MAX_NDEF_BYTES` import that #1172's final round left behind — a lint warning currently on main.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1189,6 +1189,9 @@ ipcMain.handle("get-sync-status", (event) => {
     lastSyncAt: null,
     error: null,
     progress: null,
+      // GH #1164: mode switches destroy the service, so the field resets
+      // with it — an absent list must not read as a stale one.
+      nameConflicts: [],
   };
 });
 
--- a/electron/nfc-service.ts
+++ b/electron/nfc-service.ts
@@ -31,7 +31,6 @@ import {
   resolveNtagEraseSize,
   NTAG_NAME_TO_NDEF_BYTES,
   NTAG_PHYSICAL_LAST_PAGE,
-  NTAG216_MAX_NDEF_BYTES,
   type NtagSizeName,
 } from "../src/lib/ntagVersion";
 import {
--- a/electron/sync-service.ts
+++ b/electron/sync-service.ts
@@ -13,6 +13,7 @@ import {
   type TrimNameConflict,
   trimEntityNames,
   describeTrimResult,
+  TRIMMABLE_COLLECTIONS,
   type MinimalTrimDb,
 } from "../src/lib/trimEntityNames";
 import {
@@ -515,12 +516,22 @@ export class SyncService extends EventEmitter {
     // Declared outside the try so the catch below can publish them: a cycle
     // that dies after the trim passes still knows what those passes saw.
     const trimConflicts: SyncNameConflict[] = [];
-    // Only a scan that covered BOTH peers AND every collection is
-    // publishable. A partial one — thrown, aborted mid-loop, or silent about
-    // a collection the trim skipped — would drop rows it never looked at,
-    // which is worse than the stale list it would replace.
-    let trimScanCompleted = false;
-    let trimSkipped = false;
+    // Which `side:collection` pairs the trim actually LOOKED at (Codex P2
+    // round 6). Publishability is per pair, not global: the snapshot is
+    // authoritative about what it scanned and silent about the rest, so a
+    // single skip must not suppress a fresh conflict another collection just
+    // found — that hid it indefinitely whenever the skip and the failure both
+    // repeated. A pair the pass skipped, or a side it threw on, keeps
+    // whatever the previous status said.
+    const scannedPairs = new Set<string>();
+    const pairKey = (side: string, collection: string) => `${side}:${collection}`;
+    /** Fresh rows for scanned pairs; the previous status for everything else. */
+    const mergeWithPrevious = (fresh: SyncNameConflict[]): SyncNameConflict[] => [
+      ...fresh,
+      ...(this.status.nameConflicts ?? []).filter(
+        (c) => !scannedPairs.has(pairKey(c.side, c.collection)),
+      ),
+    ];
 
     try {
       await local.connect();
@@ -726,15 +737,18 @@ export class SyncService extends EventEmitter {
         // A collection the pass SKIPPED has un-normalized names by
         // definition, and it doesn't know WHICH — so that one really is
         // collection-wide.
-        for (const sk of trimResult.skipped) conflictedCollections.add(sk.collection);
         // A skip means the pass never looked at that collection's rows at all
         // — it gave up before querying any candidate (Codex P2 round 4). The
-        // snapshot is therefore silent about it, and publishing that silence
-        // would erase a previously-reported conflict there rather than
-        // refresh it. Same rule as the abort below, for the same reason.
-        if (trimResult.skipped.length > 0) trimSkipped = true;
+        // snapshot is therefore silent about it rather than clean, so record
+        // only what was really scanned; everything else keeps the previous
+        // status's rows. A side this loop THROWS on records nothing, since
+        // the throw discards whatever it had found so far.
+        const skippedHere = new Set(trimResult.skipped.map((sk) => sk.collection));
+        for (const sk of trimResult.skipped) conflictedCollections.add(sk.collection);
+        for (const collection of TRIMMABLE_COLLECTIONS) {
+          if (!skippedHere.has(collection)) scannedPairs.add(pairKey(side, collection));
+        }
       }
-      trimScanCompleted = !this.aborted && !trimSkipped;
       // (GH #1164 round 2: the conflict list is published AFTER the
       // collection copies — see the rescan near the end of the cycle. This
       // pre-copy point is too early: a conflict the user resolved locally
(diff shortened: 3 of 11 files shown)


hyiger:
@codex review

hyiger:
Fixed in 77bc2629 — you're right that the trim pass's snapshot is taken too early, and #1149 already provided the right tool for the fix: the list is now built by a post-copy RESCAN using the read-only `scanTrimConflicts` (no index build, no writes, classifies identically to the migration), so it reflects post-sync reality instead of pre-copy state. Best-effort by design: a rescan failure leaves the previous list rather than failing an otherwise-good cycle, and only a completed rescan overwrites the field. The gating path (`conflictedNames`, built during the trim pass) is untouched — this only changes reporting.

New test seeds your exact shape (remote carries the untrimmed twin; the newer local row renames it during the copy) and asserts an empty list; it fails against the pre-copy publication.

@codex review

hyiger:
Fixed in ae11dfef — the all-clear now requires BOTH sides clean, so "your data is healthy" can't render above an amber remote list. I also derived the remote set once (`remoteConflicts`) and used it for both the banner condition and the section, so the two can't disagree about what "remote conflicts exist" means — the duplicated `.filter()` was how they diverged in the first place. On web (no Electron API) the list is always empty, so the banner behaves exactly as before.

@codex review

hyiger:
@codex review

hyiger:
Fixed in 125460fa — the status listener refetches `/api/name-conflicts` when a cycle completes, so a conflict the sync copies INTO the local database moves from the read-only remote section into the actionable list the copy points at. Keyed on `lastSyncAt` changing rather than on any status message: progress ticks within a cycle carry the same stamp and are ignored (no refetch loop), and the first observation is skipped because that IS the mount snapshot.

@codex review

hyiger:
Fixed in 3ca8596d — correct diagnosis: I overloaded `null` as both "unset" and a real stamp value, so a page mounting mid-initial-sync (or after connection failures) treated the first genuine completion as its own mount snapshot and skipped the refetch. A separate `seenInitial` flag now distinguishes the two: the mount observation is recorded whatever its value, and every subsequent change refetches.

@codex review

hyiger:
Fixed in 7d104186 — "a cycle ended" is now *stamp advanced OR left the syncing state*, so error and partial terminals refetch too. Your partial-write path is exactly right: the terminal error status keeps the previous `lastSyncAt`, but collections copied before the throw can already have brought a remote conflict local. Progress ticks stay inside the syncing state with an unchanged stamp, so the no-loop property is unaffected.

@codex review

hyiger:
Fixed in f9057501.

The trim passes now keep their conflict snapshot for the error path. It publishes when the cycle throws before the rescan (and as a fallback when the rescan itself fails), gated on the scan having covered **both** peers — a partial scan would silently drop the unscanned side, which is worse than the stale list it replaces. On the success path the post-copy rescan still wins, so the r2 "don't report what this cycle just fixed" property is intact; the pre-copy snapshot only over-reports for one cycle in the failure case, and that self-corrects.

Pinned by `publishes the trim-phase conflicts when the cycle DIES before the rescan (r3)` — a remote-only bedtype conflict with `reconcileNozzlesByName` stubbed to reject (the same shape as a mid-cycle remote read failing). Verified failing before the fix.

@codex review

hyiger:
Fixed in fcc9a0fd.

Right — a skipped collection isn't clean, it's *silent*: `trimEntityNames` gives up before reading any candidate row when it can't establish the protective index. Publishing that silence would delete a live conflict there rather than refresh it. `trimScanCompleted` is now `!aborted && !trimSkipped`, so the snapshot may replace the published list only when it covered both peers **and** every collection; otherwise the previous list stands.

Pinned by `does not erase a conflict in a collection the trim SKIPPED (r4)`: one instance, two cycles — cycle 1 publishes a remote-only bedtypes conflict, then the index is dropped and a duplicate active name inserted so cycle 2's trim refuses the rebuild and skips, with the cycle dying at the same injected point. The remote conflict must survive. Verified failing before.

@codex review

hyiger:
Fixed in 1e1d01c8.

Confirmed the race: the listener can deliver before `getSyncStatus()` resolves, that event claimed the mount baseline, and the snapshot arriving afterwards carried the same terminal stamp — so no refetch ever fired for the very completion the rule exists for. Now the snapshot **only** seeds the baseline (and only if no event already did, so a late-resolving promise can't rewind it and re-report), and every live event is treated as post-mount: with no baseline yet, a non-syncing event *is* an ending.

The rule itself moved into `src/lib/syncCycleWatcher.ts` — pure, coverage-gated, 9 tests. This page has no component-test harness, and this rule has now been wrong three separate ways, each leaving the actionable list silently stale; same reasoning that put the NTAG page arithmetic in `ntagPages.ts`/`ntagVersion.ts` rather than in `electron/`. The tests pin the beats-the-snapshot case, the late-snapshot rewind, the failed-cycle stamp, progress ticks, and null-is-a-real-stamp.

Gate green: 214 files / 4670 tests, thresholds clear.

@codex review

hyiger:
Fixed in 490aa99b — both, and they're the same miss: I protected the watcher's baseline from the losing-race snapshot but not the React state.

**Health page.** The mount promise could resolve after a terminal event and reinstall its older `nameConflicts`, dropping a remote-only conflict from the page until some later status arrived. It now yields to a live event (`sawLiveStatus`), for both the state and the seed.

**SyncStatusIndicator.** Same shape, and you're right that it matters more there: the indicator's conflict count and its Data health link may be the only notice a remote-only conflict ever gets, so rolling back to a pre-completion snapshot hides it entirely. Same guard.

Worth noting this wasn't strictly new in the indicator — the unconditional `setStatus` predates the PR — but the PR is what made it user-visible, so it belongs here.

@codex review

hyiger:
Fixed in 8a389711.

**Per-collection publishability.** Agreed — the all-or-nothing flag traded one silent-loss mode for another, and the repeating skip + repeating failure case makes it indefinite. The snapshot is authoritative about the `side:collection` pairs it actually scanned and silent about the rest, so it merges rather than replacing: fresh rows for scanned pairs, the previous status for everything else. A side the loop *throws* on records nothing (the throw discards whatever it had found), so a cycle that dies before the trim scans anything merges to exactly the previous list — the pre-PR behaviour, preserved as the floor.

**Health page request ordering.** Also real: the mount scan and the refetch are independent `/api/name-conflicts` requests, and the mount one can capture the pre-copy database and resolve last. Every scan now takes a ticket from a `scanSeq` ref and only the newest may write — covering `load()` (the completion refetch and the action handlers) and the mount IIFE alike.

New test `still publishes a SCANNED collection's fresh conflict when another was skipped (r6)`: remote bedtypes carries duplicate active names so its index build is refused and it's skipped before any row is read, while remote locations holds a real trim conflict; the cycle dies at the injected point and the locations conflict must still be published. Verified failing against a whole-side skip rule.

Gate green: 214 files / 4687 tests, thresholds clear.

@codex review

PR #1175, opened 2026-08-16
feat(api): calibration lookup can be scoped to a printer (#1047 Phase 0)

hyiger:
The zero-migration quick win from #1047. `calibrations[].printer` has been populated since #107, but the read route had no way to **ask** for it — so a slicer on the XL received whichever same-nozzle entry happened to sort first, possibly one tuned for a different machine. The compound nozzle identity, the canonical `Nozzle.printer` side and the dedupe migration stay in #1047's later phases; this changes no schema.

`?printer=<name|ObjectId>` follows the convention `bed_type` already uses, and is **soft** — matching the `nozzle_type` posture directly beside it: a printer match wins, else the shareable printer-less defaults, else the unscoped list. A printer/data mismatch never turns a working lookup into a 404.

**One deliberate behavior change**, documented in the OpenAPI description: with *no* printer param the printer-less defaults are now preferred instead of "whatever sorted first". That's deterministic, and it matches what `prusaSlicerBundle` already prefers when baking a preset. A filament whose only entries are printer-scoped still answers with one (test included), so nothing that used to resolve stops resolving.

Five route tests: named printer, ObjectId form, unknown-printer fallback, the no-param default preference, and printer-scoped-only. Failing-before verified by removing the block (three fail).

Note for the fork side: the sync-back WRITE path still has no printer context, so a printer-scoped row can shadow slicer edits — that's called out in the issue and needs a coordinated hyiger/PrusaSlicer change, not this PR.

### DIFF
--- a/public/openapi.json
+++ b/public/openapi.json
@@ -1290,6 +1290,15 @@
             "example": "Smooth PEI"
           },
           {
+            "name": "printer",
+            "in": "query",
+            "required": false,
+            "schema": {
+              "type": "string"
+            },
+            "description": "Printer NAME or ObjectId (same convention as bed_type). Selects the calibration tuned for that machine (GH #1047 Phase 0). Soft: an unknown printer falls back to the shareable printer-less defaults, then to any same-nozzle entry — never a new 404. With NO printer param the printer-less defaults are preferred (deterministic; previously the first matching entry won regardless of its printer)."
+          },
+          {
             "name": "format",
             "in": "query",
             "schema": {
--- a/src/app/api/filaments/[id]/calibration/route.ts
+++ b/src/app/api/filaments/[id]/calibration/route.ts
@@ -113,7 +113,7 @@ export async function GET(
       nozzle?: { diameter?: number; name?: string; type?: string; highFlow?: boolean };
       // `_id` is an ObjectId at runtime (populated doc); typed loosely like
       // the bedType entry below so the lean() cast still overlaps.
-      printer?: { _id?: unknown; name?: string };
+      printer?: { _id?: unknown; name?: string; _deletedAt?: unknown };
       bedType?: { _id?: string; name?: string; material?: string } | null;
       extrusionMultiplier?: number;
       maxVolumetricSpeed?: number;
@@ -192,8 +192,15 @@ export async function GET(
       // above. Ids are compared case-folded because a populated `_id`
       // renders canonical lowercase.
       const looksLikeId = /^[0-9a-fA-F]{24}$/.test(raw);
+      // Only a LIVE printer is addressable (Codex P2 round 4). `populate()`
+      // does not filter tombstones, so a calibration still pointing at a
+      // soft-deleted printer would satisfy an id match and outrank an ACTIVE
+      // printer named with that id — the impostor problem inverted. Applied to
+      // every rung below, so "exists" means the same thing here as it does in
+      // the existence queries and in the rest of the printer API.
+      const addressable = scopedMatches.filter((cal) => cal.printer && !cal.printer._deletedAt);
       const idMatches = looksLikeId
-        ? scopedMatches.filter(
+        ? addressable.filter(
             (cal) => String(cal.printer?._id ?? "").trim().toLowerCase() === wanted,
           )
         : [];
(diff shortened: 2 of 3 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
Fixed in a435f160 — both sides now normalize through the same trimmed, lowercased `wanted` the name compare already used, so an uppercase-hex or space-padded id matches. Worth naming why this mattered more than a typical parse nit: the soft fallback made the failure SILENT — the caller got a calibration, just not its machine's. Pinned with an uppercased, space-padded id (pre-fix it returned the printer-less default).

@codex review

hyiger:
Both fixed in ff1407e5.

**Bedless fallback**: the narrowing now PRIORITIZES instead of filtering — printer matches first, shareable defaults retained behind them. `scopedMatches[0]` still prefers the machine, but the bed-type step can still see the generic entry, so your `(P,Textured)+(null,null)` case with `bed_type=Smooth` returns the bedless default as documented. The no-param branch does the same in reverse, which also preserves the printer-scoped-only case.

**Exact name first**: right that the Printer name index is case-sensitive, so `XL`/`xl` can coexist and array order was deciding. An exact trimmed match is tried first, with the case-folded (and ObjectId) compare as fallback.

Both pinned; failing-before verified against the filter-based narrowing.

@codex review

hyiger:
Fixed in c71a921f — a 24-hex input is now resolved as an ObjectId before any name matching (id → exact trimmed name → case-folded name), which matches how the filament itself is resolved in this same route and keeps the documented ObjectId support meaningful. Pinned with a printer literally named as another printer's id, where the impostor previously won.

@codex review

hyiger:
Fixed in 2945b61a.

Existence is now asked of the Printer collection rather than inferred from the already-filtered rows — `Printer.exists({_id: raw})`, queried only when the populated rows didn't already prove it, so the hit path adds no query. When the id belongs to a real printer that simply has no calibration for the requested nozzle, name matching stays off entirely and the lookup falls to the shareable default; the name path is reachable only when the 24-hex string belongs to nobody.

Two tests: an existing-printer-with-no-row case that must resolve to the default rather than the name twin (verified failing before), and its converse — a hex-shaped name no printer owns as an id still resolves by name.

@codex review

hyiger:
Fixed in 2b174220 — both findings, since they're the same question asked of two identities.

**Exact name.** An exact-name match now gets the same existence check as the id: with `XL` and `xl` both active and `XL` having no row for the requested nozzle, case folding no longer hands back `xl`'s calibration — the lookup falls to the shareable default. Queried only when folding would actually change the answer (`loose` non-empty, no exact row, id path not already decided), so the ordinary lookup still adds no query.

**Soft-deleted printers.** Both existence checks now carry `_deletedAt: null`, matching what the rest of the printer API means by "exists". A tombstoned row can no longer suppress the name path for an active printer named with its id.

The ternary chain became an if/else ladder at four rungs — it reads as the precedence list it is. Two more tests, each verified failing before: the exact-name-twin case, and a soft-deleted id whose active namesake must still resolve.

@codex review

hyiger:
Fixed in f54e8f06 — both, since they're the same predicate applied inconsistently.

**Verbatim spelling.** The name ladder is now three rungs: verbatim (caller's spelling vs stored spelling, no normalization on either side) → trimmed-exact → case-folded. Only the first can tell `X` from `X `, which raw-driver sync writes make reachable. The trimmed rung stays so a caller's stray whitespace still resolves.

Also moved the name existence query to the **native driver**: the schema's `trim` setter rewrites query *values* too, so `Printer.exists({name: "X "})` casts to `"X"` and answers about the wrong row — the GH #1116 trap. It now asks `{name: {$in: [rawInput, trimmed]}, _deletedAt: null}`, which also matches rows predating the field.

**Populated tombstones.** Right — `populate()` doesn't filter, so a calibration pointing at a soft-deleted printer satisfied the id match and outranked an *active* printer named with that id. Every rung now matches only live printers via one `addressable` filter, so "exists" means the same thing in the rows and in the queries.

Two tests, each verified failing before: a whitespace twin where the verbatim spelling must win (raw-driver inserts to bypass the trim setter), and a deleted printer whose own row must not outrank its active namesake.

@codex review

hyiger:
Fixed in 8a15fbdb.

Confirmed: the combined existence check never fired when the trimmed rung had matches, so a live `X ` with no row for the requested nozzle was still answered by `X`'s row. The question is per-rung, so it's now asked per rung — `liveNamed(spelling)` sits between each rung and the next weaker one, reached (and therefore queried) only when the rung above found nothing and a weaker one would otherwise answer. The ordinary lookup stops at the verbatim rung and issues no query at all; the trimmed-spelling query is additionally skipped when it would repeat the verbatim one.

Pinned by `an existing VERBATIM name with no row here does not fall to its trimmed twin (r5)` — `VT ` and `VT` both active via raw-driver inserts, `VT ` calibrated only on 0.6, request for `VT ` at 0.4 must return the shareable default. Verified failing against the previous combined gate.

@codex review

PR #1176, opened 2026-08-17
Bump version to 1.78.0

hyiger:
Version bump across the three tracked files ahead of the v1.78.0 tag:

- `package.json`
- `package-lock.json` (both occurrences)
- `public/openapi.json` (`info.version`)

Bump-before-tag is machine-enforced since #1037 — the `tag-version` job fails the gate when a pushed `v*` tag doesn't match `package.json`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "filament-db",
-  "version": "1.77.0",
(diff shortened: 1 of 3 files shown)


hyiger:
@codex review

PR #1181, opened 2026-08-27
fix(openprinttag): tokenized multi-field search in the OPT link dialog + browser

### LINKED ISSUE #1173: Improving the searching when linking a filament to openPrintTag database
Hello, I share some findings during my discovery of filament-db 

When trying to link a filament to the OpenPrintTag database, I have some difficulties to easily find my filament.

I have the feeling we can't use keywords for brand and/or material and/or name in the same time. Example:
 - I start typing the brand name: Arianeplat -> OK the UI show some resuts of the Brand
 - I continue adding the first letter of the color (used in the title) -> no result
 - I try with brand name + the first letter of the material (like PL for PLA) -> no result

Could be interesting to add a fuzzy search for this UI.

hyiger:
## Summary

The "Link to OpenPrintTag" dialog (`OptLinkDialog.tsx`) and the `/openprinttag` browser had byte-identical filters requiring the **whole query as one contiguous substring inside a single field** (`name | brandName | type`). The row title is the material name *without* the brand (brand is the subtitle), so the reporter's natural queries — "arianeplast b" (brand + color initial), "arianeplast pl" (brand + type prefix) — matched nothing: no single field contains those strings.

- New pure helper [src/lib/materialSearch.ts](src/lib/materialSearch.ts): `matchesTokenizedQuery(fields, query)` — split on whitespace; **every token must substring-match at least one field** (AND across tokens, OR across fields), case- and accent-insensitively. Accent folding is NFD + the explicit combining range, **not** `\p{M}` (ES2018 property escapes; the repo targets ES2017 and `tsc` validates regex flags but not bodies — the tsplEncoder lesson). Matters for this international dataset ("Améthyste" ↔ "amethyste").
- Both surfaces swap to the shared helper so they can't drift. Empty-query gates and the dialog's 50-result cap unchanged.
- **Strict superset for single-word queries** (identical results), and a contiguous phrase inside one field ("pla blanc") still matches under token-AND — no regressions.
- Deliberately out of scope: true fuzzy/typo-tolerant matching (needs a scoring dep + reordering — can be a follow-up if wanted), and the FilamentPicker / inventory searches (different surfaces the issue didn't name).

## Test plan
- [x] `tests/materialSearch.test.ts` — the two reported flows verbatim, AND/OR semantics, accent folding both directions, case-insensitivity, whitespace/empty queries, null fields; 100% line/branch coverage on the new lib (coverage-gated dir)

Fixes #1173

### DIFF
--- a/src/app/openprinttag/page.tsx
+++ b/src/app/openprinttag/page.tsx
@@ -9,6 +9,7 @@ import { useTranslation } from "@/i18n/TranslationProvider";
 import { useNumberFormat } from "@/hooks/useNumberFormat";
 import { safeHttpUrl } from "@/lib/safeRenderUrl";
 import { formatMinutesAsHm } from "@/lib/formatDuration";
+import { matchesTokenizedQuery } from "@/lib/materialSearch";
 
 // Row layout constants — the virtualized List needs known heights so it
 // can compute the absolute scroll position of every row without mounting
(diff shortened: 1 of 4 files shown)


hyiger:
@codex review

hyiger:
@codex review

hyiger:
@codex review

Issue #1177, opened 2026-08-19
Ability to import from OpenPrintTag scan like via Prusa QR Code - v1.78.0


tonysurma:
Cannot easily import from an OpenPrintTag nfc tag on a spool which also causes the second issue.

Reproduce:
a) get a new spool
b) scan openprinttag
c) get note saying not in db, view current filament or create variant
d) choose create variant
e) form is shown with NO data imported from tag, thus you have to manually enter and thus you are not linked to the openprinttag data nor the spool id
f) if you do the above (don't want to) and write nfc, it tells you, you are overwriting which don't want to do

However related issue of what works and then doesn't: if you do import via the prusa qr code it brings in all the right info and links to the spool.  But if you then try to scan the same open print tag it doesn't know it is related and if you try to write the nfc tag (say with updated weight) it says it will overwrite and fear is it will lose all the info the spool started with.  Should be able to map the two via the prusa code or something like that unique in the tag if it is there.

Ideally it would at step (c):
* see an existing filament & spool match up to spool via unique code
* or create a new spool under existing filament if filament is imported from openprinttag db but that spool isn't there yet
* or offer to link to existing filament and/or create variant as it does today BUT with bringing in the info from the tag (color, etc.) and bonus have a way to link into the openprinttag database for updates.

PR #1183, opened 2026-08-27
fix(nfc): create-variant-from-tag actually prefills the tag's data (#1177 Phase 1)

hyiger:
## Summary

Phase 1 of the #1177 triage plan — the reported step (e): scanning an OpenPrintTag and choosing **Create variant** opened an empty form.

**Root cause** (documented in-code as a deferred follow-up, Codex #706 r9): on `/filaments/new`, the `from_nfc` effect seeds the tag data synchronously, then the async `?parentId=` parent-loader **replaces `initialData` wholesale** and always lands last (the `parentLoading` gate keeps the form unmounted until it resolves) — every tag field was discarded.

## Changes

- The `from_nfc` param mapping is extracted to a module-level `nfcPrefillFromParams`; the `from_nfc` effect **skips** when `?parentId=` rides along; the parent-loader **composes** tag prefill + `parentId` + `_parent` in one `setInitialData`.
- **Standalone-only defaults fall through to the parent**: an absent `diameter` param no longer stamps 1.75 on a 2.85 mm family; absent `color` no longer stamps gray; absent `type` no longer stamps PLA. Family identity (vendor/type) prefers the **parent's** — the tag's brand string ("Prusament") routinely differs from the stored vendor ("Prusa Research").
- **GH #106 guard** — new pure [src/lib/nfcVariantPrefill.ts](src/lib/nfcVariantPrefill.ts) (100% line+branch coverage): a tag value **equal** to the parent's stored value is pruned so the variant keeps inheriting dynamically (the v1.52 `pruneOptPayloadAgainstParent` rule applied client-side). Inheritable scalars drop when equal; temperature subfields null individually; `optTags`/`secondaryColors` prune on set-equality; identity (name/color/colorName) and inventory (`totalWeight`) always survive; the Codex #706 r7/r8 zero-tare pin is preserved when the parent's tare differs.
- `NfcReadDialog`'s two create paths share one `buildTagParams`, so Create-as-variant forwards the **same full field set** as Create New (weights, density, temps, colorName, optTags, …). The old three-field variant path existed only because the race discarded everything anyway; its deferred-follow-up NOTE is deleted.

## Verification

Browser (dev server, read-only): `/filaments/new?from_nfc=1&parentId=<PC Blend Jet Black>&name=…&color=%23FF0000&density=1.22&nozzle=275&weight=1000` renders **Add Color Variant** with the tag name + color seeded, density `1.22` kept as an override (parent: 1.2), and the nozzle temp **blank with placeholder 275** — it equals the parent's, so it was pruned and inherits live. Before the fix the form rendered empty.

## Test plan
- [x] `tests/nfcVariantPrefill.test.ts` (11 tests, 100% line/branch on the new lib)

Part of #1177 — Phases 2–5 (spool-identity adoption, matched-spool dialog actions, factory-tag write preservation, OPT-DB linkage) are separate PRs pending the security sign-off + hardware diagnostic flagged in the triage report.


hyiger:
@codex review

hyiger:
Addressed in 27c4e256 — parent-tare gross-up for net-only tag weights (verified in-browser: 500 net + 186 inherited tare seeds 686), positional secondary-color comparison, and settings-bag pruning.

@codex review

hyiger:
@codex review

hyiger:
Addressed in 6f890a87 — density/diameter are snapped to the #570 grid on both sides before the equality judgment (half-float dust no longer persists parent-equal overrides), and the settings prune compares against the first element of an array-valued parent bag entry, matching the form's display semantics.

@codex review

hyiger:
Addressed in 33454664 — the settings prune now mirrors getSettingVal's String-coercion (array-first + numeric scalars), judging exactly what the form displays.

@codex review

hyiger:
Addressed in e4f320a4 — only the tag side is snapped (what the form submits); the parent side compares exact, so an off-grid parent no longer swallows a genuinely different tag value.

@codex review

hyiger:
Addressed in 187f41c3 — the three form-unwrapped bag keys (filament_notes, start/end gcode) compare through unwrapIniString on both sides, judging what the form displays and round-trips.

@codex review

hyiger:
Addressed in 396380b2 — a secondary-only tag now seeds color: null (the coextruded spec shape) instead of inheriting the fresh-form gray via undefined.

@codex review

hyiger:
Addressed in 230587ab — the param mapping now matches decodedTagToFilamentPayload: first-layer null, min/max on nozzleRangeMin/Max, bed min dropped (no schema fields).

@codex review

hyiger:
Addressed in 6bc1a95e — buildTagParams forwards the aux recommended temps as nozzleRec/bedRec and the mapping prefers them over the range maxima, matching decodedTagToFilamentPayload.

@codex review

hyiger:
Addressed in 8be4a3ae — preheat/drying/TD ride the URL flow and map exactly like decodedTagToFilamentPayload; drying + TD join the parent-equal prune.

@codex review

PR #1182, opened 2026-08-27
feat(dashboard): in-app "Log print job" dialog

hyiger:
## Summary

Adds the first first-party writer of `POST /api/print-history` — a **Log print job** dialog on the Dashboard's Recent-prints header. Until now jobs only entered via slicer/printer integrations, so Recent prints and the Analytics Print-jobs tile stayed empty for everyone else (the empty-state copy even said "none ship built-in yet").

Pure UI over the shipped, hardened endpoint — spool debit, `"manual"` source default, validation, the GH #305 all-retired no-debit contract, and the #1121 legacy-roll migration all stay server-side.

- **Dialog** (`src/components/LogPrintJobDialog.tsx`, OptLinkDialog focus-trap pattern): job label, printer select, date defaulting to today and submitted as the bare `YYYY-MM-DD` (v1.63 UTC-midnight convention → analytics day-buckets land on the picked calendar day in every timezone), notes, and dynamic filament/spool/grams rows (cap 100) with an **Auto** spool option that omits `spoolId`.
- **Semantics guardrails**: templates excluded from the picker (#605 — no inventory); an inline amber warning when a row's filament cannot be debited (all spools retired, or no spools and no legacy weight) so the GH #305 no-debit path never silently no-ops. A spool-less filament *with* a legacy top-level weight is deliberately not warned — the server materializes and debits it (#1121). Footer hint warns against also logging the print as manual spool usage (double-count). Server errors, incl. the 409 retry contract, surface verbatim.
- **Pure helper** `src/lib/printJobForm.ts` (validate + build body, 100% line/branch coverage) so the submit logic is unit-tested despite no component-test infra. `Number()` not `parseFloat` so `"12abc"` grams is rejected, not silently 12.
- **i18n**: 32 new `printJob.*` keys in EN + DE, plus updates to the two now-false strings (`dashboard.recentPrints.empty`, `analytics.jobsApiHint`).
- `onLogged` bumps the dashboard's existing `reloadKey` so the card refetches.

## Verification

Against the dev server (production data, read-only): dialog renders in dark mode; printer + filament pickers populate (templates filtered — 83 filaments incl. 11 templates → 74 options); selecting a filament fills its spool select with `label/instanceId (formatted grams)`; submitting with an empty label shows the client-side "Job label is required" with no network call. No real job was submitted (dev server writes to production Atlas).

## Test plan
[...]

Issue #1069, opened 2026-08-06
Question: Recent Prints


Rebel154:
Hi,
How is the Recent Prints section on the Dashboard populated? Is there a setting (either in FilamentDB or your version of Prusa Slicer) that needs setting? Same with the Print Jobs box on the Analytics tab.
Also, is there anyway to review the manual usage entries other than the Usage by Day graph on the Analytics tab? At the moment I don't see the point of entering the optional job label if it can't be recalled somewhere.

PR #1184, opened 2026-08-27
feat(history): print-history / usage browser page

### LINKED ISSUE #1168: feat: print-history / usage browser page
Split from #1069 (the maintainer comment there promised tracked follow-ups). The reporter's core point: the optional job label on manual usage entries has no surface where it can be recalled — the only views are the per-spool disclosure on the filament detail page (made visible by the #1069 fix) and the aggregate Usage-by-day chart.

A browser page consuming the already-shipped `GET /api/print-history` (filters: filamentId, printerId, limit) with per-job rows (label, printer, date, grams, per-filament usage) and delete-with-refund via the existing DELETE handler. Optionally a cross-spool usage-entry search by label, which would also cover manual entries that never became PrintHistory rows.

The jobs-vs-manual counting separation stays as is (deliberate double-count avoidance, documented in docs/api.md).

hyiger:
## Summary

Two-tab `/history` page (nav entry after Analytics) — the first surface where a usage entry's `jobLabel` can be recalled outside a single spool's disclosure. Split from #1069; the jobs-vs-manual counting separation documented in docs/api.md stays exactly as is, which is *why* this is two tabs rather than one merged list (job/slicer ledger entries are projections of PrintHistory rows and would double-show).

### Print jobs tab (shipped endpoints only)
- `GET /api/print-history` rows: label, printer, date, source, expandable per-filament breakdown linking to `/filaments/{id}?spool=…` (the existing deep-link highlight + promotion self-healing).
- Client-side label search + printer filter; a "most recent 200" note when the window fills.
- **Delete-with-refund** via the existing idempotent DELETE: ConfirmDialog wording states the refund explicitly ("Delete and refund"); the 409 concurrent-edit message surfaces verbatim.
- Date-only backfills render as UTC calendar days (the #941 `isUtcMidnight` convention — same class as the Codex P2 on #1182).

### Spool usage ledger tab (one new read-only endpoint)
- **New `GET /api/spools/usage-search`**: manual entries (`source:"manual"`, `jobId:null`) exist *only* inside `spools[].usageHistory` — nothing unwound that array across spools, so a manual entry's label had no recall surface at all. Aggregation over active filaments: label substring (`escapeRegex` + the GH #513 128-char cap — regex-compiling GET), source filter, limit 1..1000, date-desc, **strict #1005 projection** (photoDataUrl/dryCycles never ride — pinned by test).
- Tab defaults to `source=manual`; footnote discloses the 1,000-entry per-spool cap (manual entries evict first).
- Documented in `public/openapi.json` + `docs/api.md`.

### Pending follow-up commit on this PR
A "View all →" link on the Dashboard Recent-prints header — deferred until #1182 (which reworks that same header) merges, to avoid a pointless conflict.

## Verification
Browser against the dev server: nav entry + both tabs render (dark mode), jobs tab shows the correct empty state on a DB with no PrintHistory rows, ledger tab returns a well-formed empty set on live data with no usage entries; filters and source select wired.

## Test plan
- [x] `tests/usage-search-route.test.ts` (8 tests, mongodb-memory-server): flat-row shape + date-desc ordering, case-insensitive literal-regex label match (`.*` matches nothing), source filter, trash exclusion, 400s for bad label/source/limit, limit-after-sort, no-photo-blob leak
- [x] i18n parity (25 new keys, EN + DE)

Fixes #1168


hyiger:
@codex review

hyiger:
Addressed in de7b9448 — refund wording is now conditional on the spools still existing, trashed filament refs render unlinked with an in-trash marker (populate now selects _deletedAt), and the 200-job window disclosure survives an empty search.

@codex review

hyiger:
Addressed in 060312a9 — the delete confirmation quotes the sum of debitedGrams (what the refund actually restores), the jobs-tab truncation disclosure is captured at fetch time so a local delete can't hide it, and the ledger tab now discloses its own 200-row window. Also merged main and added the Dashboard View-all entry point deferred behind #1182.

@codex review

hyiger:
Addressed in 3ba21730 — the ledger disclosure now states only the window limit (a full window can't distinguish exactly-200 from more-than-200 with the current response shape), per the suggested wording.

@codex review

hyiger:
Addressed in c3ae8f17 — job totals sum through sumUsageGrams, the refund sum clamps per-entry debitedGrams/grams via safeGrams, and ledger rows clamp the same way (GH #1030/#1078 posture).

@codex review

hyiger:
Addressed in 4715742c — expanded per-filament rows clamp through safeGrams like the total and the ledger rows.

@codex review

hyiger:
Addressed in a186df15 — the desktop nav / hamburger breakpoint moves to lg so the eight-link set never shares 768-1024 px with the brand and status pills (verified at 800 px).

@codex review

hyiger:
Addressed in 4e5f11ba — spool-less (GH #305 no-debit) rows contribute zero to the refund promise, and the nav breakpoint moves to xl so the widest German pill states never share the bar with all eight links.

@codex review

hyiger:
Addressed in 1379ae8f — the printer filter unions in printers referenced by the fetched jobs (marked "in trash"), so trashed printers' rows stay filterable.

@codex review

hyiger:
Addressed in 8cba78dc — the trashed marker waits for a successful /api/printers load (no race/failed-fetch mislabels), and the refund confirmation quotes "up to {grams} g" naming the capacity clamp.

@codex review

hyiger:
Addressed in 81ae6e36 — the success toast carries the same up-to-capacity qualification as the confirmation.

@codex review

hyiger:
Addressed in 70fbce0e — the persistent footer note now carries the same still-exists + up-to-capacity qualification as the confirmation and toast. Also merged main (#1183 landed).

@codex review

hyiger:
Addressed in a7336a73 — the filter now fetches /api/printers?includeTrashed=1 (new opt-in param, OpenAPI-documented + route-tested) and labels on the explicit _deletedAt, independent of the job window; the absence-inference union and its race gate are gone.

@codex review

hyiger:
Addressed in b344feaa — the quote applies the DELETE handler's own debit validity rule (negative/non-finite/over-grams debits fall back to the clamped requested grams; GH #305 spool-less rows stay zero).

@codex review

hyiger:
Addressed in 7e32f2d4 — the skip checks spoolId alone; a spool-less row quotes zero regardless of any bypassed-write debitedGrams, matching the handler.

@codex review

hyiger:
Addressed in f7289f56 — the toast confirms only the deletion (the one fact the client knows); the qualified refund semantics stay on the confirmation and footer. This closes the class terminally: no refund claim remains in any always-true position.

@codex review

hyiger:
Addressed in be45e83b — the populate selects _purged and purged refs render through the removed fallback; only restorable trash gets the in-trash label.

@codex review

hyiger:
Addressed in 47c2dac3 — the confirmation drops the numeric quote (the PrintHistory debit copy is unverifiable against the ledger copy the DELETE refunds from) and states only the semantics; the estimator is deleted. With r15's toast, no client-computed refund number remains anywhere.

@codex review

hyiger:
Addressed in e6f9e2b6 — the endpoint clamps grams through safeGrams before serializing (raw-driver-write route test pins Infinity → 0, keeping the documented numeric shape).

@codex review

hyiger:
Follow-up fcaf2fdb: the raw-driver-write sanitization test the previous commit's message promised (a patch anchor mismatch had dropped it — Infinity → 0 is now actually pinned, 9 route tests).

hyiger:
The clean review landed on e6f9e2b640 — one test-only commit behind HEAD (the promised sanitization test that a patch anchor mismatch had dropped). Re-requesting for the exact HEAD.

@codex review

PR #1185, opened 2026-08-28
fix(tds): Gemini extraction self-heals when its model is retired

### LINKED ISSUE #1179: Gemini API Usage is out of date - 1.78.0
In short, if you try to use TDS import with a google gemini key, you get this error stating the code needs to be updated to use a current model, thanks.

Image

hyiger:
## Summary

Google retired `gemini-2.5-flash` out from under v1.78.0, breaking TDS import for every Gemini user — the **second** hardcoded-model death in this feature (#916 was `gemini-2.0-flash`, which surfaced as a misleading 429; this one at least says to update the code). Bumping the constant again would just schedule this issue's sequel, so the bump to Google's stated migration target (`gemini-3.1-flash`) ships with a self-healing fallback:

- **`isGeminiModelGoneError`** classifies the model-gone shape (404, or a model-referencing not-found / deprecated / retired / no-longer / not-supported body) while explicitly excluding 429s (rate limits keep their own retry/backoff path) and key errors.
- On that shape, **`callGemini` discovers a currently-served model via the ListModels endpoint** (already used for key validation), retries once, and caches the discovery for the process lifetime — a future retirement degrades to one extra round-trip instead of a broken feature.
- **`pickGeminiModel`** (pure, exported, tested) picks a text `generateContent` model: flash tier first (the feature's free-tier cost posture), stable over preview/exp, higher version, plain flash over `-lite`; modality-suffixed models (image/tts/audio/embedding/live/veo) excluded.
- When nothing can be discovered, the error finally says what's wrong — the model is no longer available and no replacement was found — instead of a raw HTTP body.

`docs/usage.md`'s model table is updated.

## Test plan
- [x] 12 new tests in `tests/tdsExtractor.test.ts`: the picker's ranking/exclusion rules, the classifier's shapes (incl. the 429 exclusion), and the discover-retry-cache flow end-to-end through `extractFromTdsContent` (404 → ListModels → retry succeeds → second extraction goes straight to the cached model; no-replacement and network-failure error paths; a non-model failure on the retry surfaces verbatim)
- [x] Full coverage gate: 99.3% lines / 98.8% stmts / 98.0% funcs / 97.1% branches (exit 0)

Fixes #1179


hyiger:
@codex review

hyiger:
Addressed in ed46397b — an explicit model-gone marker now beats the 429 status (the #916 shape heals end-to-end, pinned), and the replacement caches at discovery time so backoff re-entries go straight to it (pinned: one default attempt, one discovery).

@codex review

hyiger:
Addressed in 0417b773 — the discovery cache is keyed per API key (module-private Map); pinned by a cross-key test: key B starts from the default and runs no inherited discovery.

@codex review

hyiger:
Addressed in dd04675e — discovery follows nextPageToken (bounded to 5 pages) before picking; two-page test pins the later-page flash winning over the first page's pro.

@codex review

hyiger:
Addressed in 6da05d6b — the cache keys by SHA-256 digest with a 32-entry insertion-order eviction bound (no plaintext keys, bounded memory), and failed ListModels bodies drain before the pagination loop breaks (the GH #955 undici class).

@codex review

hyiger:
Addressed in 722c40d9 — pageSize=1000 + token-to-the-end pagination (the 50-page bound is purely a runaway guard, beyond any real roster), and the German docs table is synced.

@codex review

